PR #395, opened 2014-07-31
Adds examples to .dir-locals.el section in README

joelmccracken:
CF #387.

I think this might show that it would be nice to have a more-convenient way to override various pieces of functionality, if possible. That would make customizing Projectile's behavior easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,11 +428,81 @@ use it you must create file named `.dir-locals.el` inside project
 directory. This file must contain something like this:
 
 ```
-((nil . ((eval . (
-                  (setq secret-ftp-password "secret")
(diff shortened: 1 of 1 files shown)


nickmccurdy:
For anyone looking for this now, it's moved to http://projectile.readthedocs.io/en/latest/configuration/#storing-project-settings.

TheJJ:
Now it's here: https://docs.projectile.mx/en/latest/projects/#storing-project-settings :)

joelmccracken:
haha wow! I completely forgot that I did this. I only rememberd doing the switch-to-project funcitonality

PR #427, opened 2014-08-31
add html and js to projectile-other-file-alist

jeroentbt:
I've added this, to great delight - Thank you! - to my emacs.d for when I'm writing meteor.js code.
No idea if this comes in handy for other frameworks' naming conventions.
Just throwing it out there.


bbatsov:
I don't think we can add this as some generic mapping. Ideally this type of config should be specific to a particular project type.

PR #433, opened 2014-09-06
Allow wildcards in projectile-project-root-files

markus1189:
So after we got haskell project identification working which is enough if the project is under version control, it would be nice if projects without vcs are recognized as well. :)

To do so, we have to allow wildcards when searching for project-root-files as well.

What do you thing about replacing `projectile-file-exits-p` with `file-expand-wildcards` again?

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -615,7 +615,7 @@ Returns a project root directory path or nil if not found."
   (projectile-locate-dominating-file
    dir
    (lambda (dir)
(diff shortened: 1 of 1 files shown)


fiddlerwoaroof:
This would be really nice for Common Lisp too: I'm not really a huge fan of cluttering my filesystem with files just to make my tools happy, when there's an obvious way to detect the information those files would provide.

As far as performance goes, why not just find the project root when a buffer is opened and stash it in a buffer-local variable or some other caching mechanism?

Issue #498, opened 2014-10-30
projectile-project-root-cache should be used without calls to file-truename


thomasf:
Why: I have tried to do a quick search/replace port of ibuffer-vc to ibuffer-projectile a couple of times but the performance has been horrible.. Its kind of usable right now, just need to shave off a few thousand file-truename calls to make it snappy.

I do not have time to implement right now so I'll just leave this here.. The projectile-ibuffer is comong though. It works without causing extremely painful update times now thanks to the project root cache .

PR #499, opened 2014-10-30
Ibuffer projectile (WIP)

thomasf:
A few things todo.

* clean up possible left overs from quick and dirty vc -> -projectile string rename.
* investigate if the additional project root cache in ibuffer-projectile can be obsoleted by modifying projectile.el instead (#498).
* dig about in my own init.el that have a few ibuffer/projectile integrations that might fit here
* update readme
* add tests
* move other ibuffer functions from projectile.el, no need to require ibuffer there for people who are not using it


bbatsov:
@thomasf Any progress here?

bbatsov:
Reopen this if you make more progress with it.

PR #566, opened 2014-12-06
projectile-follow-symlinks in projectile-file-truename

jacobono:
I ran into this issue using clearcase -- I was trying to create a directory with soft links to the dynamically mounted vobs, and once I was in the vob, projectile couldn't resolve the project because file-truename follows symlinks :(

I tried just disabling `file-symlink-p` in .dir-locals, but then I needed symlinks resolved for `vc-mode`.

So I have been using this PR for a day or so, and I have seen no problems using it with clearcase or another project where symlinks are being followed.

Let me know if there is anything else I can do for this!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -592,7 +592,7 @@ Returns nil if no window configuration was found"
 (defadvice delete-file (before purge-from-projectile-cache (filename &optional trash))
   (if (and projectile-enable-caching (projectile-project-p))
       (let* ((project-root (projectile-project-root))
-             (true-filename (projectile-file-truename filename))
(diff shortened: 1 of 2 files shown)


jacobono:
@bbatsov, what do you think of just using this PR when trying to resolve `projectile-project-root`?

gfredericks:
This monkeypatch-workaround fixed the issue for me: https://github.com/gfredericks/dotfiles/blob/448fa6ac1e6714fda55c16e00899bec1af5e6799/base/.emacs.d/user/gfredericks-base.el.symlink#L53

PR #571, opened 2014-12-08
projectile-project-root-of-filepath

### LINKED ISSUE #570: project-root-of-file function
There is currently no way to get the project root of an arbitrary (perhaps unopened) file in projectile. projectile-project-root does approximately the right thing, and could be adapted very easily.

IvanMalison:
closes #570. This function provides a simple way to lookup the projectile project to which an arbitrary filepath belongs.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -655,12 +655,12 @@ Returns a project root directory path or nil if not found."
    nil
    (or list projectile-project-root-files-top-down-recurring)))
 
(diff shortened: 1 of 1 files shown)


emacs18:
Pull request 669 is an alternate solution that generalizes projectile-project-root to take optional argument rather than creating another function.

IvanMalison:
Closing this in favor of #669

PR #573, opened 2014-12-09
 Fix and refactor getting files using external tools

tarsius:
```
The entry point is `projectile-dir-files-external'.  When the project is
a Git repository it also returns the files in submodules.  For other
project types nothing equivalent is done, but an abstraction to do that
in the future already existed.

Unfortunately that was implemented using six additional functions.  This
lead to a lot of accidental complexity, which in turn lead to work being
done which did not need to be done.  Also it didn't quite work in the
only case where we actually made use of this; Git submodules that are
located outside the directory we are interested in where also recursed
into.

So remove all of that and just keep `projectile-dir-files-external'.
When the project is a Git repository, then that function calls out to
the new function `projectile-dir-files-git', which as the name suggests
only knows about Git.

Provided there actually is another vcs which supports the equivalent of
Git submodules, then that should implemented as a separate function.
That function might share some code with the Git variant, but that's
okay; a bit of harmless code duplication is better than a broken
abstraction.

`git submodules foreach COMMAND' calls COMMAND for every initialized
submodule.  This includes submodules which are not below the current
directory but $path is always relative to that directory.  We need to
filter out such submodules, and that's what is responsible for most
of the complications in `projectile-dir-files-git'.

On the other hand, because `git submodules foreach COMMAND' ignores
submodules that have not been initialize yet, there is no need to
explicitly deal with modules which have themselves or their superproject
as a submodule.  There is no risk of infinite recursion because for that
to happen the submodules would have had to be initialized recursively,
which would have failed eventually due to there being no space left on
the harddrive.
```

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -707,16 +707,13 @@ A thin wrapper around `file-truename' that handles nil."
 (defun projectile-dir-files (directory)
   "List the files in DIRECTORY and in its sub-directories.
 Files are returned as relative paths to the project root."
-  ;; check for a cache hit first if caching is enabled
-  (let ((files-list (and projectile-enable-caching
-                         (gethash directory projectile-projects-cache)))
-        (root (projectile-project-root)))
-    ;; cache disabled or cache miss
-    (or files-list
+  (or (and projectile-enable-caching
+           (gethash directory projectile-projects-cache))
+      (let ((root (projectile-project-root)))
         (if (eq projectile-indexing-method 'native)
             (projectile-dir-files-native root directory)
-          ;; use external tools to get the project files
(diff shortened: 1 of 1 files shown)


tarsius:
Since I am going on vacation now it will actually be a few weeks until I can get back to this.

bbatsov:
@tarsius Roger that! Enjoy your vacation!

bbatsov:
@tarsius ping :-)

bbatsov:
@tarsius Any plans to revisit this or should I close it?

tarsius:
Ups. I have now added this to my private todo list, but don't see how I can get to this anytime soon, because of https://github.com/magit/magit/pulse/monthly. But feel free to close, actually let me do that for you.

PR #592, opened 2014-12-30
Use `projectile-log' instead of `message' when non-interactive

abo-abo:
* projectile.el (projectile-do-log): New custom.

  (projectile-log): New function to be used instead of message unless
  the current command was invoked interactively.

  (projectile-cache-current-file): Use `projectile-log' instead of
  `message'.


thomasf:
The messages in this patch seems like a good candidate to hide if `projectile-verbose` is `nil`... 
I agree with that we should not need log levels at all, I'm very ok with that errors are always printed since they generally does not happen anyway..

PR #598, opened 2015-01-07
feature; helm-projectile-ag support ag options

jiahut:
In my project, it has many files like  `.js`, `.html`,  `.rb`, `.coffee` , etc .   I  alway  only search Javascript files or html files at once. We know `ag` support many options , like:

```bash
ag searchterm -G js$ # Only search Javascript files. -G allows for regular expressions
```

so, add some option before doing `helm-projectile-ag` is necessary


tuhdo:
I think it should not ask every time `helm-projectile-ag` is invoked. Should only ask when a prefix argument is supplied. You can also ignore file by extensions by customize one of the 4 variables:

- `grep-find-ignored-files`: List of file names which rgrep and lgrep shall exclude, and helm-projectile-ack also uses this variable.
- `grep-find-ignored-directories`: List of names of sub-directories which rgrep shall not recurse into. helm-projectile-ack also uses this variable.
- `projectile-globally-ignored-files`: A list of files globally ignored by Projectile. Note that setting this variable also filters files listed in `projectile-find-file`.
- `projectile-globally-ignored-directories`: A list of directories globally ignored by Projectile. Note that setting this variable also filters files listed in `projectile-find-dir`.

i.e. if you want `helm-projectile-ag` to ignore html files, add `.*.html` to either of those variables.

jiahut:
@tuhdo 
> customize one of the 4 variables ...

do you mean `M-x customize-variable` ? 
In other words  , I want ignore some file dynamic . now is `html`  files , for a while is `ruby` files . 
I want add some options dynamic .

tuhdo:
Maybe you should add a variable that when set to `t`, always asks for option. Otherwise, use prefix argument to ask for options.

jiahut:
@tuhdo 
>  use prefix argument to ask for options

yeah, It  will be better .

thinks.

bbatsov:
@tuhdo Ping :-)

tuhdo:
Yes, you can merge it :)

bbatsov:
It has to be rebased on top of the current `master` first.

bbatsov:
@jiahut ping :-)

semanticart:
gentle bump, @jiahut 

I'd love to see this merged.

jiahut:
@semanticart @bbatsov 
think you.

bbatsov:
Squash the commits together and update the commit message to match the style suggested in CONTRIBUTING.md.

jiahut:
@bbatsov 
goto #687

Issue #458, opened 2014-10-09
Projectile helm sources signal error when not in a project


tsdh:
I use helm and want to use the helm-projectile sources, too.  Therefore, I configured

```` elisp
(setq helm-for-files-preferred-list (append helm-projectile-sources-list
                                            helm-for-files-preferred-list))
````

so that `helm-for-files` shows the projectile sources on top the normal file sources.

That works fine when the current buffer visits a file in some project.  However, I get an error if I call `helm-for-files` from some other buffer, e.g., Gnus' *Group* buffer.  The backtrace is:

````
Debugger entered--Lisp error: (error "You're not in a project")
  signal(error ("You're not in a project"))
  error("You're not in a project")
  projectile-project-root()
  projectile-project-buffers()
  projectile-project-buffer-names()
  (setq helm-projectile-buffers-list-cache (projectile-project-buffer-names))
  (lambda nil (setq helm-projectile-buffers-list-cache (projectile-project-buffer-names)) (let ((result (cl-loop for b in helm-projectile-buffers-list-cache maximize (length b) into len-buf maximize (length (with-current-buffer b (symbol-name major-mode))) into len-mode finally return (cons len-buf len-mode)))) (unless helm-buffer-max-length (setq helm-buffer-max-length (car result))) (unless helm-buffer-max-len-mode (setq helm-buffer-max-len-mode (cdr result)))))()
  apply((lambda nil (setq helm-projectile-buffers-list-cache (projectile-project-buffer-names)) (let ((result (cl-loop for b in helm-projectile-buffers-list-cache maximize (length b) into len-buf maximize (length (with-current-buffer b (symbol-name major-mode))) into len-mode finally return (cons len-buf len-mode)))) (unless helm-buffer-max-length (setq helm-buffer-max-length (car result))) (unless helm-buffer-max-len-mode (setq helm-buffer-max-len-mode (cdr result))))) nil)
[...]

PR #601, opened 2015-01-09
Enable helm projectile sources reusable

tuhdo:
Currently, the sources can only be used in a project recognized by
Projectile. It cannot be composed to be used with other commands like
helm-for-files. This change supports such use case without compromise
the consistency of Projectile: Helm Projectile commands are responsible
for throwing the same errors as stock Projectile, but not the source.


tuhdo:
Really fix https://github.com/bbatsov/projectile/issues/458 this time.

PR #602, opened 2015-01-09
Check project root for helm-projectile-grep/ack/ag

tuhdo:


helm-projectile.el
```diff
@@ -685,27 +685,31 @@ If it is nil, or ack/ack-grep not found then use default grep command."
...
+  (if (projectile-project-p)
+      (funcall'run-with-timer 0.01 nil
+                              #'helm-projectile-grep-or-ack nil)
+    (error "You're not in a project")))
 
 ;;;###autoload
 (defun helm-projectile-ack ()
   "Helm version of projectile-ack."
   (interactive)
-  (let ((ack-ignored (mapconcat
```
bbatsov: Please, remove everything ack related. `ack-and-a-half` is dead and `ag` is much better anyways.

tuhdo: But `helm-projectile-ack` does not depend on `ack-and-a-half`. The only difference between `helm-projectile-ack` and `helm-projectile-grep` is actual external command, similar to how stock `helm-do-grep` can be used with `ack`.

And `ag` is really difficult to install on Windows.

bbatsov: OK.

PR #603, opened 2015-01-09
Fix a bug in creating new virtual dir

tuhdo:
Inserting file into new Dired buffer fails if the inserting file is an
empty string. Check to make sure no empty string is in the file list.

Also remove duplicate `C-c f` binding that shadows the command that
creates virtual Dired buffer.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -250,12 +250,14 @@ CANDIDATE is the selected file, but choose the marked files if available."
            (not helm-projectile-virtual-dired-remote-enable))
       (message "Virtual Dired manager is disabled in remote host. Enable with %s."
                (propertize "helm-projectile-virtual-dired-remote-enable" 'face 'font-lock-keyword-face))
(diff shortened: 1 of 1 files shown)


tuhdo:
@bbatsov ping

helm-projectile.el
```diff
@@ -250,12 +250,14 @@ CANDIDATE is the selected file, but choose the marked files if available."
            (not helm-projectile-virtual-dired-remote-enable))
       (message "Virtual Dired manager is disabled in remote host. Enable with %s."
                (propertize "helm-projectile-virtual-dired-remote-enable" 'face 'font-lock-keyword-face))
-    (let ((new-name (completing-read "Select a Dired buffer:"
+    (let ((new-name (completing-read "Select or enter a new Dired name: "
```
bbatsov: name or buffer name?

tuhdo: Done.

PR #606, opened 2015-01-12
[Fix #537] Correct command line arguments

### LINKED ISSUE #537: helm-projectile-grep does not work
`helm-projectile-grep` does not show any search result. I am using Emacs 24.4 on OS X Yosemite, latest version of bbatsov/projectile and helm. Other grep and helm commands (such as helm-do-grep) run perfectly. Please help. Thanks

@tuhdo @bbatsov

tuhdo:
Using (projectile-project-root), helm-projectile-grep breaks on Windows
because GNU Grep requires Unix path, while the call returns Windows
path. Fix it by using current directory and use default-directory to set
grep location.

Also make helm-projectile-grep works with Grep 2.5.4 (BSD) on Mac OS X.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -634,16 +634,16 @@ If it is nil, or ack/ack-grep not found then use default grep command."
          (grep-find-ignored-files (-union projectile-globally-ignored-files  grep-find-ignored-files))
          (grep-find-ignored-directories (-union projectile-globally-ignored-directories grep-find-ignored-directories))
          (helm-grep-default-command (if use-ack-p
-                                        (concat ack-executable " -H --smart-case --no-group --no-color " ack-ignored-pattern " %p %f")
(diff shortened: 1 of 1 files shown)


bbatsov:
> helm-projectile-grep now doesn't account for projectile-globally-ignored-directories
> and projectile-globally-ignored-files anymore.

This doesn't sound like an improvement to me.

tuhdo:
But it still accounts for `grep-find-ignored-files` and `grep-find-ignored-files`, and better stability.

tuhdo:
I think it again. I will take time to implement this properly independent of functions from `helm-grep`.

tuhdo:
@bbatsov I fixed it while not changing the original behaviour. Here is a screenshot showing that it works on Mac OS X:

screen shot 2015-01-16 at 9 18 48 pm 2

PR #611, opened 2015-01-19
Update helm-projectile.el

emacs18:
Use built-in string= rather than undefined string-empty-p function.


emacs18:
Closing this one since I submitted a new one that has two issues combined.

PR #612, opened 2015-01-19
Update helm-projectile.el

emacs18:
Use built-in cl-remove-if-not rather than undefined "filter" function.


tuhdo:
I think you should squash this and the other commit into one. Thank you for noticing this. Or change the commit title to something more meaningful, i.e. your comment above could be used as a much better descriptive title.

emacs18:
I just submitted a new pull request with the two issues combined.

PR #614, opened 2015-01-19
Use built in string= and cl-remove-if-not

emacs18:
Replace one use of string-empty-p with string= and one instance of "filter" with cl-remove-if-not to prevent "udefined function" errors for users who do not have string-empty-p and filter defined.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -253,8 +253,8 @@ CANDIDATE is the selected file, but choose the marked files if available."
     (let ((new-name (completing-read "Select or enter a new buffer name: "
                                      (helm-projectile-all-dired-buffers)))
           (helm--reading-passwd-or-string t)
-          (files (filter (lambda (f)
-                           (not (string-empty-p f)))
+          (files (cl-remove-if-not (lambda (f)
+                           (not (string= f "")))
                          (mapcar (lambda (file)
                                    (replace-regexp-in-string (projectile-project-root) "" file))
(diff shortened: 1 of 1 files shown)


helm-projectile.el
```diff
@@ -253,8 +253,8 @@ CANDIDATE is the selected file, but choose the marked files if available."
     (let ((new-name (completing-read "Select or enter a new buffer name: "
                                      (helm-projectile-all-dired-buffers)))
           (helm--reading-passwd-or-string t)
-          (files (filter (lambda (f)
```
bbatsov: `filter` -> `-filter`

helm-projectile.el
```diff
@@ -253,8 +253,8 @@ CANDIDATE is the selected file, but choose the marked files if available."
...
-                           (not (string-empty-p f)))
+          (files (cl-remove-if-not (lambda (f)
+                           (not (string= f "")))
```
bbatsov: `string-empty-p` was introduced in Emacs 24.4, so this is a reasonable change.

emacs18:
I see that string-empty-p is in subr-x.el in emacs.  However string-empty-p does not seem to be autoloaded.
Hence isn't (require 'subr-x) required in order to use string-empty-p?

bbatsov:
It's not a good idea to tie Projectile to 24.4 just for a single function.

emacs18:
I'm closing this now that I put up another request with both changes.

PR #615, opened 2015-01-19
prevent infinite loop with new virtual dired name

emacs18:
The proposed change is not meant to be checked in.
It is merely to document the problem.  

As documented in the code comment, I can't seem to exit the minibuffer when I type a new name.
First I type C-c p f then start selecting files via hitting Control-SPC on a few files on the ensuing helm buffer.
To create a new virtual dired buffer I type C-c f which asks for a dired buffer name in the mini buffer.
I type a new name, i.e., not a pre-exising dired buffer name then hit enter.
Rather than accepting the new name, emacs goes into infinite loop with "Display not ready" error message.

The hack in this request demonstrate that infinite loop is caused in part by helm-maybe-exit-minibuffer.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -250,6 +250,15 @@ CANDIDATE is the selected file, but choose the marked files if available."
            (not helm-projectile-virtual-dired-remote-enable))
       (message "Virtual Dired manager is disabled in remote host. Enable with %s."
                (propertize "helm-projectile-virtual-dired-remote-enable" 'face 'font-lock-keyword-face))
+    ;; When RET is hit to exit the minibuffer after typing a new dired buffer
+    ;; name as part of completing-read below, emacs goes into infinite loop with
+    ;; "[Display not ready]" message.  Apparently new names are not accepted.
+    ;; One way to prevent this infinite loop is to set
+    ;; helm--reading-passwd-or-string to non-nil which prevents
+    ;; helm-maybe-exit-minibuffer from going into infinite loop.
+    ;; This hack probably is not an acceptible solution to the problem.
+    ;; 
+    (let ((helm--reading-passwd-or-string t))
     (let ((new-name (completing-read "Select or enter a new buffer name: "
                                      (helm-projectile-all-dired-buffers)))
           (helm--reading-passwd-or-string t)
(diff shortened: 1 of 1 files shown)


emacs18:
The partial call stack for the infinite loop is shown below.

  helm-maybe-exit-minibuffer
  call-interactively(helm-maybe-exit-minibuffer
  command-execute
  read-from-minibuffer
  completing-read-default
  completing-read
  helm-projectile-dired-files-new-action
  apply
  helm-funcall-with-source
  helm-execute-selection-action-1
  helm-execute-selection-action
  helm-internal
  apply
  helm
  apply
  helm
  helm-projectile-find-file
  call-interactively
  command-execute

tuhdo:
I did not have this problem when entering new names. May I see your Helm config? Did you update to the latest Helm?

emacs18:
Hmm.  I too can't seem to reproduce the problem now.  Perhaps I made an error while experimenting.
I'll be on the lookout for this, and get back to you if I encounter this again.
Following is how I setup emacs after starting it with "emacs -q".
I then visited my git clone of emacs source code to hit C-c p f, etc.
My emacs is my own build of emacs from emacs-24 brandch of the git repo.
I usualy do "git pull" of emacs source once a week.
I use ubuntu 14.04.

(unless (string< "24.4" emacs-version)
  (error "You must use emacs 24.4 or later"))

;; Do not use standard path such as ~/.emacs.d so as not to pollute it.
(custom-set-variables
 `(user-emacs-directory ,(expand-file-name "~/.emacs.d-helm")))

(mkdir user-emacs-directory t)

(require 'package)

(custom-set-variables
 '(package-archives
   '(("gnu" . "http://elpa.gnu.org/packages/")
     ("melpa" . "http://melpa.org/packages/"))))

(package-refresh-contents)
(package-initialize)
(package-install 'helm)
(package-install 'projectile)
(package-install 'helm-projectile)

;; This sets (among other things) up global key binding with `C-x c' prefix.
(require 'helm-config)

(projectile-global-mode)
(custom-set-variables '(projectile-completion-system 'helm))
(helm-projectile-on)

;; Hack to work-around some temporary issues
(defalias 'filter 'cl-remove-if-not)
(defun string-empty-p (s) (string= s ""))

PR #616, opened 2015-01-19
Allow specifying regexp to projectile-grep

birkenfeld:
This is for consistency with -ack, and makes overriding the "query the user" part easier.

I put the new argument last to retain backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ buffers by name
 * Collect search tool (`grep`, `ag`, `ack`) keybindings under a common keymap prefix (`C-c p s`)
 * Remove `defcustom` `projectile-remember-window-configs` in favor of
   `persp-projectile.el`.
+* Add an argument specifying the regexp to search to `projectile-grep`.
 
 ### Bugs fixed
 
--- a/projectile.el
+++ b/projectile.el
@@ -1661,18 +1661,21 @@ This is a subset of `grep-read-files', where either a matching entry from
                (and ext (concat "*." ext)))))
       (or default-alias default-extension))))
 
-(defun projectile-grep (&optional arg)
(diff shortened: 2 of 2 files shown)


bbatsov:
Could you be a bit more specific about the use case(s) that warrant this change. 

P.S. projectile-ack was recently removed.

birkenfeld:
Basically I wanted a slightly different style of prompting, where the default is part of the query, and not in the editable part of the buffer (so it needs not be deleted when you don't want it).

For now I did this with the following advice:

```
(defadvice projectile-grep (around better-query activate)
  (let ((orig-read-string (symbol-function #'read-string)))
    (flet ((read-string (prompt &optional initial history default inherit)
                        (funcall orig-read-string
                                 (if (string-empty-p initial) prompt
                                   (concat prompt "[" initial "] "))
                                 nil nil initial inherit)))
      ad-do-it)))
```

BTW, ack has been removed, but ag also has a search-term argument while grep doesn't.

bbatsov:
Fair enough. Please, mention this in the changelog as well.

birkenfeld:
Done!

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ buffers by name
 * Collect search tool (`grep`, `ag`, `ack`) keybindings under a common keymap prefix (`C-c p s`)
 * Remove `defcustom` `projectile-remember-window-configs` in favor of
   `persp-projectile.el`.
+* Add an argument specifying the regexp to search to `projectile-grep`.
```
bbatsov: This has to be updated to reflect the current version.

bbatsov:
You'll have to rebase this on top of the current `master`.

bbatsov:

PR #617, opened 2015-01-21
Use '-filter' rather than 'filter', and use 'string=' rather than 'string-empty-p'

emacs18:
Here is an updated patch with suggested change.
Please ignore this, if this pull request is redundant.
I'm still learning how to use github properly.


bbatsov:
:+1:

PR #618, opened 2015-01-22
Fix: Projectile cannot remember projects

tuhdo:
If a user visits a project using Dired, the project won't be recorded
into project list because no projectile-cache-projects-find-file-hook is
executed. The hook must be added to dired-before-readin-hook that runs
before Dired does anything else.


bbatsov:
:+1:

PR #619, opened 2015-01-23
Autoloads for some projectile-find-file defuns

milkypostman:
More entry points here to projectile. Not sure what your opinions or standards are about when to set an autoload or not.


bbatsov:
:+1: Most users directly enable `projectile-mode`, so I haven't paid much attention to autoloads.

PR #620, opened 2015-01-23
Make projectile-find-file-dwim handle relative path

tuhdo:
It should be able to handle relative paths that point to file in a
project. For example, we have file custom/setup-a.el, and file
personal/setup-b.el contains ../custom/setup-a.el,
projectile-find-file-dwim should be able to jump to it.


tuhdo:
@bbatsov ping

PR #621, opened 2015-01-23
Use `magit-status` instead of `magit-status-internal`

vspinu:
If magit package is not yet loaded, `projectile` cannot find `magit-status-internal` because that function is not auto-loaded. `magit-status-internal` is now an alias for `magit-status`,  so `magit-status` should be used instead.


bbatsov:
Looks good to me. :+1:

PR #622, opened 2015-01-28
[Fix #560] [Fix #580] Add new defcustom that is used for globally ignoring files by extension

### LINKED ISSUE #560: Globally ignoring certain files
I don't see the possibility to globally ignore certain files in Projectile. E.g. I do not care about *.pdf files and would rather not see them in any of my projects. Currently, projectile-globally-ignored-files does not seem to use regexp or expand wildcards. Is that a feature that I am somehow missing or would you like implemented?

### LINKED ISSUE #580: idea for ignoring file extensions in alien mode
I have a `.projectile` file in my project root with the following ignores:

https://gist.github.com/it3ration/ab8620156e16c6c72121

Directories are ignored, but none of the file extensions are.  From reading the lisp code it appears that file extensions are only ignored if you enable native indexing.  This isn't fast enough on large projects.  It seems like you could ignore file extensions using the native `find` command in the following way:

`find . -not -name "*.exe" -not -name "*.dll"`

In this way you could still use the `find` command for speed, but apply extension ignores in your `.projectile` file.  Is this possible?

Thanks so much!

-Kevin

tolchz:
Add a new defcustom projectile-globally-ignored-file-extensions that is
used by projectile-remove-ignored to remove files based off of the file
suffix.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -258,6 +258,12 @@ pattern that would have found a project root in a subdirectory."
   :group 'projectile
   :type '(repeat string))
 
+(defcustom projectile-globally-ignored-file-suffixes
+  nil
+  "A list of file suffixes globally ignored by projectile."
+  :group 'projectile
+  :type '(repeat string))
+
 (defcustom projectile-globally-ignored-directories
   '(".idea"
     ".eunit"
@@ -852,8 +858,10 @@ Operates on filenames relative to the project root."
   (let ((ignored (append (projectile-ignored-files-rel)
                          (projectile-ignored-directories-rel))))
     (-remove (lambda (file)
-               (--any-p (string-prefix-p it file) ignored))
+               (or (--any-p (string-prefix-p it file) ignored)
+                   (--any-p (string-suffix-p it file) projectile-globally-ignored-file-suffixes)))
              files)))
+
 
 (defun projectile-buffers-with-file (buffers)
   "Return only those BUFFERS backed by files."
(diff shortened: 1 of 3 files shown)


bbatsov:
The code looks good to me, although I'm not sure about the default value of the config. We should also keep in mind the performance implications of running this extra check.

tolchz:
I just set the default for what I had been using.  I'd be good with a nil default which would effectively disable the check for everyone by default.  I've been using the change for a few months, without it I get so many .o and .d files that the completion isn't effective.

bbatsov:
This has to updated to match the current `master` branch.

projectile.el
```diff
@@ -241,6 +241,12 @@ pattern that would have found a project root in a subdirectory."
   :group 'projectile
   :type '(repeat string))
 
+(defcustom projectile-globally-ignored-file-extensions
+  (list ".o" ".d" ".pyc")
```
bbatsov: I'd suggest keeping this `nil` by default.

projectile.el
```diff
@@ -855,7 +861,8 @@ Operates on filenames relative to the project root."
   (let ((ignored (append (projectile-ignored-files-rel)
                          (projectile-ignored-directories-rel))))
     (-remove (lambda (file)
-               (--any-p  (s-starts-with-p it file) ignored))
+               (or (--any-p (s-starts-with-p it file) ignored)
```
bbatsov: We no longer use `s.el`.

bbatsov:
I also want to see some unit tests for this functionality.

tolchz:
Rewrote to work on top of the current master.

bbatsov:
:+1:

PR #623, opened 2015-01-29
Stop depending on s and f

tarsius:
If I remember correctly @bbatsov said that he would like these dependencies too.


bbatsov:
Yeah, I did say this and I welcome your PR. I also hope to replace `dash.el` with `seq.el` from ELPA (it will also be built-in in Emacs 25). 

P.S. The build is currently broken.

tarsius:
One problem is that it uses `string-suffix-p`, that wasn't added until 24.4. Should I add add

```lisp
(unless (fboundp 'string-suffix-p)
  (defun string-suffix-p ...
```

But that's not the cause of the error, will investigate

tarsius:
I am not familiar with `seq.el` yet, or 25.0.50 for that matter. I suspect it lacks many of the forms provided by `dash.el`. So you could probably start replacing some forms with variants from `seq` now but will likely have to wait until 25.1 until you can fully drop the `dash` dependency. Maybe we'll have `#(foo % constant)` by then and no longer need the anaphoric variants.

bbatsov:
> One problem is that it uses string-suffix-p, that wasn't added until 24.4. Should I add

Yep, you should do this.

> I am not familiar with seq.el yet, or 25.0.50 for that matter. I suspect it lacks many of the forms provided by dash.el. So you could probably start replacing some forms with variants from seq now but will likely have to wait until 25.1 until you can fully drop the dash dependency. Maybe we'll have #(foo % constant) by then and no longer need the anaphoric variants.

Dropping `dash.el` is not high on my to-do list, but I'd prefer if we used standard libraries and reduced the number of external deps. Who knows, maybe some day projectile will end up in Emacs (or ELPA). :-)

tarsius:
The tests pass now. My implementations of two functions had bugs which I fixed. I had to adjust one test. Before making my changes I adjusted that same test to make it more obvious what is being tested.

After doing that I have also cleaned up the tests. The intention was to remove the dependency on `f`, but I did run into a huge mess, which had to cleaned up at the same time. Most importantly the tests no longer define any prefix-less variables and functions. Instead there are two new macros, which actually get the job done.

tarsius:
You might want to look into why `projectile-parse-dirconfig-file` that the keeped files are directory names, while the same is not done for ignored files. Seems rather arbitrary.

bbatsov:
> The tests pass now. My implementations of two functions had bugs which I fixed. I had to adjust one test. Before making my changes I adjusted that same test to make it more obvious what is being tested.
> 
> After doing that I have also cleaned up the tests. The intention was to remove the dependency on f, but I did run into a huge mess, which had to cleaned up at the same time. Most importantly the tests no longer define any prefix-less variables and functions. Instead there are two new macros, which actually get the job done.

Good work!

> You might want to look into why projectile-parse-dirconfig-file that the keeped files are directory names, while the same is not done for ignored files. Seems rather arbitrary.

Likely an oversight. Unfortunately I haven't been paying Projectile lately as much attention as I should.

PR #626, opened 2015-02-01
projectile-find-other-file sort candidates

LefterisJP:
In projects where there are multiple file names with the same
header/source combinations we can still provide more accurate results by
sorting the candidates list to put the other-file under the same
directory name as this file to the top of the list.

So now if we have a directory tree like this:
```
|-src-
     |
     moduleA-
            |
            filename.c
     moduleB-
            |
            filename.c
|-include-
         |
         moduleA-
                |
                filename.h
          moduleB-
                 |
                 filename.h
```

`projectile-find-other-file` will always return the proper other file and there is no need to check the return list and choose which one it is that you actually wanted.

I tested this in the projects I am working on and it seems to work perfectly.

Note: In the sorting of the list the `other-file` is not used and as such there is a warning when you byte compile. So I suppose there is gotta be a better function to accomplish this. Any hints are more than welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ to behave like `helm-find-files`, such as multifile selection and opening or del
 
 ### Changes
 
+* get-other-files returns more accurate results for files with the same name placed under different directories
 * Collect search tool (`grep`, `ag`, `ack`) keybindings under a common keymap prefix (`C-c p s`)
 * Remove `defcustom` `projectile-remember-window-configs` in favor of
   `persp-projectile.el`.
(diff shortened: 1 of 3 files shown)


LefterisJP:
I saw a test fails. Will have to check out why.

bbatsov:
You'll have to update your code as we recently removed `f.el` from projectile. Unused variables should be prefixed with `_`. Add a few unit tests, squash the commits together & update the changelog.

LefterisJP:
Thanks for the reply. Will try to take care of the requested changes as soon as I get some free time and update the PR.

bbatsov:
@LefterisJP ping :-)

LefterisJP:
done

bbatsov:
:+1:

Issue #634, opened 2015-02-04
Document that helm-projectile-fuzzy-match must be set before loading helm-projectile


tsdh:
I don't like fuzzy matching, so I wanted to set `helm-projectile-fuzzy-match` to nil.  So I added

````elisp
(require 'helm-projectile)
(setq helm-projectile-fuzzy-match nil)
;; more helm-projectile settings
````

to my ~/.emacs.  Unfortunately, that didn't work.  The problem is that the value has to be set before the helm sources are defined.  Thus, one needs to set it before helm-projectile is required.

It would be good if the docstring of `helm-projectile-fuzzy-match` would mention this.

PR #635, opened 2015-02-04
Improve helm-projectile-fuzzy-match docstring

tsdh:
Document that this variable needs to be set before helm-projectile is
loaded.  (Fixes bug#634)


tsdh:
This fixes issue #634.

tuhdo:
I intended to fix this issue a while ago, so that source can be dynamically generated with latest user setting. But this would be a good workaround before I have time to finally do that.

PR #636, opened 2015-02-04
Issue 579: Add .projectile to the search of VCS roots. 

smolyn:
Stop searching when we come across the .projectile.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1509,6 +1509,7 @@ PROJECT-ROOT is the targeted directory. If nil, use
    ((projectile-file-exists-p (expand-file-name ".bzr" project-root)) 'bzr)
    ((projectile-file-exists-p (expand-file-name "_darcs" project-root)) 'darcs)
    ((projectile-file-exists-p (expand-file-name ".svn" project-root)) 'svn)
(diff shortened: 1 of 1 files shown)


bbatsov:
I haven't looked at the problem this is supposed to solve, but the solution doesn't seem right to me - this is not a `vcs` root and shouldn't be treated as one.

smolyn:
In a scenario where you have 
a/
   .git/
   b/
   b/.projectile

If you try to open a project in a/b/, it instead thinks it's a .git.  It then attempts to use .git VCS to determine the files, however if your a/b/ is not under git control, you get the empty set.

This is the fix I've used to make that work on my system, there may be a better one, but I don't have time to debug further than that.

smolyn:
Any updates?

Let me rewrite the scenario, the formatting above seems wrong:
a/.git
a/b/
a/b/.projectile

If I try to open the project as a/b/, it thinks .git is the VCS and ignores the .projectile.  This is incorrect since b/ isn't under git control.  So all of b/ does not get indexed.

aspiers:
This sounds like a genuine problem, although I can't comment on the validity of the proposed fix.

Issue #633, opened 2015-02-03
Presence of .projectile file breaks getting the list of files


avsej:
As @bbatsov said, things worked before https://github.com/bbatsov/projectile/commit/5e46193ed2c34028c7ee1ff65c5d2a16149131fe

rejeep:
Same here. Also breaks Magit commit.

jodonnell:
+1

IvanMalison:
Is anyone working on a fix?

bogus34:
Looks like I've managet to got it to work.
Here is my fix:

```diff
--- projectile.el-old	2015-02-04 19:26:15.769116419 +0300
+++ projectile.el	2015-02-04 19:25:50.196045381 +0300
@@ -1027,9 +1027,9 @@
 
 (defun projectile-paths-to-ignore ()
   "Return a list of ignored project paths."
-  (--mapcat (and (string-prefix-p "/" it)
-                 (directory-file-name it))
-            (cdr (projectile-parse-dirconfig-file))))
+  (-non-nil (--map (and (string-prefix-p "/" it)
+                (directory-file-name it))
+           (cdr (projectile-parse-dirconfig-file)))))
 
 (defun projectile-patterns-to-ignore ()
   "Return a list of relative file patterns."
```

mbme:
@bogus34 thank you, it works :+1:

jodonnell:
@bogus34 works like a charm

bogus34:
unfortunalety as @bbatsov mentioned in #639 file ignoring is broken even with this fix

bogus34:
new shot

```diff
diff -ur b/projectile.el a/projectile.el
--- b/projectile.el	2015-02-06 15:56:14.456497423 +0300
+++ a/projectile.el	2015-02-06 15:55:22.690639841 +0300
@@ -725,9 +726,12 @@
   "Get the files for ROOT under DIRECTORY using external tools."
   (let ((default-directory directory)
         (files-list nil))
-    (setq files-list (-map (lambda (f)
-                             (file-relative-name (expand-file-name f directory) root))
-                           (projectile-get-repo-files)))
+    (setq files-list (-non-nil (-map (lambda (f)
+                               (let ((file-name-expanded (expand-file-name f directory)))
+                                 (and
+                                  (not (projectile-ignored-rel-p file-name-expanded directory (projectile-patterns-to-ignore)))
+                                  (file-relative-name file-name-expanded root))))
[...]

PR #639, opened 2015-02-05
filter nil elements from projetile-paths-to-ignore

cowboyd:
This applies the fix suggested by @bogus34 in #633

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1027,7 +1027,7 @@ Only buffers not visible in windows are returned."
 (defun projectile-paths-to-ignore ()
   "Return a list of ignored project paths."
   (--mapcat (and (string-prefix-p "/" it)
-                 (directory-file-name it))
+                 (list (directory-file-name it)))
             (cdr (projectile-parse-dirconfig-file))))
 
 (defun projectile-patterns-to-ignore ()


projectile.el
```diff
@@ -1026,9 +1026,9 @@ Only buffers not visible in windows are returned."
 
 (defun projectile-paths-to-ignore ()
   "Return a list of ignored project paths."
-  (--mapcat (and (string-prefix-p "/" it)
-                 (directory-file-name it))
-            (cdr (projectile-parse-dirconfig-file))))
+  (-non-nil (--map (and (string-prefix-p "/" it)
+                     (directory-file-name it))
```
bbatsov: This is not indented properly.

bbatsov:
I've tried this locally and ignoring still doesn't work. The error is gone, but nothing is actually ignored...

projectile.el
cowboyd: hmm.. That's weird. I have no customizations on elisp or lisp indentation, although I did recently update prelude. 

Do you know offhand where these settings are? I can't seem to find any customizations for elisp indentation.

cowboyd:
I also noticed this behavior last night. Apologies for not verifying before submitting the request.

projectile.el
cowboyd: nvm. when I never re-indented after copying the patch from @bogus34

cowboyd:
```
-mapcat is a compiled Lisp function in `dash.el'.

(-mapcat FN LIST)

Return the concatenation of the result of mapping FN over LIST.
Thus function FN should return a list.
```

Just wrapping the directory name in the mapping function in a list does appear to work. If this seems like a reasonable approach, I can wrap this with a test to ensure the correct behavior.

cowboyd:
I believe this is closed by  41424b9fe2f1e12c3f2abbbad3d866ec23bd995d

PR #656, opened 2015-02-21
Fix `projectile-regenerate-tags` in dirs with spaces

ryanartecona:
Just like it says in the title! :sparkles:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ buffers by name
 
 ### Bugs fixed
 
+* Fix `projectile-regenerate-tags` to work in directories that include spaces.
 * Prevent `projectile-kill-buffers` from trying to kill indirect
 buffers.
 * [#412](https://github.com/bbatsov/projectile/issues/412): Handle multiple possible targets in `projectile-toggle-between-implementation-or-test`.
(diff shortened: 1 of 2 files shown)


bbatsov:
:+1:

dgoel:
With this change, the excluded tags patterns (for example, those mentioned in .projectile) are no longer considered. The correct fix is the following:

```
--- a/projectile.el
+++ b/projectile.el
@@ -170,7 +170,7 @@ Otherwise consider the current directory the project root."
   :group 'projectile
   :type 'string)
 
-(defcustom projectile-tags-command "ctags -Re -f \"%s\" \"%s\""
+(defcustom projectile-tags-command "ctags -Re -f \"%s\" %s"
   "The command Projectile's going to use to generate a TAGS file."
   :group 'projectile
   :type 'string)
@@ -1734,7 +1734,7 @@ regular expression."
 
 (defun projectile-tags-exclude-patterns ()
   "Return a string with exclude patterns for ctags."
-  (mapconcat (lambda (pattern) (format "--exclude=%s"
+  (mapconcat (lambda (pattern) (format "--exclude=\"%s\""
              (projectile-ignored-directories-rel) " "))

```

bbatsov:
@dgoel File a PR with the fix.

PR #660, opened 2015-02-23
Fix generation of command for tag generation

dgoel:
Previously, command line that was being generated was this:
ctags -Re -f TAGS "--exclude=<pattern> .... "
Due to the inclusion of quotes, the exclude pattern is completely
ignored by ctags.

This commit fixes this issue and generates the command line as follows:
ctags -Re -f TAGS --exclude="<pattern>" ...
This does not break the exclude option while still works with spaces in
filenames.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -170,7 +170,7 @@ Otherwise consider the current directory the project root."
   :group 'projectile
   :type 'string)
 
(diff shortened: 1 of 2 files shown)


bbatsov:
:+1:

PR #662, opened 2015-02-25
Relative path is better than absolute path

syohex:
See also
- https://help.github.com/articles/relative-links-in-readmes/


bbatsov:
:+1:

PR #669, opened 2015-03-01
Generalized projectile-project-root to take optional argument

emacs18:
This is an alternate fix for issue 570 which requested for a function that takes file name as argument which returns project root directory.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -654,11 +654,15 @@ Returns a project root directory path or nil if not found."
    nil
    (or list projectile-project-root-files-top-down-recurring)))
 
-(defun projectile-project-root ()
-  "Retrieves the root directory of a project if available.
-The current directory is assumed to be the project's root otherwise."
+(defun projectile-project-root (&optional path)
+  "Retrieve the project root directory for PATH which defaults to current
+directory.  Return directory of PATH if project root cannot be found."
+  (if path
+      (unless (file-directory-p path)
+        (setq path (file-name-directory path))) ; ensure PATH is a directory
+    (setq path default-directory))
   (file-truename
-   (let ((dir (file-truename default-directory)))
+   (let ((dir (file-truename path)))
      (or (--reduce-from
           (or acc
               (let* ((cache-key (format "%s-%s" it dir))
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm still not sure we need such a function at all, although this solution is reasonable.

IvanMalison:
@bbatsov Why are you so reluctant to include this? I would find it to be useful for some code that I have written at https://github.com/IvanMalison/org-projectile where I have actually duplicated the function definition because projectile does not support this feature.

Maybe no one else will use this, but what is the downside? I'd rather not maintain a separate definition of `projectile-project-root` in case it changes in any way.

bbatsov:
I'm reluctant to include anything that I'll have to maintain, but I'm unlikely to ever use. :-) 

At any rate - I'll accept this change provided:

* unit tests are added
* a changelog entry is added
* my comments are addressed

projectile.el
```diff
@@ -654,11 +654,15 @@ Returns a project root directory path or nil if not found."
    nil
    (or list projectile-project-root-files-top-down-recurring)))
 
-(defun projectile-project-root ()
-  "Retrieves the root directory of a project if available.
-The current directory is assumed to be the project's root otherwise."
+(defun projectile-project-root (&optional path)
+  "Retrieve the project root directory for PATH which defaults to current
```
bbatsov: `M-x checkdoc`

projectile.el
```diff
@@ -654,11 +654,15 @@ Returns a project root directory path or nil if not found."
...
+directory.  Return directory of PATH if project root cannot be found."
+  (if path
```
bbatsov: I don't like the name `path` - maybe `file-path` or something like this.

IvanMalison:
@emacs18 Can I take over this pull request to make the requested changes?

emacs18:
IvanMalison <notifications@github.com> writes:

> @emacs18 Can I take over this pull request to make the requested changes?

Yes, please!

I got involved only because Bozhidar lamented that there were 72 open issues.
I was just trying to help out.

bbatsov:
@IvanMalison Ping :-)

bbatsov:
@IvanMalison ping 2

IvanMalison:
I'll take a look at this today

bbatsov:

PR #674, opened 2015-03-04
[Fix #672] Don't quit when build Helm source

### LINKED ISSUE #672: helm-projectile-find-file-dwim doesn't work in the 1 candidate case
When using `helm-projectile-find-file-dwim` such that the filename at point has only one possible match, the following happens:

The matched file is opened in one buffer, but helm still prompts me for a filename with an empty \*helm projectile\* buffer open. Since that buffer is empty, I can't select any canditate and C-g jumps back to the file I started `helm-projectile-find-file-dwim` in.

tuhdo:
Instead, only build Helm dwim source when applicable. That is, when the
number of candidates are greater than 1 or no candidate found and the
whole project files are displayed as usual projectile-find-file.

This change also replaces static Helm source with dynamically generated
helm DWIM source. In the future, Helm sources should be changed like
this to adapt to user settings easily. Currently, user has to set fuzzy
matching before loading Helm Projectile, which is inconvenient.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -377,27 +377,18 @@ CANDIDATE is the selected file.  Used when no file is explicitly marked."
      . "Add files to Dired buffer `C-c a'"))
   "Action for files.")
 
-(defvar helm-source-projectile-files-dwim-list
+(defun helm-projectile-build-dwim-source (candidates)
+  "Dynamically build a Helm source definition for Projectile files based on context with CANDIDATES."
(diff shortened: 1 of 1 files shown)


tuhdo:
@bbatsov ping.

tuhdo:
@bbatsov Is there any problem?

bbatsov:
No problem. I'm just really short on free time...

PR #681, opened 2015-03-08
Fix: use helm-buffers-match-function rather than helm-buffers-list--matc...

duncanmortimer:
...h-fn

Commit 6613f8 on helm-emacs/helm replaced helm-buffers-list--match-fn
with helm-buffers-match-function, breaking helm-projectile.  This commit
updates helm-projectile to use the new function.


tuhdo:
@bbatsov Please merge this commit. People are reporting everywhere. Thanks.

bbatsov:
:+1:

PR #685, opened 2015-03-10
fix inversion in projectile-files-with-string

fommil:
`-L` (capital L) shows NON-matching files `-l` is what we want: matching ones. This is true for grep and git-grep .


bbatsov:
:+1:

PR #688, opened 2015-03-16
Add ivy to `projectile-completion-system'

abo-abo:
* projectile.el (projectile-completion-system): Update.
(projectile-completing-read): Update.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -150,6 +150,7 @@ Otherwise consider the current directory the project root."
           (const :tag "Ido" ido)
           (const :tag "Grizzl" grizzl)
           (const :tag "Helm" helm)
+          (const :tag "Ivy" ivy)
           (const :tag "Default" default)
           (function :tag "Custom function")))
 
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -1116,6 +1117,12 @@ https://github.com/emacs-helm/helm")))
           (grizzl-completing-read prompt (grizzl-make-index choices))
         (user-error "Please install grizzl from \
 https://github.com/d11wtq/grizzl")))
+     ((eq projectile-completion-system 'ivy)
+      (if (fboundp 'ivy-read)
+          (ivy-read prompt (cl-delete-duplicates choices
+                                                 :test #'equal) initial-input)
+        (user-error "Please install ivy from \
+https://github.com/abo-abo/swiper")))
```
bbatsov: Why don't you spin ivy as a separate project?

abo-abo: > Why don't you spin ivy as a separate project?

It's separate in MELPA. But since I'm still figuring out the responsibilities for each package, they
are in the same repository (easier to refactor that way).

PR #690, opened 2015-03-21
Remove the uniqueness enforcement for `ivy-read'

abo-abo:
* projectile.el (projectile-completing-read): Update.
`cl-delete-duplicates' for thousands of candidates is just too slow.
Anyway, the choices are supposed to be unique in the first place.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1119,8 +1119,7 @@ https://github.com/emacs-helm/helm")))
 https://github.com/d11wtq/grizzl")))
      ((eq projectile-completion-system 'ivy)
       (if (fboundp 'ivy-read)
-          (ivy-read prompt (cl-delete-duplicates choices
-                                                 :test #'equal) initial-input)
+          (ivy-read prompt choices initial-input)
         (user-error "Please install ivy from \
 https://github.com/abo-abo/swiper")))
      (t (funcall projectile-completion-system prompt choices)))))


abo-abo:
Did you ever get an issue for `projectile' offering the same file
multiple times in the list?

tuhdo:
> Did you ever get an issue for `projectile' offering the same file
> multiple times in the list?

No I never. Projectile only retrieves the file list from external commands, so it should stay as is. How to reproduce it? You should test it with other completion system like `ido` or `helm` to see if the problem is there as well.

abo-abo:
> You should test it with other completion system like ido or helm to see if the problem is there as well.

Did you look at the code? The `choices` var is generated by `projectile` and is given in the same form to all completion packages (except maybe `grizzl`). And I did get a non-unique `choices` list at one point, that's why I've added `cl-remove-duplicates`, because `ivy-read` doesn't handle non-unique candidates well.

While other completion systems might handle the non-unique candidates, I still think that for projectile they have no business being non-unique in the first place.

tuhdo:
But then, Projectile only presents users with the file list from external command, as you can see [here](https://github.com/bbatsov/projectile/blob/master/projectile.el#L838) and then perform filtering on it based on the patterns defined in `.projectile`. So, if the external command gives duplicate candidates, then it's not Projectile's fault. Both `helm-projectile-find-file` and stock `projectile-find-file` with Ido have no such problem when I use it on projects with different VCS. Or are you using plain `find` to get project file list?

abo-abo:
> Or are you using plain find to get project file list?

I have this:

```cl
(setq projectile-indexing-method 'alien)
(setq projectile-enable-caching t)
```

I don't know if `find` is involved.

PR #691, opened 2015-03-22
file-truename when caching current file. Closes #667

### LINKED ISSUE #667: Symlinks pointing to directories inside a projectile project are added to projectile files cache
Sorry for the obscure/confusing title. I'm finding it difficult to accurately articulate exactly what the issue is in an abstract way, so let me provide an example to clarify:

I have a source controlled dotfiles project that lives in ~/dotfiles. Everything inside of ~/dotfiles/dotfiles, including, for example, ~/dotfiles/dotfiles/emacs.d, is pointed at by a symlink in ~ with an identical name (though prefixed with a dot). If a file is opened through the symlinked path, projectile correctly recognizes that the file is part of the ~/dotfiles project, but it then adds the relative path of the file through the symlink (rather than inside of the project) to the project's filepath cache. This means that the file cache gets populated with things like

../.emacs.d/elpa/projectile-20150223.8/projectile.el

Now that I am investigating a bit more, it seems that this only happens with files that are ignored by projectile in the actual directory. It seems that what might be happening is that projectile is not recognizing that the relative filepath ultimately links to a path that is ignored. It might be difficult to come up with a proper resoultion to this issue since you might not always WANT to ignore symlinks that link to other ignored files. However, I would make the argument that any filepath that is outside of the main project should never be added to the projectile file cache.

Sorry if any of this is confusing; let me know if there is anything that needs elucidation.

IvanMalison:


bbatsov:
Change to commit message to "[Fix #667] Use file-truename ...".

bbatsov:
Mention the bugfix in the changelog and remove the `.` from the commit message (commit message titles are not sentences).

CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ to behave like `helm-find-files`, such as multifile selection and opening or del
   buffers by name
 * New `defcustom` `projectile-globally-ignored-file-suffixes` allows
   you to globally ignore files with particular extensions
+* Use file-truename when caching filenames to prevent
```
bbatsov: This should be in the `Bugs fixed` section of the `master`.

CHANGELOG.md
```diff
@@ -45,6 +45,9 @@ to behave like `helm-find-files`, such as multifile selection and opening or del
 * Prevent `projectile-kill-buffers` from trying to kill indirect
 buffers.
 * [#412](https://github.com/bbatsov/projectile/issues/412): Handle multiple possible targets in `projectile-toggle-between-implementation-or-test`.
+* Use `file-truename` when caching filenames to prevent
```
bbatsov: The place of this entry has to be updated.

IvanMalison: What? Is it not currently in the right place?

bbatsov: 0.12 was already released. See the latest version of the changelog and you'll understand what I mean.

CHANGELOG.md
IvanMalison: Okay, so really, what you mean is that this needs to be merged with master...

bbatsov: Yes.

On Saturday, April 25, 2015, IvanMalison <notifications@github.com> wrote:

> In CHANGELOG.md
> <https://github.com/bbatsov/projectile/pull/691#discussion_r29098818>:
>
> > @@ -30,6 +30,9 @@ to behave like `helm-find-files`, such as multifile selection and opening or del
> >    buffers by name
> >  * New `defcustom` `projectile-globally-ignored-file-suffixes` allows
> >    you to globally ignore files with particular extensions
> > +* Use file-truename when caching filenames to prevent
>
> Okay, so really, what you mean is that this needs to be merged with
> master...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bbatsov/projectile/pull/691/files#r29098818>.
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

bbatsov:
@IvanMalison ping

IvanMalison:
I can take a look at this soon

bbatsov:
//cc @vspinu

vspinu:
I think this patch indeed fixes a bug. As project root's name is always true name. Ths all files expanded against it should be truename as well. Also the defadvice on `delete-file` uses a similar logic to `projectile-cache-current-file` and it already calls `file-truename`.

This function is not critical in any sense. It's called either interactively or from the `file-find-hook` and only when `projectile-enable-caching` is on, which it's nil by default.

projectile.el
```diff
@@ -548,7 +548,7 @@ The cache is created both in memory and on the hard drive."
   "Add the currently visited file to the cache."
   (interactive)
   (let* ((current-project (projectile-project-root))
-         (abs-current-file (buffer-file-name (current-buffer)))
+         (abs-current-file (file-truename (buffer-file-name (current-buffer))))
```
vspinu: Could you please move these last two bindings inside the following `(when `.  There is no point for them unless that conditional is true.

Also remove the double call to `projectile-project-root` on this occasion.

IvanMalison:
status?

bbatsov:
:+1:

vspinu:
> In that case, the truename is horrendous

That's a relative name, not true name.  From what you have said, I don't understand what's the exact problem nor how the new behavior "is a  bug".

kini:
FWIW, even if you made that an absolute path, it would look like `/foo/bar/sos_cache/prj_name/prj_name_sos_cache/prj_name/FILES/dir1#dir1_1#dir1_1_1#file_name_v_1.23_456789`, which is still horrendous :)

PR #700, opened 2015-04-02
Don't forcefully insert symbol at point for searching

tuhdo:
Currently, helm-projectile-grep/ack/ag inserts symbol at point by
default. It could be annoying and we should let users do it manually by
simply pressing `M-n`.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -676,7 +676,7 @@ If it is nil, or ack/ack-grep not found then use default grep command."
      :sources 'helm-source-grep
      :input (if (region-active-p)
                 (buffer-substring-no-properties (region-beginning) (region-end))
-              (thing-at-point 'symbol))
+              "")
      :buffer (format "*helm %s*" (if use-ack-p
                                      "ack"
                                    "grep"))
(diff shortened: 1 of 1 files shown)


tuhdo:
@bbatsov and merge this as well.

bbatsov:
I don't see the problem with the current behaviour - a lot of commands default to the symbol at point.

tuhdo:
Well, in many cases it gets in the way. That is, you don't want to search for symbol at point but for random things in your head. It's easy with Helm to insert symbol at point with `M-n`, while inserting symbol at point making you have to `C-a` then `C-k`. And `helm-projectile-ag` can use `helm-ag-insert-at-point` to customize.

But, I will add customize variables for ack/grep to insert at point or not.

trishume:
Just weighing in here that 90% of the time I use search it is to get somewhere that I am not so the symbol is just some random thing my cursor happened to be on. I'm also an Evil user that doesn't know any Emacs commands so I end up having to backspace out the symbol every time I search.

tuhdo:
@trishume you can press 'M-<backspace>' to delete a whole symbol.

tuhdo:
I mean 'M-backspace'.

tuhdo:
Since @trishume PR is merged, this can be closed. Good job!

Issue #657, opened 2015-02-22
Projectile causes emacs lag on remote filesystems


barrkel:
I'm running projectile-20150216.843 from melpa.

Cursor movement and scrolling in particular is incredibly slow when editing files on locally mounted high latency remote filesystems.

Using system monitoring tools, I can see that the file system is accessed on every keystroke. If the filesystem doesn't respond quickly, emacs lags. Remote filesystems don't necessarily respond quickly; this makes emacs lag.

When concurrent activity uses remote filesystem bandwidth, emacs effectively freezes.

I previously reported this as #448. I've just upgraded; it's still happening.

I can verify that changing the mode line variable to a constant string removes the lag:

    (setq projectile-mode-line "foo")

With this set, everything becomes smooth again.

Projectile is pretty much unusable for me in its current state.

I simply don't see why the projectile mode line needs to be updated for every cursor move. The project a file belongs to almost never changes after being opened.

bbatsov:
> Projectile is pretty much unusable for me in its current state.

> I simply don't see why the projectile mode line needs to be updated for every cursor move. The project a file belongs to almost never changes after being opened.

Originally the modeline was updated just on file opening, but I opted to to remove some custom code in favour of the standard Emacs tooling, which sadly updates the modeline on each keystroke. Changing this back would be trivial or we can simply disable it for remote files. I understand your frustration but for me this is a tiny problem with known workaround in one of the dozen of projects I maintain (meaning I can't deal with every problem in a prompt manner).

Silex:
I was very close to make a PR reverting the `:lighter` change, but it feels "wrong". The more correct thing to do is to refactor the cache system once and for all so it caches the project root correctly instead of the hacks we have in place at the moment.

Unfortunately, it seems that some of the problems brought by `:lighter` are simply not solvable, see #523.

So maybe it's better to revert back to the "file opening" method?

vspinu:
The problem seems to come from the use of [`file-truename`](https://github.com/bbatsov/projectile/blob/master/projectile.el#L658). `file-truename` is very slow because it basically checks for simlinks at each level of the directory. When I remove both instances all the remote problems seem to vanish. I am moving inline for further discussion.

behrica:
I am not sure, if this bug report is the correct place or it should be a new bug.

I noticed still with latest projectile a slide lag every second or so, while scrolling down/up in a  directory which is on a mounted webdrive under linux.

[...]

PR #701, opened 2015-04-02
Don't show Projectile directory when not in project

tuhdo:
Currently, Projectile shows any directory it's in. This is not useful
since we don't know whether or not we are in a project directory. This
change makes it only shows project root in modeline when we are actually
in a project, so we know whether we are in a Project or not when
entering a buffer, and also save screen estate.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -694,8 +694,10 @@ A thin wrapper around `file-truename' that handles nil."
   (let ((project-root
          (condition-case nil
              (projectile-project-root)
-           (error default-directory))))
-    (file-name-nondirectory (directory-file-name project-root))))
+           (error nil))))
+    (if project-root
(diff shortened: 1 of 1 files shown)


bbatsov:
This probably has to be changed in `projectile-project-name` - guess it can return something like `~` when not in a project dir. As this is invoked on every keystroke we definitely don't want calling more functions than necessary.

bbatsov:
There's also this to consider https://github.com/bbatsov/projectile/issues/657 - probably we should modify the lighter in a find-file-hook to avoid it...

bbatsov:
@tuhdo ping

tuhdo:
@bbatsov Sorry I updated my PR.

tuhdo:
@bbatsov This as well.

bbatsov:
:+1: Thanks!

PR #703, opened 2015-04-03
Fix perspective switching behavior

501st-alpha1:
Perspective should only be changed if user is actually in a project directory.


bbatsov:
:+1: Thanks!

arichiardi:
:+1:

PR #710, opened 2015-04-12
Set default-directory before compilation command

kyleam:
Set `default-directory` before reading the compilation command so that
`completion-at-point` offers appropriate targets even when
`projectile-compile-project` is called from a subdirectory.  Also,
remove unnecessary check on DIR argument.


bbatsov:
:+1: Thanks!

PR #713, opened 2015-04-16
Update the call to `ivy-read'

abo-abo:
projectile.el (projectile-completing-read): Update the call to
`ivy-read'


bbatsov:
:+1:

PR #714, opened 2015-04-16
Don't ignore user preference in helm-projectile-ag

trishume:
`helm-ag-insert-at-point` is a user preference and shouldn't be overriden by helm-projectile. The way it worked before this made it nearly impossible for the user to choose not to have the field auto-filled. If users want the field auto-filled they can set the helm-ag config variable.


bbatsov:
Makes sense. Thanks!

PR #716, opened 2015-04-17
Support detecting project type from composer.json

marcinant:
[Composer](http://getcomposer.org/) supports multiple project [installers](https://github.com/composer/installers).

This PR detects Composer project type defined in composer.json.
Detected type can appear as part of project description shown by `projectile-project-info` or can be used to run custom commands associated with project type.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -40,6 +40,7 @@
 (require 'dash)
 (require 'ibuffer)
 (require 'ibuf-ext)
+(require 'json)
 
 (eval-when-compile
   (defvar ggtags-completion-table)
@@ -1490,6 +1491,17 @@ With a prefix ARG invalidates the cache first."
   :group 'projectile
   :type 'function)
 
+(defun projectile-composer ()
+  "Function to determine project type from composer.json."
+  (when (projectile-verify-file "composer.json")
+    (let ((composer-json (json-read-file (projectile-expand-root "composer.json"))))
+      (cdr (assoc 'type composer-json)))))
+
+(defcustom projectile-composer-function 'projectile-composer
+  "Function to determine Composer project type."
+  :group 'projectile
+  :type 'function)
+
 (defun projectile-project-type ()
   "Determine the project's type based on its structure."
   (cond
@@ -1514,6 +1526,7 @@ With a prefix ARG invalidates the cache first."
    ((projectile-verify-files projectile-haskell-cabal) 'haskell-cabal)
    ((projectile-verify-files projectile-rust-cargo) 'rust-cargo)
    ((projectile-verify-files projectile-r) 'r)
+   ((funcall projectile-composer-function) (funcall projectile-composer-function))
    ((funcall projectile-go-function) 'go)
    (t 'generic)))
 


bbatsov:
Same issue as with the other PR - how do you differentiate between an actual npm project and some other project (e.g. a Rails project having those inside - which an actually situation from projects I've worked on).

marcinant:
Simple answer is: I don't.

However it's not my fault. It's because `projectile` is sort of `broken by design` ;)
Thing is that there are `hybrid projects`. You can have PHP project with `composer.json` to manage PHP dependencies and `bower.json` to get scss assets and `package.json` to manage some js.

Projectile does not support [project natures](http://help.eclipse.org/luna/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2FresAdv_natures.htm) just like Eclipse.

So, you have a problem here. Projectile supports only one `compile` command for project and only single `test` command for project. While project can contain `composer` scripts to test and for example `gulp` or `grunt` to build.

You have a choice. You can deal with it and accept this PR while you know it's not perfect. And then you can expect that developers will use it properly. Or you can redesign `projectile` to support multiple scenarios and combinations just like Eclipse does (a lot of work).

bbatsov:
> You have a choice. You can deal with it and accept this PR while you know it's not perfect. And then you can expect that developers will use it properly. Or you can redesign projectile to support multiple scenarios and combinations just like Eclipse does (a lot of work).

I can also do nothing and let others do the work for me. :-) 

I agree that the current support for projects is pretty simplistic and if I had any time to work on this I'd happily do it. Alas, I do not. There are few things that should be done:

* project types should be defined using some function instead of using a ton of variables and a bit switch. 
* all the types a project matches should be collected together - if a project matches several types, then the commands like compile should ask the use which project type to use

This is not a lot of work actually and in due time I might actually do it. Until then - anyone is welcome to take a stab at it.

bbatsov:
Btw, have a look at my comment on #758. When we can reliably auto-detect the project type, setting it explicitly is probably the way to go. So we can just have many project types without auto-detection, which we can set by `.dir-locals.el`.

marcinant:
Personally I'm just overriding `projectile-project-type` right now.
I run my `composer-project-type` function and if it returns `nil` then fallback to `projectile-project-type`.

However I think that this variable should be list. Because this still doesn't resolve problem when some project can be of multiple types.

arichiardi:
Let's write ```projectile-project-type``` ideas on the other PR or I am going to miss them :smile: 
I am planning to work on this addition and would like to discuss the possible use cases...see you there!

arichiardi:
@marcinant 

Check #758, does it solve your issue? You can now define ```(projectile-project-type . (projectile-composer))``` in your ```.dir-locals.el``` and see what happens :smile:

marcinant:
@arichiardi  Sort of. However I still don't like it and I'm not going to use `.dir-locals.el`.

This solution is a nice workaround unfortunately it's not portable. It relies on external custom local function `(projectile-composer)` which can be missing and not available for some users.

It's a step in good direction. However if there is a better and complete solution which can be implemented into `projectile.el` then I see no point in using `.dir-locals.el`.

Personally I got extensions to projectile implemented in my `init.el` and I'm going to switch them into external package.

arichiardi:
Good, yes I guess now you have some sort of ```hook``` that your project can rely upon in order to identify it within projectile.
Projectile's team decision aside, even Eclipse does not explicitly support all the possible ```natures```, but it allows you to plug-in what you need. I guess this is what to strive for, provide defaults plus ways to customize. Don't you agree?

bbatsov:
Closing due to lack of activity. Refactoring the notion of project type is essential for something like this to be supported.

PR #717, opened 2015-04-18
Support NPM packages

marcinant:
Support for [NPM JS](https://www.npmjs.com/) packages.

Detect NPM package if there is 'package.json' file and 'node_modules' dir.

Run 'npm test' and 'npm start' scripts.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -40,7 +40,6 @@
 (require 'dash)
 (require 'ibuffer)
 (require 'ibuf-ext)
-(require 'json)
 
 (eval-when-compile
   (defvar ggtags-completion-table)
@@ -1481,7 +1480,7 @@ With a prefix ARG invalidates the cache first."
 (defvar projectile-haskell-cabal '("*.cabal"))
 (defvar projectile-rust-cargo '("Cargo.toml"))
 (defvar projectile-r '("DESCRIPTION"))
-(defvar projectile-npm '("package.json" "node_modules"))
+(defvar projectile-npm '("package.json"))
 
 (defun projectile-go ()
   (-any? (lambda (file)
--- a/projectile.el
+++ b/projectile.el
@@ -2015,6 +2015,7 @@ For git projects `magit-status-internal' is used if available."
 (defvar projectile-r-compile-cmd "R CMD INSTALL .")
 (defvar projectile-r-test-cmd (concat "R CMD check -o "
                                       temporary-file-directory " ."))
+(defvar projectile-npm-compile-cmd "npm start")
 (defvar projectile-npm-test-cmd "npm test")
 
 (--each '(projectile-rails-compile-cmd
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -40,6 +40,7 @@
 (require 'dash)
 (require 'ibuffer)
 (require 'ibuf-ext)
+(require 'json)
```
bbatsov: This seems redundant.

marcinant:
Yes. Sorry. This is not required here. However if you accept my second PR then it will be ok.

projectile.el
```diff
@@ -1480,6 +1481,7 @@ With a prefix ARG invalidates the cache first."
 (defvar projectile-haskell-cabal '("*.cabal"))
 (defvar projectile-rust-cargo '("Cargo.toml"))
 (defvar projectile-r '("DESCRIPTION"))
+(defvar projectile-npm '("package.json" "node_modules"))
```
moonlite: Will this look for `package.json` *and* a `node_modules` dir? Your PR message seems to indicate that, if so I think it should rather be an *or* instead, since if you clone an npm module from, say, GitHub it won't have a node_modules directory until you issue an `npm install`.

marcinant: Well I have to agree but only partially.
Doc string for this function states: 'Determine the project's type based on its structure.'
I know that 'package.json' is enough to detect project as NPM. So, no 'or', just '(defvar projectile-npm '("package.json"))' only. Thing is that 'node_modules' is hard coded and characteristic structural element for NPM projects and this is why I decided to add this directory. NPM projects without 'node_modules' are rare. 

If you think that 'package.json' is enough to detect project type as NPM then I'm not going to argue :)

Different kind of problem is that Projectile does not support multiple "project natures" as for example Eclipse does.

There is a lot of projects which have: 'bower.json', 'package.json' and 'grunt.js' or 'gulpfile.js' at the some time and it's pretty hard to find if they are NPM packages with bower support or bower packages with NPM used for dependencies or both.

moonlite: Yeah it's hard and annoying that they are all using `package.json`, however the node_modules directory is only available after the first run of npm install which means that if you clone a repo using magit, open it with projectile and expect to be able to "compile" with C-c p c you'll not get what you expect.

marcinant:
I just updated this PM with @moonlite suggestion.
Cleaned up unnecessary '(require 'json)'

@bbatsov  how long it takes to have PM reviewed and merged?

bbatsov:
Many non-node projects have a `package.json` file in them, so I don't think this is a sufficient check for them.

marcinant:
What kind of projects have `package.json` ? I don't know any...

If it's not sufficient then the only way is to read and validate json and detect project in this way (this is what I do for composer projects)

bbatsov:
Refactoring the notion of project type is essential for something like this to be supported.

azer:
@marcinant @bbatsov non-JavaScript projects usually have a subfolder with package.json in it. and recently, [even NPM projects began having multiple NPM packages inside](https://lernajs.io/)

I think this PR should be reverted.

PR #726, opened 2015-04-24
Add support for the tox testing automation tool

mineo:
Tox [0] is a tool that allows running tests in multiple Python
environments sequentially. This is especially useful when working on
projects supporting both Python 2 and 3 because the standard unittest
runner can only run against one Python version at a time.

[0] https://testrun.org/tox/latest/


bbatsov:
:+1: Thanks!

PR #502, opened 2014-10-31
Add tramp/network/latency benchmarking suite (WIP)

thomasf:
## What?

This is a benchmark system made up of a fig/docker combination.

## Motivation

Many projectile changes affects slow file systems like TRAMP.
We need something to do benchmark regression tests and a way for different people to run the same test to make reasoning about performance issues a lot easier.

## How it works

I have built this prototype using a fig/docker combination, everything is runnable through the Makefile.

It should start by first running `make fig-up` followed by `make fig-test`.

A forced full rebuild/run cycle can be triggered using fig: `fig build --no-cache && fig up`.

## The docker containers

There are four images and three running instances:

**Projectile-testdata**

This is a base image for all other images that needs the test data files. It's does not run by its own.

The content is currently a few shallow git checkouts, an emacs source release expanded and some empty directories.

**Projectile**

The tests are run from this instance, it is basically emacs + cask + the projectile repository + tools like scm and what else might be required for tests. It inherits from the projectile-testdata image.

**sshd**

Runs a sshd daemon that the projectile instance connects to. It inherits from the projectile-testdata image.

**polipo**

A caching http proxy to support quicker cask rebuilds from scratch.

### Top priority (Must be done before merge)

* Emit benchmark results as csv or similar and add some diagram creation script.
* Verify that the results are reliable for finding performance issues. (I think it's good, wait for diagrams to verify further)
* Evaluate if ert really is the right tool for running this suite. It might not be needed. 

### Medium priority (Should be done before merge)

* Add even more tests, try to cover the major areas of whats most often called.
* Sober everything up to contribution standards.
  
### Low priority (can be done after merge)

* Make it possible to run suite outside docker.
* Add complex high level tests that simulates a user working on code. Test cleanup might be a nut to sovle.
* Add sensible functions that does seletive instrumentation when results needs to be looked at in more detail.

PR #728, opened 2015-04-26
[Fix #657] Remove `file-truename` calls from `projectile-project-root`

### LINKED ISSUE #657: Projectile causes emacs lag on remote filesystems

vspinu:
I have no  global picture so feel free to close if it's nonsense.

`file-truename` is very slow on remotes because it has to check if each subdir is a simlink.  I think inner `file-truename` is not needed because you don't care what you use as key in your cache as long that key stays the same, right?  Outer `file-truename` is not needed either. It's the task of the  cacher to store the truename. Also, if you pass a truename to the `root-function` you need not truename it afterwards because guaranties to produce truename if it gets truename.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -621,7 +621,7 @@ which we're looking."
             ((equal file (setq file (file-name-directory
                                      (directory-file-name file))))
              (setq file nil))))
-    (if root (file-name-as-directory root))))
+    (and root (expand-file-name (file-name-as-directory root)))))
 
 (defun projectile-root-bottom-up (dir &optional list)
   "Identify a project root in DIR by looking at `projectile-project-root-files-bottom-up'.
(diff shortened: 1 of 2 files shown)


Silex:
You are correct, but see #502 first.

vspinu:
> You are correct, but see #502 first.

That's truly great, but how does it exactly connect to the patch?

 I know for sure that `(benchmark-run 1 (file-truename (buffer-file-name)))` runs 0.45s on my remote as opposed to 0.00005s on a local machine. That function is called twice in a super critical `projectile-project-root` 
 `projectile-project-root` is called like 6 times during an opening of a file and is also called on each redisplay. I haven't investigated this further but it left no doubt on my side that `file-truename` should be removed from there with any cost.

I have familiarized myself with the code a bit and I am pretty confident now what the proposed change is safe. All tests are also good.

Silex:
My point was that there are discussions about refactoring projectile itself. Anyway, I guess your PR doensn't hurt... but the `file-truename` was there for a reason iirc, maybe @bbatsov remembers?

vspinu:
> My point was that there are discussions about refactoring projectile itself.

Aha. I see. Looking forward to seeing the final result.

>  file-truename was there for a reason

Of course. It's still there, but only where it's needed - in the place where the actual caching is happening.

Silex:
>> My point was that there are discussions about refactoring projectile itself.
> Aha. I see. Looking forward to seeing the final result.

Well, given how swamped we all are unfortunaly that will probably take a while. If you want to make it move forward feel free :)
 
>> file-truename was there for a reason
> Of course. It's still there, but only where it's needed - in the place where the actual caching is happening.

Yeah, personally I think your patch is fine but I remember @bbatsov saying something about wanting to resolve symlinks before storing, so maybe `file-truename` is needed :(

Let's wait for his input.

bbatsov:
I'm reasonably sure this `file-truename` appeared here to fix some bug (maybe it was something about same project appearing twice in the known projects or something). I'll have to take a closer look at the patch, but I swamped with other work.

vspinu:
I have been using this for more than a  week now and it really solves all the problems. 

> I'm reasonably sure this file-truename appeared here to fix some bug (maybe it was something about same project appearing twice in the known projects or something).

It's still there, but only for the root path and only during caching.

bbatsov:
Let's merge this and see what will happen. :-)

PR #729, opened 2015-04-26
When cache cleared, user expects no old files

swaroopch:
Fixes https://github.com/syl20bnr/spacemacs/issues/1256


tuhdo:
@bbatsov this should be merged because this is a bug in Projectile: when you set `projectile-sort-order` to `recentf`. Projectile will then use recentf file list and select the files in project, then present those files first without checking for its existence. If you set `projectile-sort-order` to `'default`, this bug won't happen.

PR #730, opened 2015-04-27
Tests

vspinu:
Various improvements to the test suite. 

  - Make test macros edebug friendly
  - Don't span our emacs with dummy `*nrepl-messages*<i>` buffers.
  - Disable yes/no questions from `find-file-noselec` on repeated manual erts


bbatsov:
:+1: Thanks!

PR #731, opened 2015-04-28
Don't require helm-config

lunaryorn:
It's entirely redundant here, and we should not force a given set of preferences upon the user.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -40,7 +40,6 @@
 ;;; Code:
 
 (require 'projectile)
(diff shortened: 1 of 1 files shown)


lunaryorn:
Besides, that file is dire need of some checkdoc fixes…

bbatsov:
Thanks!

> Besides, that file is dire need of some checkdoc fixes…

I don't use helm, so I don't pay much attention to the style employed here. PRs are always welcome.

PR #735, opened 2015-05-07
Proposal: change projectile-mode-line to P[%s]

vspinu:
Projectile's mode line takes a quoter of my modeline on a left-right split.  I think `P[project-name]` is as suggestive as `Projectile[project-name]`.  I would expect most users changing it anyways, then why not change it by default?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * `projectile-compile-project` now offers appropriate completion
   targets even when called from a subdirectory.
+* `projectile-mode-line` was changed to P[%s] to save mode-line space.
 
 ## 0.12.0 (03/29/2015)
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Nobody has complained about this and it's configurable. I've thought about this a few times myself, but I don't want the default to be something obscure (which `P` clearly is).

vspinu:
> Nobody has complained about this and it's configurable.

Sure. Who would open an issue for such a thing? Everyone is annoyed but, given that's configurable, no one takes an effort to "complain" :)

> something obscure (which P clearly is). 

I am not that sure. What other mode uses `xx[project-name]` syntax? It's a very specific and sugestive label as far as I am concerned.

There are some modes out there that you have no idea what that abbreviated lighter means. `P[project-name]` is definitely not like that. And even if it were "obscure", I don't think anyone would mind. Lighter should as short as possible to save space.

bbatsov:
Nobody supported this change, so I'm closing the PR.

PR #739, opened 2015-05-09
Simplify helm-projectile command

tuhdo:
If invoked outside of a project, helm-projectile only shows a list of
projects. Currently, helm-projectile-sources-list includes project
roots, project buffer and project file listings, but empty when used
outside of a project effectively shows only project listing.


bbatsov:
:+1:

PR #744, opened 2015-05-13
[Fix #743] Allow buffer source to be composable

### LINKED ISSUE #743: [bug] helm-projectile outside of project no longer "displays list of projects"
Documented behavior for `helm-projectile` is such that, if called outside of a project, it "displays a list of known projects to jump [to]."

This was the behavior up until about roughly a week ago, and is no longer the case; calling outside of a project now simply prints to minibuffer "You're not in a project"

tuhdo:
Similar to other sources, simply returns nil when not in a project, so
it can be used with other Helm commands. For explicity
helm-projectile-switch-to-buffer, the default error is still signaled.


bbatsov:
:+1:

PR #745, opened 2015-05-17
Align projectile-compile-project even more closely with M-x compile

akopytov:
`projectile-compile-project` now respects the value of
`compilation-ask-about-save` just like `M-x compile` does, with the
exception that only buffers from the current project are considered for
saving.


bbatsov:
:+1: Thanks!

Issue #757, opened 2015-05-31
Add Midje jump-to-test support


arichiardi:
[Midje](https://github.com/marick/Midje) is a better testing library for Clojure and it might be worth having it in the list of projects that can jump from sources to tests.

I have already implemented the (easy) addition, following @bbatsov suggestion of adding ```lein-test``` and ```lein-midje``` project types, but I am stuck at the point where ```projectile-project-type``` tries to find out the right project.

Basically ```midje``` and ```clojure.test``` are both including ```project.clj``` and there is no way to differentiate them based on the file system only (the only difference is the test file name prefix ```t_```.

My question now is:
- Is there another way to differentiate, for example adding some property to ```.projectile``` ?
- What do you suggest @bbatsov (et al.) ? Because ```rails-spec``` and ```rails-test``` have different files in their belly and therefore they easily can be distinguished.

Thanks!

PS.: Additional question, would it make sense to open a PR now?

EDIT: according to [this](https://github.com/marick/Midje/wiki/Configuration-files), ```midje``` has (optional) conf files. The question is still valid...because they could be put in ```$HOME``` only and not in the root of the project.

arichiardi:
Fixed by #758

Issue #642, opened 2015-02-07
Mechanically create test files


expez:
Often there are strong conventions in the community on the name and location of test files.  It would be nice if `projectile-toggle-between-implementation-and-test` created test files in the event they are missing.

I wrote some code for doing this in clojure projects [here](https://github.com/clojure-emacs/clj-refactor.el/issues/113). 

To guard against mavericks, not following the conventional project structure, the functionality should be toggled with `projectile-create-missing-test-files` or some such setting.

arichiardi:
:+1:

arichiardi:
@expez @bbatsov 

I have created the following, works quite well here on 24.3.1 and combined with arichiardi/projectile@fb02c9eb32eae984bf21402a3cd5aa64b482a240 could be a real improvement. For example, I am using ```midje``` and the below code correctly generates ```t_``` tests in my test folder (note that I have not handled other cases then @expez's code in the linked issue above)
I have two problem to handle before a PR as I am not a proper ```elisper```:
- How to set options as the above ```projectile-create-missing-test-files```.
- How to handle the fact that ```cljr--add-ns-if-blank-clj-file``` might be missing.

```co
    (defun projectile--testify-impl-name (impl-file-name)
      (let* ((project-type (projectile-project-type))
             (test-prefix (funcall projectile-test-prefix-function project-type))
             (test-suffix (funcall projectile-test-suffix-function project-type)))
        (cond
         (test-prefix (concat test-prefix impl-file-name))
         (test-suffix (concat impl-file-name test-suffix))
         (t (error "Project type not supported!")))))
    
    (defun projectile-create-test-file-for (impl-file)
      (let* ((test-file (projectile--testify-impl-name (file-name-nondirectory impl-file)))
             (test-dir (replace-regexp-in-string "src/" "test/" (file-name-directory impl-file))))
        (unless (file-exists-p (expand-file-name test-file test-dir))
          (progn (unless (file-exists-p test-dir)
                   (make-directory test-dir :create-parents))
                 (find-file-other-window test-file)
                 ;; (cljr--add-ns-if-blank-clj-file)
                 ;; (save-buffer)
                 ))))
    
    (defadvice projectile-toggle-between-implementation-and-test
      (around expez--create-missing-test-file)
      "Creates a test file if not found in the test/ folder of the project."
      (message "here")
      (let ((file-name (buffer-file-name)))
        (if (projectile-test-file-p file-name)
            ad-do-it
[...]

PR #758, opened 2015-05-31
General test functions improvements + add midje support

arichiardi:
According to issue #757 I have implemented support for [Midje](https://github.com/marick/Midje) here.
I apologize but I am not a proper elisp expert...I am ready for the bashing :smile:

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1496,9 +1496,12 @@ With a prefix ARG invalidates the cache first."
   :group 'projectile
   :type 'function)
 
+(defvar-local projectile-project-type nil "Buffer local var for overriding the project type")
+
 (defun projectile-project-type ()
   "Determine the project's type based on its structure."
   (cond
+   (projectile-project-type projectile-project-type)
    ((projectile-verify-files projectile-rails-rspec) 'rails-rspec)
    ((projectile-verify-files projectile-rails-test) 'rails-test)
    ((projectile-verify-files projectile-ruby-rspec) 'ruby-rspec)
--- a/projectile.el
+++ b/projectile.el
@@ -1476,7 +1476,8 @@ With a prefix ARG invalidates the cache first."
 (defvar projectile-maven '("pom.xml"))
 (defvar projectile-gradle '("build.gradle"))
 (defvar projectile-grails '("application.properties" "grails-app"))
-(defvar projectile-lein '("project.clj"))
+(defvar projectile-lein-test '("project.clj"))
+(defvar projectile-lein-midje '("project.clj" ".midje.clj"))
 (defvar projectile-rebar '("rebar"))
 (defvar projectile-sbt '("build.sbt"))
 (defvar projectile-make '("Makefile"))
@@ -1507,7 +1508,8 @@ With a prefix ARG invalidates the cache first."
    ((projectile-verify-files projectile-python-pip)'python)
    ((projectile-verify-files projectile-python-egg) 'python)
    ((projectile-verify-files projectile-symfony) 'symfony)
-   ((projectile-verify-files projectile-lein) 'lein)
+   ((projectile-verify-files projectile-lein-midje) 'lein-midje)
+   ((projectile-verify-files projectile-lein-test) 'lein-test)
    ((projectile-verify-files projectile-scons) 'scons)
    ((projectile-verify-files projectile-maven) 'maven)
    ((projectile-verify-files projectile-gradle) 'gradle)
@@ -1608,13 +1610,14 @@ PROJECT-ROOT is the targeted directory.  If nil, use
 (defun projectile-test-prefix (project-type)
   "Find default test files prefix based on PROJECT-TYPE."
   (cond
-   ((member project-type '(django python)) "test_")))
+   ((member project-type '(django python)) "test_")
+   ((member project-type '(lein-midje)) "t_")))
 
 (defun projectile-test-suffix (project-type)
   "Find default test files suffix based on PROJECT-TYPE."
   (cond
    ((member project-type '(rails-rspec ruby-rspec)) "_spec")
-   ((member project-type '(rails-test ruby-test lein go)) "_test")
+   ((member project-type '(rails-test ruby-test lein-test go)) "_test")
    ((member project-type '(scons)) "test")
    ((member project-type '(maven symfony)) "Test")
    ((member project-type '(gradle grails)) "Spec")))
@@ -2000,6 +2003,7 @@ For git projects `magit-status-internal' is used if available."
 (defvar projectile-grails-test-cmd "grails test-app")
 (defvar projectile-lein-compile-cmd "lein compile")
 (defvar projectile-lein-test-cmd "lein test")
+(defvar projectile-lein-midje-test-cmd "lein midje")
 (defvar projectile-rebar-compile-cmd "rebar")
 (defvar projectile-rebar-test-cmd "rebar eunit")
 (defvar projectile-sbt-compile-cmd "sbt compile")
@@ -2037,6 +2041,7 @@ For git projects `magit-status-internal' is used if available."
           projectile-maven-test-cmd
           projectile-lein-compile-cmd
           projectile-lein-test-cmd
+          projectile-lein-midje-test-cmd
           projectile-rebar-compile-cmd
           projectile-rebar-test-cmd
           projectile-sbt-compile-cmd
@@ -2069,7 +2074,7 @@ For git projects `magit-status-internal' is used if available."
    ((eq project-type 'django) projectile-django-compile-cmd)
    ((eq project-type 'python) projectile-python-compile-cmd)
    ((eq project-type 'symfony) projectile-symfony-compile-cmd)
-   ((eq project-type 'lein) projectile-lein-compile-cmd)
+   ((member project-type '(lein-test lein-midje)) projectile-lein-compile-cmd)
    ((eq project-type 'make) projectile-make-compile-cmd)
    ((eq project-type 'rebar) projectile-rebar-compile-cmd)
    ((eq project-type 'scons) projectile-scons-compile-cmd)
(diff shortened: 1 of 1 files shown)


expez:
@bbatsov is going to ask you to reference the issue this is solving, squash the commits into one and likely limit the length of your commit messages to a 50 char summary.

Otherwise this looks good to me!

arichiardi:
Oh ok, I kept it separated because they are two different things: one adds the ```midje``` support, the other changes some ```projectile``` behaviour...aka I sneaked in something that I thought it was an improvement...I was even thinking of appending here yesterday's work on test generation, but maybe is too much :smile:

bbatsov:
What's this `.midje.clj`? I've never head of it. Is it some marker you made up yourself?

arichiardi:
Ahaha no :smile:, not myself, it is ```midje```'s actual configuration file, which unfortunately has to be used as marker as I had no other way to distinguish between normal and ```midje``` tests...of course if there is a better way I am all ears...
It is optional and might go in ```$HOME``` as well as in project's root.

bbatsov:
Why not simply set the project type in `.dir-locals`. This might be useful for all project types that can't be easily auto-detected.

arichiardi:
Yep is a good idea, I just did not have enough knowledge about ```projectile``` for implementing this :smile: It could either be identified ```.midje.clj``` or ```.dir-locals```... can you please point me to where this is done for other ```project-type```s as well? I will add it.

bbatsov:
We should have a `projectile-project-type` buffer local variable, which is checked in the `projectile-project-type` function. If it's set we use it to determine the type of the project, otherwise we try to autodetect it. This variable should be set `.dir-locals.el` in the root of the respective project.

arichiardi:
Oh wow great, in that case I don't think I need to add anything else. I will test it though to see if it's working fine here.

bbatsov:
> Oh wow great, in that case I don't think I need to add anything else. I will test it though to see if it's working fine here.

We don't really have this behaviour. I suggested it might be a good idea to have it, because of all project types we can't auto-detect. Adding this would be trivial.

arichiardi:
Ok agree, will work on that. I'd better change the PR title and I wanted also to ask you if it would be ok to add #642 as well here or have a separate PR for that.

arichiardi:
So, would ```projectile-project-type``` be a list? At the moment the test prefix/suffix is a unique string and I guess that if the project can be of multiple types, also the test generating functions will need to change accordingly..

marcinant:
@arichiardi  It's just my suggestion. Eclipse supports project `natures` and this is pretty good idea.
A lot of projects can be detected as multiple types for example: wordpress-theme with bower and composer support.

Projectile doesn't provide specialized tools for these projects however this variable could be used by some external tools to run commands on detected project types.

arichiardi:
Yes, I am familiar with Eclipse's natures, and we could take some ideas from it. There are buildspec too inside a ```.project``` file. We should put all this inside our ```.dir-locals.el``` I guess.

See this excerpt:
```xml
		<buildCommand>
			<name>org.eclipse.pde.SchemaBuilder</name>
			<arguments>
			</arguments>
		</buildCommand>
	</buildSpec>
	<natures>
		<nature>org.eclipse.pde.PluginNature</nature>
		<nature>org.eclipse.jdt.core.javanature</nature>
		<nature>ccw.nature</nature>
	</natures>
</projectDescription>
```

bbatsov:
Let's make changes one step at a time - for now having the ability to set the project type manually would be more than enough.

marcinant:
Personally I gave up on projectile again ;) 101 issues unresolved, 21 unmerged PR.
I don't have time to wait for changes in `projectile` core and would like to GTD and move forward.

So, what I do with project type detection. For example something like this:
This function detects composer project type and is a part of PR which is still awaiting.
```
 (defun composer-project-type ()
   "Function to determine project type from composer.json."
    (when (projectile-verify-file "composer.json")
      (let ((composer-json (json-read-file (projectile-expand-root "composer.json"))))
         (cdr (assoc 'type composer-json)))))
```
Then with `clock-in-prepare-hook` I would like to detect project type and do something when I start to work on TODO (for example: I can start/stop apache server on clock-in/clock-out)
```
 (defun idemacs/clock-in-prepare-hook ()
      (let* ((project-type (if (composer-project-type)
                               (composer-project-type)
                             (projectile-project-type)))
```
So, I got additional variable `project-type` which is a kind of wrapper for `projectile-project-type`.
I don't need to use `.dir-locals.el`. Project type is defined in `composer.json` and is properly detected. If not, then fallback to projectile.

This sucks, this is ugly workaround, but as long as `projectile` development is so slow then I see no better solution. `.dir-locals.el` are ok. However I don't like to maintain any additional config file unless I have to.

arichiardi:
It is a nice way to get the project-type...using a customizable lamdba instead of a just var. This will require executing code from .dir-locals.el ... But hey, if you control the project why not?

bbatsov:
> Personally I gave up on projectile again ;) 101 issues unresolved, 21 unmerged PR.
> I don't have time to wait for changes in projectile core and would like to GTD and move forward.

Most of the open issues are features suggestions and small bugs. I'm an extremely busy person and this is an open-source project. If you don't like something you can always fix it.

> This sucks, this is ugly workaround, but as long as projectile development is so slow then I see no better solution. .dir-locals.el are ok. However I don't like to maintain any additional config file unless I have to.

I wish I had more time to appease everyone, but I don't. People can either complain or contribute. I'm fine  either way.

marcinant:
@bbatsov  Don't take this personal. I just described things how they are.
Issues not resolved, PR not accepted. 

I perfectly understand this is Open Source and non-profit.

So, I had choice: wait... , fork and maintain branch with my patches accepted by myself, keep using `vanilla` branch of `projectile` and use workarounds and extensions.
And this is what I did.

bbatsov:
> PR not accepted.

I don't accept PRs which I feel are incomplete. I should be crazy to start merging everything that was ever proposed. Check the outstanding PRs and you'll see that most of them have at least a few comments from me.

arichiardi:
@bbatsov ```projectile-project-type``` is already the function we use to check the project using its files.

To summarize:
1. A ```.dir-local.el``` var (in the future can be a list of "natures") called ```projectile-custom-project-type``` will be the first way to define a project, it would great to have a list of supported project symbols somewhere for the user.
2. The normal ```projectile-verify-files``` is the second way that at the moment checks if all files in the specified list are matching.

Tell me please if everything sounds right :smile: .

bbatsov:
In elisp you can have a variable and a function with the same name. They
live in separate namespaces.

On Wednesday, June 3, 2015, Andrea Richiardi <notifications@github.com>
wrote:

> @bbatsov <https://github.com/bbatsov> projectile-project-type is already
> the function we use to check the project using its files.
>
> To summarize:
> 1. A .dir-local.el var (in the future can be a list of "natures") called
> projectile-custom-project-type will be the first way to define a project,
> it would great to have a list of supported project symbols somewhere for
> the user.
> 2. The normal projectile-verify-files is the second way that at the
> moment checks if all files in the specified list are matching.
>
> Tell me please if everything sounds right [image: :smile:] .
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bbatsov/projectile/pull/758#issuecomment-108451139>.
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

arichiardi:
Oh that story about lisp-1 vs lisp-2 right :+1:

arichiardi:
Of course always ready to change in case of elisp mistakes or bugs :+1:

marcinant:
> I don't accept PRs which I feel are incomplete. I should be crazy to start merging everything that was ever proposed. Check the outstanding PRs and you'll see that most of them have at least a few comments from me.

It's your opinion. My is different. My PR contain code to detect some project types and it does it's job.
I need it so I had options as I described above.

arichiardi:
Mmm I might have made a mistake, why is ```symbol is void: projectile-test-prefix-function```  ? Should it be define to ```projectile-test-prefix``` by default?

arichiardi:
Ok found, will squash when is fine :wink:

arichiardi:
This last commit adds [#642](https://github.com/bbatsov/projectile/issues/642).

It does not save the buffer because from what I have seen it looks like it is not the way ```projectile-find-implementation-or-test-other-window``` and ```projectile-toggle-between-implementation-and-test``` work but it creates the necessary folders if it cannot find them so that the user can just save when needed.

Of course it would be great to save automagically but I didn't do it as maybe this is something we need to discuss (and I can leave with it). An option could be to set the buffer modified.

@expez @bbatsov

bbatsov:
This PR should have been 3 PRs - one for midje, one for the project-type and one for the test creation.

arichiardi:
Yes I agree, do you want me to split? Or we can just keep three separate commits...

bbatsov:
Both options are fine. You'll also have to rebase your branch on top of the current `master`.

arichiardi:
What about my question above, would it be ok to add a set buffer modified when we create a test buffer?

bbatsov:
> Of course it would be great to save automagically but I didn't do it as maybe this is something we need to discuss (and I can leave with it). An option could be to set the buffer modified.

It's OK not to save it.

arichiardi:
and will it read it from ```.dir-local.el``` ? Clever! :smile: Will change

arichiardi:
The advantage of my previous approach is that if you change ```.dir-local.el```, that ```alist``` gets loaded again.. allowing runtime changes..```defvar-local``` does not seems to behave like that here on my machine...but I guess it is not a big issue...

arichiardi:
I have also fixed a bug that when the project was ```lein-test``` instead of ```lein-midje``` (suffix instead of prefix)

arichiardi:
Gerat! I forgot to modify the README and Changelog...

bbatsov:
I saw. I'll update them myself.

arichiardi:
Ok thank you very much b :smile:

expez:
Great work @arichiardi!

arichiardi:
:balloon: :beers: :tada: :balloon: :tada: :beers: :cake:

:sunglasses: 

:bow: 

Ok, enough with my emotions :smile:
Thanks for your (and Bozhidar) hints and corrections...

caioaao:
From my understanding, midje doesn't enforce a naming for test files. All it does is loading tests from the project's `:test-paths` (see [this wiki page](https://github.com/marick/Midje/wiki/Lein-midje-basics)). Is it possible to add this kind of flexibility to projectile? Is there an API for adding test-finding functions to projectile? That may be a better design rather than hard-coding every test-finding strategy here. With this design, one could write a package that inserts the desired functionality to projectile.

arichiardi:
I think projectile is hard coding just common patterns but still gives the flexibility to add your own stuff. See above for `.dir-locals.el` customizations.., it sounds like a reasonable approach. More sophisticated solutions were taken into consideration (like Eclipse's natures) but probably someone needs to step in and take charge of that.

caioaao:
This is not what I meant. I know there is a way of providing a custom `projectile-project-type`, but I wanted to know if there is a way of providing specific functions to find test files. For instance, providing a function that uses CIDER to get the file namespace and tries to find the correct test file based on that.

PR #760, opened 2015-06-01
add a flag that activates helm fuzzy matching.

### LINKED ISSUE #759: Enable fuzzy matching for projectile-completing-read
Helm just got fuzzy matching recently, and it would be nice if there's a setting that would enable fuzzy matching wherever helm can be used. I think it's as simple as adding the `:fuzzy` keyword parameter here: https://github.com/bbatsov/projectile/blob/master/projectile.el#L1112-L1115

aronasorman:
Fixes #759. 

Specifying a truthy value for `projectile-helm-fuzzy-matching` activates helm's fuzzy matching wherever projectile uses helm.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -199,6 +199,14 @@ and `projectile-buffers-with-file-or-process'."
   :group 'projectile
   :type 'symbol)
 
+(defcustom projectile-helm-fuzzy-matching nil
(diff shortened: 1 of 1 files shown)


aronasorman:
Already implemented, it seems (see #759).

PR #767, opened 2015-06-07
Use project ignored files for helm-projectile-ack

### LINKED ISSUE #766: grep/ack/ag don't ignore folders/files specified in .projectile
folders that I add the the `.projectile` to be ignored are still searched with ack/grep/...

My .projectile for [mixxx](https://github.com/mixxxdj/mixxx) is
```
-/plugins
-/lin64_build
-/vamp-plugins
```

when I now search for eg `#define` it will also show me files from the ignored folders. Only the lin64_build folder is in the list of ignored folders for ack.

My original expectation was that the search functions would also ignore folders/files defined in `.projectile`

This is related to #727

kain88-de:
So far only the globally ignored files/directories have been used
while the grep command uses also the ignored files/directories
defined in `.projectile`

This should fix #766. Seem like I was wrong and only the helm-ack command is affected.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -718,10 +718,10 @@ If it is nil, or ack/ack-grep not found then use default grep command."
                           'identity
                           (-union (-map (lambda (path)
                                           (concat "--ignore-dir=" (file-name-nondirectory (directory-file-name path))))
-                                        projectile-globally-ignored-directories)
+                                        projectile-ignored-directories)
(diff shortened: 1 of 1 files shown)


bbatsov:
Looks good. :+1:

cwhatley:
This change made helm-projectile-ack fail on my setup. Why are those variable references and not function calls?

PR #776, opened 2015-06-13
removed check for executable

mijoharas:
`helm-ag` supports more than just `ag` (as outlined in the docs [here](https://github.com/syohex/emacs-helm-ag#helm-agel-with-the-platinum-searcher-or-ack)). Having this check in breaks code such as:

```
(let ((helm-ag-base-command "pt -e --nocolor --nogroup"))
      (call-interactively 'helm-projectile-ag))
```

unless you happen to have ag present (when it's not needed).


bbatsov:
I'm fine with the change. Please, update the commit message to match our standard format.

mijoharas:
Ahh, sorry, I'd forgotten, will change later.

mijoharas:
Hi, I think that should be right. let me know if it is?

bbatsov:
:+1:

PR #789, opened 2015-07-02
Fix #695: 'If ignored directory doesn't have proper permissions, projectile-find-file displays "find: 'foldername': permission denied"'

### LINKED ISSUE #695: If ignored directory doesn't have proper permissions, projectile-find-file displays "find: 'foldername': permission denied"
Project structure : 
-----------------------
.projectile
-- folder_2
    --- test1.c
    --- test2.c
    --- test3.c

.projectile:
-------------
-/folder_1

mope:
projectile-remove-ignored now checks to see if file names begin with "find: `" and contain "': Permission denied" and removes them if they do.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -884,7 +884,9 @@ Operates on filenames relative to the project root."
                          (projectile-ignored-directories-rel))))
     (-remove (lambda (file)
                (or (--any-p (string-prefix-p it file) ignored)
-                   (--any-p (string-suffix-p it file) projectile-globally-ignored-file-suffixes)))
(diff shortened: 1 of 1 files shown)


bbatsov:
Not sure if there's any point doing this. After all it's unlikely you'll have permission issues within a project. Every check on all the project files is pretty expensive...

bbatsov:
@mope Having thought a bit about this I think the proper solution is to simply adjust the `find` command we invoke. See http://stackoverflow.com/questions/762348/how-can-i-exclude-all-permission-denied-messages-from-find for details.

mope:
Thanks @bbatsov. I'll close this pull request and implement your solution in a different one.

PR #793, opened 2015-07-05
Define helm-projectile-projects-map

whilp:
Previously, this map was defined anonymously inside the Helm source. Defining it separately allows users to further customize the map if desired.


stardiviner:
+1

PR #795, opened 2015-07-08
[Fix #794] Update helm-projectile-file-actions for latest helm version

zachlatta:
- Fixes #794 
- Fixes emacs-helm/helm#1088
- Fixes syl20bnr/spacemacs#2282

Took @thierryvolpiatto's suggested patch from emacs-helm/helm#1088, updated it per @bsuh's suggestion, and then turned it into a PR. All ready for merge :shipit:.

PR #796, opened 2015-07-09
helm-projectile.el: compatible with the latest Helm

YorkZ:
Recent changes in Helm (f71ce5e) broke the `helm-projectile-hack-actions`. This commit addresses the incompatible issue introduced by Helm f71ce5e.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -359,8 +359,7 @@ CANDIDATE is the selected file.  Used when no file is explicitly marked."
 
 (defvar helm-projectile-file-actions
   (helm-projectile-hack-actions
-   (cdr (assq 'action (or helm-source-find-files
-                          (helm-make-source "Find files" 'helm-source-ffiles))))
+   helm-find-files-actions
    ;; Delete these actions
    'helm-ff-browse-project
    'helm-insert-file-name-completion-at-point


csaura:
I think does not solve the helm-projectile-witch-project. I've tried and doesn't work.

tuhdo:
There's a similar PR here: https://github.com/bbatsov/projectile/pull/795

bbatsov:
I haven't used helm lately myself, so someone should just tell me which of the two competing PRs should I merge. :-)

YorkZ:
@Bozhidar: I'm the author of `helm-projectile-hack-actions` and `helm-projectile-file-actions`. Unfortunately, the latest Helm broke it which is something hard to avoidable anyway. However, the changes in Helm is good because it has actually made `helm-projectile-file-actions` simpler, i.e., it can now call `helm-projectile-hack-actions` with `helm-find-files-actions` directly, instead of having to calculate it like what we did before.

I've noticed fix f4dc243 as well, but I'm not sure if that also works because I've never tested that. However, I've test this fix, and I think this fix is more concise; Thierry also recommended using `helm-find-files-actions` directly.

bbatsov:
OK, I'll merge this PR.

YorkZ:
@csaura: I just tested `helm-projectile-switch-project` and it worked for me. However, I think it's less likely that the recent changes in Helm would have affected `helm-projectile-switch-project` as well. It's probably not related to this one after all.

YorkZ:
Thanks Bozhidar.

Issue #797, opened 2015-07-09
Lisp error: (error "Given parent class helm-type-file is not a class")


tomprogrammer:
I updated projectile and helm today from MELPA. On startup emacs reports an error: Given parent class helm-type-file is not a class.

I'm not familiar with elisp so I append the output of the debugger here:

```
Debugger entered--Lisp error: (error "Given parent class helm-type-file is not a class")
  signal(error ("Given parent class helm-type-file is not a class"))
  error("Given parent class %s is not a class" helm-type-file)
  eieio-defclass(helm-locate-source (helm-source-async helm-type-file) ((init :initform (quote helm-locate-set-comman
d)) (candidates-process :initform (quote helm-locate-init)) (requires-pattern :initform 3) (history :initform (quote helm-file-name-history)) (keymap :initform helm-generic-files-map) (help-message :initform helm-generic-file-help-message) (candidate-number-limit :initform 9999) (mode-line :initform helm-generic-file-mode-line-string)) nil)           byte-code("\300\301\302\303\304$\207" [eieio-defclass helm-locate-source (helm-source-async helm-type-file) ((init 
:initform (quote helm-locate-set-command)) (candidates-process :initform (quote helm-locate-init)) (requires-pattern :initform 3) (history :initform (quote helm-file-name-history)) (keymap :initform helm-generic-files-map) (help-message :initform helm-generic-file-help-message) (candidate-number-limit :initform 9999) (mode-line :initform helm-generic-file-mode-line-string)) nil] 5)                                                                                      require(helm-locate)
  eval-buffer(#<buffer  *load*-522533> nil "/home/thomas/.emacs.d/elpa/helm-projectile-20150708.2314/helm-projectile.
el" nil t)  ; Reading at buffer position 1610                                                                          load-with-code-conversion("/home/thomas/.emacs.d/elpa/helm-projectile-20150708.2314/helm-projectile.el" "/home/thom
as/.emacs.d/elpa/helm-projectile-20150708.2314/helm-projectile.el" nil t)                                              require(helm-projectile)
  byte-code("\306\307!\210\310\311!\210\310\312!\210\313\314!\203^T^@\315^P\315\211^Q^R\315\211^S^T\315^U\316\317\320
[...]

PR #798, opened 2015-07-09
helm-projectile.el: make compatible with Helm refactoring

jkl1337:
Helm change 6cd108f moved types to helm-types.el. Require this in
the way expected by helm.

Fixes bbatsov/projectile#797


bbatsov:
:+1: Thanks!

PR #799, opened 2015-07-13
Fix typo that cause emacs fails to start

stshine:
The newest helm-projectile on melpa causes emacs fails to start. Fix it.


bbatsov:
Guess I beat you to this. :-)

stshine:
Yeah a happy lose :-)

PR #802, opened 2015-07-14
Replace helm-ff-mode-line-string

tuhdo:
Since helm-ff-mode-line-string no longer exists and is replaced with
helm-read-file-name-mode-line-string.


bbatsov:
:+1: Thanks!

PR #803, opened 2015-07-14
[Fix #454] Add Nested extensions for other files

### LINKED ISSUE #454: helm-projectile-find-other-file should work with nested extensions
There must be a way to make it work with multiple extensions not just the outer most. Unfortunately my elisp foo is so rudimental that it'd take me ages to figure out the fix. 

;; ;; This currently doesn't work because `projectile-get-other-files'
;; ;; uses (file-name-base current-file) which strips only the outer
;; ;; most extension
;; (add-to-list 'projectile-other-file-alist '("bemhtml" "bh.js"))
;; (add-to-list 'projectile-other-file-alist '("bh.js" "bemhtml"))

tuhdo:


bbatsov:
:+1:

PR #804, opened 2015-07-14
Remove a couple of mode-line references from helm-projectile

mdorman:
It would appear there was a mass-deprecation of helm's mode-line settings in emacs-helm/helm@b95e5b049381b953c2d203eb39486344cc7165cd.  `helm-projectile.el` needs to remove those references or it breaks; I specifically witnessed the issue with the `helm-grep-mode-line-string`, and then did a little inspection to find the reference to `helm-buffer-mode-line-string`, which appears to also be removed in the referenced commit.


bbatsov:
Looks good. Just remove the `.` from the commit message's title.

mdorman:
@bbatsov Done.  Thanks a lot for projectile, BTW.

bbatsov:
You're welcome! :-)

PR #806, opened 2015-07-16
Fix #695 'If ignored directory doesn't have proper permissions, projectile-find-file displays "find: 'foldername': permission denied"' (again)

### LINKED ISSUE #695: If ignored directory doesn't have proper permissions, projectile-find-file displays "find: 'foldername': permission denied"

mope:
Change projectile-generic-command so that it prunes unreadable files and
prevents projectile from seeing them.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -796,7 +796,7 @@ Files are returned as relative paths to the project root."
   :group 'projectile
   :type 'string)
 
-(defcustom projectile-generic-command "find . -type f -print0"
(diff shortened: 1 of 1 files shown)


bbatsov:
Looks good. Just change the commit message title to `[Fix #695] Ignore non-readable directories when using find` (or something like this). This is the general style for the commit titles we apply throughout the project.

mope:
Done.

PR #811, opened 2015-07-22
[Fix #695] Detect find -readable support

### LINKED ISSUE #695: If ignored directory doesn't have proper permissions, projectile-find-file displays "find: 'foldername': permission denied"

dibenede:
How does this look?

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -797,7 +797,12 @@ Files are returned as relative paths to the project root."
   :group 'projectile
   :type 'string)
 
-(defcustom projectile-generic-command "find . \\! -readable -prune -o \\( -type f -print0 \\)"
+(defun projectile-detect-find-command ()
+  (if (zerop (shell-command "find /dev/null -readable"))
+      "find . \\! -readable -prune -o \\( -type f -print0 \\)"
+      "find . \\! -perm +444 -prune -o \\( -type f -print0 \\)"))
+
+(defcustom projectile-generic-command (projectile-detect-find-command)
   "Command used by projectile to get the files in a generic project."
   :group 'projectile
   :type 'string)


bbatsov:
Looks reasonable. I'm guessing you tested it manually, right?

dibenede:
Yes, I tested it on the latest OS X and Ubuntu 14.04 with emacs 24.4. I used the example project from the initial bug report.

dibenede:
Slight correction, 24.5.1 on OS X and 24.4 on Ubuntu. Losing track of versions :)

kaushalmodi:
On Linux (RHEL 5.10), I get: `find: invalid predicate `-readable'`.
Is the `-readable` predicate specific to OS X?

kaushalmodi:
Well, it detects that `-readable` is not available on my system and sets `projectile-generic-command` to `"find . \\! -perm +444 -prune -o \\( -type f -print0 \\)"`. But it spams the \*Messages\* buffer with above error.

kaushalmodi:
How about:

```el
(defun projectile--detect-find-command ()
  "Determine the appropriate default find command based on its capabilities.
Needed because of the differences between GNU find and BSD find."
  (let (message-log-max)
    (if (zerop (shell-command "find /dev/null -readable"))
        "find . \\! -readable -prune -o \\( -type f -print0 \\)"
      "find . \\! -perm +444 -prune -o \\( -type f -print0 \\)")))
```

kaushalmodi:
Argh.. also need to prevent the \*Shell Command Output\* buffer from opening.

kaushalmodi:
Also can you make `projectile-generic-command` a `defvar` instead of a `defcustom`. With the current state of code, I cannot override that var in my setup without editing `projectile.el`. 

If I require projectile first, I will get the above error in \*Messages\* and the Shell buffer. If I set my own value for `projectile-generic-command` and then require projectile, my set value will be overridden as this var is a `defcustom`.

dibenede:
I think only the GNU find has -readable. However, CentosOS 5.10 also seems to be missing it, so I guess it's something that came later.

I don't know much about the emacs side of this, so I'm not sure why it's spamming your message buffer. Could you please tell me what version of emacs you're using? I didn't notice this problem on OS X (which also lacks -readable).

kaushalmodi:
I am on the dev version of emacs from github, built yesterday.

By default `shell-command` will output the result to \*Messages\* unless `message-log-max` is set to `nil`. That will also create separate buffers for the shell output stdout and shell error stderr.

dibenede:
Thanks for the clarification. I think I misinterpreted what you meant by spamming. I see the *Messages* entry and extra buffer you're talking about.

Would you mind posting your solution in the #695 thread? I also neglected to consider Windows and that's posing a problem now, too. It's probably best to keep everything in one place.

xuchunyang:
`shell-command` is slow, on my computer, `projectile-detect-find-command` takes about 0.3 seconds to finish:

```elisp
(measure-time (projectile-detect-find-command))
=> "0.293336"

(defmacro measure-time (&rest body)
  "Measure the time it takes to evaluate BODY."
  `(let ((time (current-time)))
     ,@body
     (message "%.06f" (float-time (time-since time)))))
```
It's bad for projectile user who cares about Emacs startup time.

bixuanzju:
I am using OSX with Emacs 24.5. It spams the *Messages* buffer with `find: -readable: unknown primary or operator` when opening emacs.

PR #815, opened 2015-07-24
Improve projectile--detect-find-command

syohex:
1. Don't invoke shell by calling call-process instead of shell-command
2. Don't output error message when 'find' does not support '-readable' option

`1.` improves performance. https://github.com/bbatsov/projectile/pull/811#issuecomment-124335174 (CC: @xuchunyang). (Your benchmark result on my machine. Original: `0.007559`, This version: `0.002801`)
`2.` resolved https://github.com/bbatsov/projectile/pull/811#issuecomment-124343190(CC: @bixuanzju)

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -800,7 +800,7 @@ Files are returned as relative paths to the project root."
 (defun projectile--detect-find-command ()
   "Determine the appropriate default find command based on its capabilities.
 Needed because of the differences between GNU find and BSD find."
-  (if (zerop (shell-command "find /dev/null -readable"))
+  (if (zerop (call-process "find" nil nil nil "/dev/null" "-readable"))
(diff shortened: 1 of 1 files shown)


xuchunyang:
> 1. improves performance. #811 (comment) (CC: @xuchunyang). (Your benchmark result on my machine. Original: 0.007559, This version: 0.002801)

On my computer, your version takes `0.018443`, this is acceptable. I guess your computer's performance is much better than mine.

Though I don't know if it is possible to just use `-perm +444` or something else which is supported by all common versions of the `find` command, i.e., old/new GNU find and BSD find.

bbatsov:
I'll merge this, as it's an improvement of what we have for sure. There's still the problem with using `/dev/null` on Windows, though. I was thinking of just having a `cond` for the 3 operating system and configurable `find` command for all of them - simple and bulletproof.

PR #816, opened 2015-07-24
[Fix #521] Don't add non-projects to project cache

### LINKED ISSUE #521: Projectile cannot find project root even with .projectile file at root
When I enter a project directory and add a `.projectile` file, Projectile throws the error "You're not in a project" until I restart Emacs to clear `projectile-project-root-cache` and enter the project directory again.

@Silex Could you have a look at this?

tuhdo:
When a user invokes Projectile on a non-project directory, such
directory is cached as 'no-project-root in projectile-project-root-cache,
and the second time when such directory is added with a .projectile file
or anything that Projectile can recognize as a project, it will never
become a project because the cache always suggests Projectile such
directory is not a project. This is a mistake. We should only cache
projects, not non-projects.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -690,11 +690,9 @@ The current directory is assumed to be the project's root otherwise."
     (or (--some (let* ((cache-key (format "%s-%s" it dir))
                        (cache-value (gethash cache-key projectile-project-root-cache)))
                   (if cache-value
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1: Makes sense!

PR #817, opened 2015-07-24
[Fix #695] Improve find detection portability

### LINKED ISSUE #695: If ignored directory doesn't have proper permissions, projectile-find-file displays "find: 'foldername': permission denied"

dibenede:
Distinguish between nul and /dev/null depending on Windows usage.
Avoid generating output buffers and spurious *Messages* entries.
Defer generic command detection until required to avoid startup time hit.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -795,17 +795,44 @@ Files are returned as relative paths to the project root."
   :group 'projectile
   :type 'string)
 
-(defun projectile--detect-find-command ()
-  "Determine the appropriate default find command based on its capabilities.
-Needed because of the differences between GNU find and BSD find."
-  (if (zerop (call-process "find" nil nil nil "/dev/null" "-readable"))
-      "find . \\! -readable -prune -o \\( -type f -print0 \\)"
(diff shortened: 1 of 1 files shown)


dibenede:
The "nul" usage on Windows is untested. Otherwise, seems to work on OS X, CentOS 5.10, and Ubuntu 14.04. Windows was tested via elisp evaluation in *scratch* and yielded the expected find command string with cygwin installed.

bbatsov:
Seems to me this spinning out of control. As someone mentioned already - perhaps `find . \\! -perm +444 -prune -o \\( -type f -print0` will work everywhere? Seems generic enough to me.

dibenede:
I completely agree. The GNU man page suggested there were some differences between -readable and the -perm usage (i.e. readable is a superset). If you're ok with the differences, I'm more than happy to give you the trivial default.

bbatsov:
Simpler would be better. We wasted too much time on this already I think.

PR #820, opened 2015-07-28
Disable current-prefix-arg before calling helm-do-ag

syohex:
helm-do-ag uses current-prefix-arg for specifying searching file extension same
as helm-do-grep. This is different from helm-projectile-ag one that setting ag
command line options. So current-prefix-arg should be disabled before calling
helm-do-ag.


bbatsov:
:+1: Thanks!

PR #823, opened 2015-07-29
Update helm variables

tuhdo:
The two variables are renamed in Helm, so we must change those in Helm
Projectile to make it work properly.


bbatsov:
:+1: Thanks!

PR #834, opened 2015-08-04
Don't require root in helm-projectile-find-file-in-known-projects

### LINKED ISSUE #831: helm-projectile-find-file-in-known-projects requires being in a project
projectile-find-file-in-known-projects does not require being in a project, so I would expect the helm version to not require it either.

xuchunyang:
Fix #831


bbatsov:
Thanks!

PR #838, opened 2015-08-07
Fix #830 (helm-projectile-ag issue)

### LINKED ISSUE #830: Error: helm-projectile-ag (void-variable grep-find-ignored-files)
`Projectile version: 20150728.2141`

Calling `helm-projectile-ag` leads to:
```
Debugger entered--Lisp error: (void-variable grep-find-ignored-files)
  helm-projectile-ag()
  call-interactively(helm-projectile-ag nil nil)
  command-execute(helm-projectile-ag)
```
Seems like `grep-find-ignored-files` is not correctly initialized.

Putting `(setq grep-find-ignored-files nil)` somewhere obviously solves that. 

More knowledgeable people may know what the `somewhere` should be (`defcustom` something like that maybe)

microamp:
This is to import `grep` globally in helm-projectile, otherwise `helm-projectile-ag` would fail due to `grep-find-ignored-files` being void. See #830 for more details. I could've included `(require 'grep)` under `helm-projectile-ag` separately, but I see no harm importing it globally. 

This PR is to address #830.

Thanks!

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -41,6 +41,7 @@
 
 (require 'projectile)
 (require 'cl-lib)
+(require 'grep)
 (require 'helm)
 (require 'helm-types)
 (require 'helm-locate)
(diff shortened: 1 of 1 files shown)


kylev:
Manually applying this change to my local install fixed my problem and the error message related to `helm-grep-ignore-files` when I tried to `C-p s s` .

:beer: for @microamp

microamp:
:beers: 

I could've replaced `grep-find-ignored-files` with `helm-grep-ignored-files` as per recent changes in Helm. However, since we're already using `grep` (for `grep` and `ack`) in Projectile, I thought I might as well stick to that. I guess it's more of a philosophical question to @bbatsov.

bbatsov:
That's all the same to me.

PR #842, opened 2015-08-13
Allow specifying regexp to projectile-grep

birkenfeld:
This makes overriding the "query the user" part easier.

This is pull request #616 rebased onto current master.


bbatsov:
The build is failing.

birkenfeld:
So it is, sorry. Fixed now.

PR #856, opened 2015-09-04
[Fix #855] Ignore files and directories in projectile-ag

### LINKED ISSUE #855: Function projectile-ag doesn't support ignored files
Function `projectile-ag` doesn't use list of ignored files (`helm-projectile-ag` works as expected).

laz2:
And i refactor `helm-projectile-ag` with same ignore list.

Upd: I update pull request. Make more simplest. Just fix `projectile-ag` for ignoring configured paths.
And i add auto-requiring of "ag" package.

### DIFF
--- a/Cask
+++ b/Cask
@@ -5,4 +5,5 @@
 
 (development
  (depends-on "noflet")
- (depends-on "helm"))
+ (depends-on "helm")
+ (depends-on "ag"))
--- a/projectile.el
+++ b/projectile.el
@@ -41,6 +41,7 @@
 (require 'ibuffer)
 (require 'ibuf-ext)
 (require 'compile)
+(require 'grep)
 
 (eval-when-compile
   (defvar ggtags-completion-table)
@@ -1829,12 +1830,16 @@ regular expression."
           (projectile-prepend-project-name (format "Ag %ssearch for: " (if current-prefix-arg "regexp " "")))
           (projectile-symbol-at-point))
          current-prefix-arg))
-  (if (fboundp 'ag-regexp)
+  (if (require 'ag nil 'noerror)
       (let ((ag-command (if arg 'ag-regexp 'ag))
+            (ag-ignore-list (-union ag-ignore-list
+                                    (append
+                                     (projectile-ignored-files-rel) (projectile-ignored-directories-rel)
+                                     grep-find-ignored-files grep-find-ignored-directories)))
             ;; reset the prefix arg, otherwise it will affect the ag-command
             (current-prefix-arg nil))
         (funcall ag-command search-term (projectile-project-root)))
-    (error "Ag is not available")))
+    (error "Package 'ag' is not available")))
 
 (defun projectile-tags-exclude-patterns ()
   "Return a string with exclude patterns for ctags."


Cask
```diff
@@ -5,4 +5,6 @@
 
 (development
  (depends-on "noflet")
- (depends-on "helm"))
+ (depends-on "ag")
+ (depends-on "helm")
+ (depends-on "helm-ag"))
```
bbatsov: This shouldn't be needed.

helm-projectile.el
```diff
@@ -758,11 +758,9 @@ If it is nil, or ack/ack-grep not found then use default grep command."
   (interactive (if current-prefix-arg (list (read-string "option: " "" 'helm-ag-command-history))))
   (if (require 'helm-ag nil  'noerror)
       (if (projectile-project-p)
-          (let* ((grep-find-ignored-files (-union (projectile-ignored-files-rel) grep-find-ignored-files))
-                 (grep-find-ignored-directories (-union (projectile-ignored-directories-rel) grep-find-ignored-directories))
-                 (ignored (mapconcat (lambda (i)
+          (let* ((ignored (mapconcat (lambda (i)
```
bbatsov: These changes definitely don't look correct to me.

projectile.el
```diff
@@ -1779,6 +1780,11 @@ This is a subset of `grep-read-files', where either a matching entry from
                (and ext (concat "*." ext)))))
       (or default-alias default-extension))))
 
+(defun projectile-grep-ignored-patterns ()
```
bbatsov: We already use the term patterns for things that came from `.projectile`, so I'd use some other terminology here.

Cask
laz2: I can delete this lines. But i don't understand how to test without this. For testing `projectile-ag` i need package `ag`. I run `script/emacs.sh` and then in *scratch* buffer execute `(require 'ag)`. Same with `helm-projectile-ag`. How to test without this lines?

projectile.el
laz2: I found `projectile-patterns-to-ignore` which returns list of `.projectile` entries without leading slash. But it doesn't used for grep ignoring.  This is feature or bug?

May be better name would be `projectile-grep-ignored` or `projectile-grep-find-ignored` (as  analog of `projectile-project-ignored` for grep/ag).

Cask
bbatsov: `helm-ag` is a package that doesn't having anything to do with `ag` or `projectile`. This is what I was pointing out.

projectile.el
bbatsov: > I found projectile-patterns-to-ignore which returns list of .projectile entries without leading slash. But it doesn't used for grep ignoring. This is feature or bug?

Pretty sure it's used with `grep` but on a bit more higher level (several sources of ignored files get merged together there).



> May be better name would be projectile-grep-ignored or projectile-grep-find-ignored (as analog of projectile-project-ignored for grep/ag).

I guess you should check the exact code we have in `projectile-grep` right now. I'm fairly certain it respects ignored files.

Cask
laz2: But i changed `helm-projectile-ag` which depends on `helm-ag`.

bbatsov: Hmm. I totally missed this - if this is really the case, then someone made a mistake. I don't use `helm-projectile` at all, so I don't pay much attention the things that are submitted to it.

bbatsov:
:+1: Looks good.

PR #863, opened 2015-09-16
Optionally run monky on Mercurial projects

lbolla:
VC handling of Mercurial repos is quite poor. I much prefer "Monky" https://github.com/ananthakumaran/monky

Arguably, the best solution would be to allow users to customize `projectile-vc`, instead of having special cases for "magit", "monky" or the like.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2040,17 +2040,24 @@ to run the replacement."
 (defun projectile-vc (&optional project-root)
   "Open `vc-dir' at the root of the project.
 
-For git projects `magit-status-internal' is used if available."
+For git projects `magit-status-internal' is used if available.
+For hg projects `monky-status' is used if available."
   (interactive)
   (or project-root (setq project-root (projectile-project-root)))
-  (if (eq (projectile-project-vcs project-root) 'git)
-      (cond ((fboundp 'magit-status-internal)
-             (magit-status-internal project-root))
-            ((fboundp 'magit-status)
-             (with-no-warnings (magit-status project-root)))
-            (t
-             (vc-dir project-root)))
-    (vc-dir project-root)))
+  (let ((vcs (projectile-project-vcs project-root)))
+    (pcase vcs
+      (`git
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -2040,17 +2040,24 @@ to run the replacement."
 (defun projectile-vc (&optional project-root)
   "Open `vc-dir' at the root of the project.
 
-For git projects `magit-status-internal' is used if available."
+   For git projects `magit-status-internal' is used if available.
```
bbatsov: These lines shouldn't be indented.

projectile.el
```diff
@@ -2040,17 +2040,24 @@ to run the replacement."
...
-  (if (eq (projectile-project-vcs project-root) 'git)
-      (cond ((fboundp 'magit-status-internal)
-             (magit-status-internal project-root))
-            ((fboundp 'magit-status)
-             (with-no-warnings (magit-status project-root)))
-            (t
-             (vc-dir project-root)))
-    (vc-dir project-root)))
+  (let ((vcs (projectile-project-vcs project-root)))
+    (cond ((eq vcs 'git)
```
bbatsov: You should probably use pcase here.

projectile.el
```diff
@@ -2040,17 +2040,24 @@ to run the replacement."
...
+           (cond ((fboundp 'magit-status-internal)
+                  (magit-status-internal project-root))
+                 ((fboundp 'magit-status)
+                  (with-no-warnings (magit-status project-root)))
+                 (t
+                  (vc-dir project-root))))
+          ((eq vcs 'hg)
+           (cond ((fboundp 'monky-status)
```
bbatsov: A simple `if` should be sufficient here.

bbatsov:
The commit message's title should not end with `.`.

bbatsov:
Squash the two commits into one and update the changelog.

lbolla:
Squashed and updated changelog

bbatsov:
10x!

PR #864, opened 2015-09-16
Honor projectile-use-git-grep in helm-projectile-grep

### LINKED ISSUE #763: helm-projectile-grep fails to respect projectile-use-git-grep variable
In a git project setting `projectile-use-git-grep` significantly increases the performance of `projectile-grep` but that does not seem to have any effect on `helm-projectile-grep`.

From: syl20bnr/spacemacs#1835

lexa:
If projectile-use-git-grep set to a non-nil then helm-projectile-grep will
use git-grep in git projects, otherwise fall back to regular grep.

This patch fixes #763


bbatsov:
:+1:

PR #868, opened 2015-09-19
Make helm-projectile-ag respects .projectile whitelist

starain:
Make helm-projectile-ag respects .projectile whitelist, so it's usable for large scale projects.

Multi-target supports for helm-do-ag has been added in: https://github.com/syohex/emacs-helm-ag/pull/169

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -769,7 +769,7 @@ If it is nil, or ack/ack-grep not found then use default grep command."
                  (helm-ag-command-option options)
                  (helm-ag-base-command (concat helm-ag-base-command " " ignored))
                  (current-prefix-arg nil))
-            (helm-do-ag (projectile-project-root)))
+            (helm-do-ag (projectile-project-root) (or (car (projectile-parse-dirconfig-file)) nil)))
         (error "You're not in a project"))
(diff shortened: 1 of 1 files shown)


helm-projectile.el
```diff
@@ -769,7 +769,7 @@ If it is nil, or ack/ack-grep not found then use default grep command."
                  (helm-ag-command-option options)
                  (helm-ag-base-command (concat helm-ag-base-command " " ignored))
                  (current-prefix-arg nil))
-            (helm-do-ag (projectile-project-root)))
+            (helm-do-ag (projectile-project-root) (or (car (projectile-parse-dirconfig-file)) nil)))
```
bbatsov: What is this supposed to be doing? The or nil is definitely redundant. Don't you want to pass some list of folder to `helm-do-ag`?

bbatsov: Ping

starain: You are right, or nil only needed in my previous hack. Current helm-ag implementation doesn't need it. Fixed.

bbatsov:
Resubmit this PR to https://github.com/bbatsov/helm-projectile

PR #870, opened 2015-09-25
Use help-projectile-grep instead of helm-find-file when selecting a project

lexa:
This patch allows user to use helm-projectile-grep with all its
features (like ignored files and git-grep) during selecting project. 
The only downside of this patch is that user loses ability to run a
non-recursive grep.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This is done via the variables `projectile-project-compilation-cmd` and `project
 * `projectile-compile-project` now offers appropriate completion
   targets even when called from a subdirectory.
 * Add an argument specifying the regexp to search to `projectile-grep`.
+* Use `help-projectile-grep` instead of `helm-find-file` when selecting a project.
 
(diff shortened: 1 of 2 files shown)


bbatsov:
The changes look good, but they should be accompanied by a changelog entry.

lexa:
Patch updated, changelog entry added.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ This is done via the variables `projectile-project-compilation-cmd` and `project
 * `projectile-compile-project` now offers appropriate completion
   targets even when called from a subdirectory.
 * Add an argument specifying the regexp to search to `projectile-grep`.
+* Use help-projectile-grep instead of helm-find-file when selecting a project
```
bbatsov: Wrap function names in backticks and finish this sentence with a `.`.

lexa: Fixed in next PR.

PR #872, opened 2015-09-28
fix non-ascii char parsing in dirconfig

joonoro:
Fixed parsing non-ascii characters in dirconfig (for example ignored directories in `.projectile`). I noticed that when ignoring directories with names like `harjoitustyö.prv` in `.projectile` they were being mangled by `projectile-parse-dirconfig-file` (read as `harjoitusty\303\266.prv`) and as a result weren't actually ignored. Same happens with other non-ascii characters like é.

Not sure if this is the smartest way to fix it.

Some reference material I went through:
https://lists.gnu.org/archive/html/bug-gnu-emacs/2015-04/msg00331.html
https://lists.gnu.org/archive/html/emacs-diffs/2015-04/msg00308.html


joonoro:
Not sure what's going wrong here to be honest, testing it locally it seems to work fine.

bbatsov:
Squash the two commits together, capitalize the commit message and update the changelog (bugs fixed section).

PR #873, opened 2015-10-03
Omit current buffer from switch-to-buffer choices

dandavison:
I'm new to projectile so I feel like this PR may be naive; please close if so! I couldn't understand why `projectile-switch-to-buffer` was offering the current buffer as a choice. This PR excludes it, thus giving similar behavior to core emacs' `read-buffer-to-switch ` for which the docstring says:

```
...
This function is intended for the `switch-to-buffer' family of
commands since these need to omit the name of the current buffer
from the list of completions and default values."
 ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ This is done via the variables `projectile-project-compilation-cmd` and `project
   targets even when called from a subdirectory.
 * Add an argument specifying the regexp to search to `projectile-grep`.
 * Use `help-projectile-grep` instead of `helm-find-file` when selecting a project.
+* Omit current buffer from `projectile-switch-to-buffer` and `projectile-switch-to-buffer-other-window` choices.
 
 ### Bugs fixed
 
--- a/projectile.el
+++ b/projectile.el
@@ -968,23 +968,29 @@ opened through use of recentf."
   "Prepend the current project's name to STRING."
   (format "[%s] %s" (projectile-project-name) string))
 
+(defun projectile-read-buffer-to-switch (prompt)
+  "Read the name of a buffer to switch to, prompting with PROMPT.
+
+This function excludes the current buffer from the offered
+choices."
+  (projectile-completing-read
+   prompt
+   (-remove-item (buffer-name (current-buffer))
+                 (projectile-project-buffer-names))))
+
 ;;;###autoload
 (defun projectile-switch-to-buffer ()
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by dandavison** [comment]
I haven't thought yet about whether other `projectile` functions should be altered to omit the current buffer.

**Comment by bbatsov** [comment]
Filtering a few open buffers is relatively cheap, but filtering a big project's files would be super slow. But this small change probably makes sense.

**Comment by bbatsov** [comment]
P.S. This has to be mentioned in the changelog.

**Comment by dandavison** [comment]
Thanks for the feedback @bbatsov. I believe that, if we are going to make this change, then we ought to make it for both `projectile-switch-to-buffer` and `projectile-switch-to-buffer-other-window`, since the emacs functions of the same names also have the exclude-current-buffer semantics. So I have updated this branch to apply the change to both projectile functions, and also edited the changelog.

(The `;;;###autoload` pragma is only required for `(interactive)` commands, is that correct?)

**Comment by dandavison** [comment]
> Filtering a few open buffers is relatively cheap, but filtering a big project's files would be super slow

Agreed, I think this should be fine from a performance point of view since it's just doing comparisons of (short) strings without disk IO.

File: CHANGELOG.md [hunk 1]
```diff
@@ -20,6 +20,7 @@ This is done via the variables `projectile-project-compilation-cmd` and `project
   targets even when called from a subdirectory.
 * Add an argument specifying the regexp to search to `projectile-grep`.
 * Use `help-projectile-grep` instead of `helm-find-file` when selecting a project.
+* Omit current buffer from `projectile-switch-to-buffer` and `projectile-switch-to-buffer-other-window` choices
```
bbatsov: This sentence should end with a `.`.

projectile.el
```diff
@@ -968,23 +968,29 @@ opened through use of recentf."
   "Prepend the current project's name to STRING."
   (format "[%s] %s" (projectile-project-name) string))
 
+(defun projectile-read-buffer-to-switch (prompt)
+  "Read the name of a project buffer to switch to.
```
bbatsov: This won't make `checkdoc` happy. The param name has to appear in the docstring.

bbatsov:
Apart from my remark - all commits have to squashed into one.

dandavison:
Made changelog entry and docstring compliant, rebased against master, squashed commits, and force-pushed to this branch.

PR #875, opened 2015-10-03
Allow specification of a compilation dir

### LINKED ISSUE #874: Add option to set a build directory other than the project root
Some build systems (cmake) build projects in directories other than the project root dir. So it would be useful if `projectile-compile-project` could be configured to run in a directory other than the root.

I'll send a pull-request to fix this soon, but I thought it would be useful to submit an issue as well.

davidshepherd7:
Other than the project root dir.

Fixes #874.

I added a test because there was some weird behavior with trailing slashes on the directory variable (fixed with `filename-as-directory`).

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2088,20 +2088,30 @@ For hg projects `monky-status' is used if available."
 (defvar projectile-compilation-cmd-map
   (make-hash-table :test 'equal)
   "A mapping between projects and the last compilation command used on them.")
+
 (defvar projectile-test-cmd-map
   (make-hash-table :test 'equal)
   "A mapping between projects and the last test command used on them.")
+
 (defvar projectile-run-cmd-map
   (make-hash-table :test 'equal)
   "A mapping between projects and the last run command used on them.")
+
 (defvar projectile-project-compilation-cmd nil
   "The command to use with `projectile-compile-project'.
 It takes precedence over the default command for the project type when set.
 Should be set via .dir-locals.el.")
+
+(defvar projectile-project-compilation-dir nil
+  "The directory to use with `projectile-compile-project'.
+The directory path is relative to the project root.
+Should be set via .dir-locals.el.")
+
 (defvar projectile-project-test-cmd nil
   "The command to use with `projectile-test-project'.
 It takes precedence over the default command for the project type when set.
 Should be set via .dir-locals.el.")
+
 (defvar projectile-project-run-cmd nil
   "The command to use with `projectile-run-project'.
 It takes precedence over the default command for the project type when set.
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -2085,6 +2085,10 @@ For git projects `magit-status-internal' is used if available."
   "The command to use with `projectile-compile-project'.
 It takes precedence over the default command for the project type when set.
 Should be set via .dir-locals.el.")
+(defvar projectile-project-compilation-dir nil
```
bbatsov: Surround this with blank lines.

projectile.el
```diff
@@ -2131,6 +2135,15 @@ Should be set via .dir-locals.el.")
                           '(compile-history . 1)
                         'compile-history)))
 
+(defun projectile-compilation-dir ()
+  "Choose the directory to use for project compilation."
+  (cond (projectile-project-compilation-dir
```
bbatsov: This has just two branches, so I'd simply use an `if`.

bbatsov:
I'm not sure your changes will play nice with the compilation command caching functionality, as it's tied to the project root and you can have several `dir-locals` in a single project. Probably this was not your intention, but I'm just mentioning this.

projectile.el
davidshepherd7: The other variables in that section aren't currently separated by blank lines. Should I add blank lines between them all?

bbatsov: Yeah, you can do this. This was likely an oversight.

davidshepherd7:
I've pushed changes for those two comments.

For the command caching: yeah I'm just using a single `.dir-locals` file in the project root. I guess there's nothing we can (reasonably) do to make the caching work with multiple `dir-locals` files setting `projectile-project-compilation-dir`, but if you have any ideas I could give it a try.

:+1:

bbatsov:
Can't we simply use the compilation dir in the cache instead of the root dir? This will for an arbitrary number of compilation folders (although I'm not sure any project would have more than one). Even in the current state of the code it's kind of strange that we're using a different key in the command cache than the target dir itself. Admittedly this is unlikely to be an issue in practice...

davidshepherd7:
Yeah actually, that should work. I'll try it when I get a chance (probably
tomorrow).

I think it could be useful, e.g. with cmake you can have separate 64 and 32
bit build dirs with slightly different compile commands.

bbatsov:
ping :-)

davidshepherd7:
Still planning to do this, should be able to give it a go this weekend.

bbatsov:
OK

davidshepherd7:
That should do it, was much easier than I expected :)

bbatsov:
Just squash all those commits together and we're good to go.

davidshepherd7:
Ok, done :+1:

PR #878, opened 2015-10-08
Fix many warnings and some errors

tarsius:


tarsius:
I just noticed that I have a commit laying around in my local master which fixes a few more warnings. I'll wrap that up and then update this pr.

davidshepherd7:
You probably know this, but for anyone else: this commit fixes the bug where projectile-ag doesn't respect ignored files lists as it should.

bbatsov:
@tarsius when can we expect the updated PR?

tarsius:
I'll do it this week.

tarsius:
Looks like I found some time tonight already.

By the way I think I'll stop providing these fixes and recommend that you instead compile before merging contributions and require that contributors fix their own mistakes before you accept their changes.

Sure I sometimes forget to do that too or forgo doing it because I have a vague feeling that it is "not necessary in this case" (which then turns out to be wrong), but with this particular package it is a pattern that has to be broken. Many contributions are from one-time contributors and this is a popular package, so I do realize that merging pull requests carefully does increase the workload quite a bit. You should do it anyway, or find a deputy who does it for you.

In the long term I intend to provide proper "pull request handling" directly in Magit, but in the short term you might want to give [magit-branch-pull-request](https://github.com/tarsius/magit-rockstar/blob/c3bdcf2528dabad8bd5070e3ac714c7970bc6d29/magit-rockstar.el#L124) from my [magit-rockstar (package description)](https://github.com/tarsius/magit-rockstar/blob/c3bdcf2528dabad8bd5070e3ac714c7970bc6d29/magit-rockstar.el#L25) a try. Still not as efficient as clicking "merge" in Github, but then again that feature is an insufficient simplification anyway.

I'll stop acting as "fixing compiler warnings as a service" for all the packages I have installed, and invest the saved time to improve Magit's pull request handling instead :-)

Silex:
Ha, magit-rockstar is nice, it's exactly what I was thinking of scripting the other day. Currently I used a keyboard macro that did `c -R a C-c C-c r r` after an interactive rebase that edited all commits :)

tarsius:
Just opened https://github.com/milkypostman/melpa/pull/3218.

davidshepherd7:
I think it should be possible to add a new make "warning check" task which compiles with `byte-compile-error-on-warn` true. Then you could run the warning check on travis and contributors would get notified about warnings.

I'll give it a try sometime.

bbatsov:
I have a byte-compiler warning check in Travis for some projects (e.g. CIDER), guess I can add one here as well.

bbatsov:
Thanks, @tarsius!

tarsius:
> compiles with byte-compile-error-on-warn true.

Good idea, need to look into that too.

PR #879, opened 2015-10-10
Add projectile-before-switch-project-hook

john2x:
Add hook right before `switch-project-action` is called.

This is related to the discussion at https://github.com/milkypostman/melpa/pull/3177

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* Add `projectile-before-switch-project-hook`.
 * Add the ability to specify the project type via `.dir-locals.el`.
 * Add support for projects using Midje.
 * Add the ability to create missing tests automatically (controlled via the `projectile-create-missing-test-files` defcustom).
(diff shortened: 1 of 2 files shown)


bbatsov:
Looks good. I think we should probably rename the hook that's executed after the project switch. This also needs a changelog entry.

john2x:
Added an entry to the changelog. Should I open a different PR for renaming `projectile-switch-project-hook`?

bbatsov:
> Added an entry to the changelog.

Squash the two commits together.

> Should I open a different PR for renaming projectile-switch-project-hook?

Yes.

john2x:
For my use case at milkypostman/melpa#3177, I need the the project name to be passed to the "before" hook function. What do you think about using
```
(run-hook-with-args 'projectile-before-switch-project-hook project-to-switch)
```
instead?

bbatsov:
Another alternative would be to let-bind whatever you need around the hook. If we opt for a hook with args it'd be nice if yielded both the current and the target project.

bbatsov:
P.S. Read your question after reading the diff.

bbatsov:
Btw, you should also read here about the conventions for abnormal hooks http://www.gnu.org/software/emacs/manual/html_node/elisp/Hooks.html#Hooks

PR #881, opened 2015-10-13
Fix issue with ack vs. grep in last update

cwhatley:
The latest update broke helm-projectile-ack. This should restore functionality.

I don't see where the dir argument is used in the grep scenario, so not sure that works here.

### DIFF
--- a/helm-projectile.el
+++ b/helm-projectile.el
@@ -735,25 +735,24 @@ DIR is the project root, if not set then current directory is used"
                               #'helm-projectile-grep-or-ack project-root nil)))
 
 ;;;###autoload
-(defun helm-projectile-ack ()
+(defun helm-projectile-ack (&optional dir)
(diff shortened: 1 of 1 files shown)


bbatsov:
Can you be more specific what exactly does this PR fix?

cwhatley:
A project-root/directory argument was added to the functions helm-projectile-grep-or-ack and helm-projectile-grep, but not to helm-projectile-ack and how helm-projectile-ack doesn't work. Without this change, "C-p s a" gives:

`Error running timer 'helm-projectile-grep-or-ack': (wrong-type-argument stringp t)`

bbatsov:
Ah, ok. Just remove the `.` at the end of the of the commit message title and we're good to go.

cwhatley:
OK, I adjusted it.

bbatsov:
Thanks!

PR #885, opened 2015-10-20
support gradle projects with gradlew script; resolves #840

### LINKED ISSUE #840: Support executing gradle wrapper script
While there is support for gradle projects, it could be extended to support the somewhat common case of having a `gradlew` (gradle wrapper script) in the root directory of the project. When detected, `./gradlew build` should be executed, instead of `gradle build`.

See https://docs.gradle.org/current/userguide/gradle_wrapper.html.

lawremi:
This won't work on Windows due to use of `./`. Presumably we could condition on the OS and drop the `./` to run the `.bat` file, but I have no easy way to test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+## 0.13.0 (10/21/2015)
+
 ### New features
 
 * Add `projectile-before-switch-project-hook`.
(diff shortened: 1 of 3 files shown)


bbatsov:
Please, update the commit message to match our commit message style - `[Fix #840] Support...`.
You should also add a changelog entry.

lawremi:
I think I've messed up this PR beyond repair. I've torn my hair out for an hour trying to figure out how to fix the history with an updated changelog. I'm just going to make a completely new PR.

PR #888, opened 2015-10-24
Add build.boot to projectile-project-root-files

pbzdyl:
In my previous pull request I forgot to include build.boot in projectile-project-root files so I am fixing this omission.

Let me know if I should update CHANGELOG with this change.


bbatsov:
:+1:

PR #889, opened 2015-10-25
Add check for byte-compiler warnings in make test

davidshepherd7:
Add a new byte-compile-strict command to Makefile which byte-compiles
`projectile.el` and `helm-projectile.el` with
`byte-compile-error-on-warn` true.

The new command is currently a dependency of `make test`. This should
ensure that contributors notice the warnings before submitting a pull
request. Alternatively we could have it as a separate make command and
run it separately on Travis if that's preferable, but this might result
in most people not seeing the warnings until the Travis build fails.

I didn't add automatic byte compilation of persp-projectile.el because it
requires a package which isn't included.

I had to use `--directory "."` so that emacs finds `projectile.el` while
compiling `helm-projectile.el`. I don't know why it doesn't
automatically have the path set correctly...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,17 @@ elpa:
 .PHONY: build
 build : elpa $(OBJECTS)
 
+.PHONY: byte-compile-strict
+byte-compile-strict : elpa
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1: Thanks!

PR #917, opened 2015-11-28
Announce cache is empty before blocking.

### LINKED ISSUE #899: Projectile shouldn't wait before announcing the cache is empty
When projectile indexes a very large repository, it blocks emacs for several seconds while collecting a list of files in the repository, then the `Empty cache. Projectile is initializing cache...` message flashes for a split second, right before the process completes. It would be more sensible to print this message before finding files.

dpathakj:
It can take a long time before
```elisp
    (-mapcat #'projectile-dir-files
             (projectile-get-project-directories))
```
completes. This makes it more evident what's happening.

Fixes #899

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -563,7 +563,6 @@ to invalidate."
   "Cache PROJECTs FILES.
 The cache is created both in memory and on the hard drive."
   (when projectile-enable-caching
-    (message "Empty cache. Projectile is initializing cache...")
     (puthash project files projectile-projects-cache)
     (projectile-serialize-cache)))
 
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -1183,6 +1182,8 @@ https://github.com/abo-abo/swiper")))
   "Return a list of files for the current project."
   (let ((files (and projectile-enable-caching
                     (gethash (projectile-project-root) projectile-projects-cache))))
+    (when (and projectile-enable-caching (not files))
```
bbatsov: I'd put this in the `unless` a couple of lines down.

bbatsov:
Reword the commit message title to `[Fix #899] ....` and drop the `.` from it.

projectile.el
dpathakj: Yeah, that's obviously better. Don't know why I didn't do that.

bbatsov:
> Reword the commit message title to [Fix #899] .... and drop the . from it.

By this I meant `[Fix #899] Announce cache is empty before blocking`. :-)

dpathakj:
I've made the change. Anything else I should do?

bbatsov:
:+1:

PR #919, opened 2015-12-01
convert defadvices to new style advice

spwhitton:
Convert projectile's use of `defadvice` to `nadvice.el`, that is, `defun` plus `advice-add`.

Thanks for projectile!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Add [boot-clj](https://github.com/boot-clj/boot) project type.
 * Add support for projects using gradlew script.
 * Prefer Haskell stack projects over cabal projects.
+* Convert defadvice to defun + advice-add: new style advice from nadvice.el
 
 ## 0.13.0 (10/21/2015)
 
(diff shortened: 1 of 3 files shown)


syohex:
nadvice was introduced at Emacs 24.4. Does projectile supports Emacs 24.4 or higher version?

spwhitton:
It looks like the Travis build failed because Travis is using a version of Emacs older than 24.4.

Emacs 24.4 is in Debian Stable.  That means it's not exactly bleeding-edge.  So perhaps we could expect anyone using an older version who wanted to use projectile to download an older tag.  But that's up to the project maintainer.

bbatsov:
Frankly, I never thought about the minimum supported version. In my mind it was 24.1 (although it's probably not). If Debian stable really ships 24.4, then I guess setting it as the minimum supported version is OK. What's the Emacs version in RHEL 7?

spwhitton:
[This page](https://git.centos.org/summary/?r=rpms/emacs) suggests 24.3.

bbatsov:
Guess we can't convert the advices just yet then.

Issue #848, opened 2015-08-25
Ability to add a custom Project name? 


ReneFroger:
I use Yasnippet in pair with Projectile. Is there any way to prefer custom project names above the current project names? 

For example in this Yasnippet: 

    #include <iostream>

    int main () {
    try  {
    }
    catch (int e)  {
            cout << "The `(buffer-name)` from `(project-name)`) encountered an error. Exception Nr. " << e << '\n';
        }
    return 0;
    }
   
Sometimes a project does have a unclear name, e.g. ws_project_fooBAR12. So I get unclear snippets. Is it possible to have any ability in Projectile to define custom project names?

glucas:
+1, I have some projects for which I'd like to have a more meaningful name without having to change my directory structure.

EHartC:
You could easily create your own function for that, e.g:

    (defun my-own-project-name ()
    ;; Checks if Projectile is currently active, or this will once run into undefinable errors.
    (when (featurep 'projectile)
        ;; Checks if the file ".custom-project-name" is currently located in the project root. 
        (if (file-exists-p (concat (expand-file-name (projectile-project-root)) ".custom-project-name"))
            ;; If that's the case, write their contents into the buffer. 
            (with-temp-buffer
            (insert-file-contents (concat (expand-file-name (projectile-project-root)) ".custom-project-name"))
            (buffer-string))
        ;; If not, then extract the projectname.
        (replace-regexp-in-string "/proj/\\(.*?\\)/.*"
                                    "\\1"
                                    (projectile-project-name)))))

After evaluating the function above, create the file `.custom-name`, which contains your own projectname and located in the current project root. If that file doesn't exist (yet), it will use the current projectname instead.

glucas:
Thanks for the suggestion, but it looks at least a bit more involved than that.

In addition to replacing (with advice, I guess) the default implementation of `projectile-project-name` with a function like what the one suggested, it looks like there are a number of places where projectile assumes the project name is actually a directory path and uses it that way -- lots of functions that let-bind `default-directory` to the project name.

bbatsov:
The best way to implement this functionality would be simply to augment the existing project tracking. Basically there should be some function which allows you to name a project and it will be saved in the list of known projects as name + folder. All the code dealing with known projects should be updated to reflect this as well. Not a lot of work, but as usual - I'm pretty low on free time.

dustinpaluch:
[...]

Issue #329, opened 2014-04-29
Reorder projectile-project-root-files-functions to detect projects files before vcs.


proofit404:
Sometimes project root doesn't match vcs root directory. For example when it stored in subdirectory like src or something. Maybe it's a good idea detect project file markers first, top-down-recurring markers second and bottom-up files last.

pronobis:
The interesting thing is that this is exactly what happens to me without any changes. If i have `.projectile` file in a folder which has multiple repositories in it resulting in sth like: `src/<repo>/.git`, then the main folder has priority and projects do not appear at all for each of the repos.

catesandrew:
This is how a lot of node projects are arranged too but under the `node_modules` directory. I typically add other root markers like `project.json` or `bower.json` to help establish the child directories under `node_modules` as another project. In vi I would open those in another tab, in emacs I could use other tabs but prefer to either switch projects or a new frame since its easier to switch between frames in emacs while its not in vi.

proofit404:
@pronobis And what for 
```
.git/
src/.projectile
```
?
In my opinion `.projectile` file must have greater priority than `.git` directory.

billop:
Are there any workarounds for making "file markers" have precedence over vcs?

PR #928, opened 2015-12-11
Enable custom project names

fvaresi:
This enables the user to customize the project name based on project-root. This is helpful in situations where the project base dir is not descriptive enough.

As an example, I use it like this:

```elisp
     (defun fvaresi/projectile-custom-project-name (project-root)
        (cond
         ((string-match "/Plugins/\\([^/]+\\)/\\(?:branches\\|tags\\)\\(?:.*\\)/\\([^/]+\\)" project-root)
          (let* ((product-name (match-string 1 project-root))
                 (branch-name (match-string 2 project-root)))
            (concat product-name "/" branch-name))) ;

         ((string-match "/Plugins/\\([^/]+\\)/trunk" project-root)
                (let* ((product-name (match-string 1 project-root)))
                  (concat product-name "/trunk")))
   
         ;; do the default
         (t
          (projectile-default-project-name project-root))))

      (setq projectile-project-name-function 'fvaresi/projectile-custom-project-name)
```

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -207,6 +207,22 @@ Two example filter functions are shipped by default -
   :group 'projectile
   :type 'symbol)
 
+(defcustom projectile-project-name nil
+  "If this value is non-nil, it will be used as project name.
+
+It has precedence over function `projectile-project-name-function'."
+  :group 'projectile
+  :type 'string
+  :package-version '(projectile . "0.14.0"))
+
+(defcustom projectile-project-name-function 'projectile-default-project-name
+  "A function that receives the project-root and returns the project name.
+
+If variable `projectile-project-name' is non-nil, this function will not be used."
+  :group 'projectile
+  :type 'symbol
+  :package-version '(projectile . "0.14.0"))
+
 (defcustom projectile-project-root-files
   '("rebar.config"       ; Rebar project file
     "project.clj"        ; Leiningen project file
@@ -738,15 +754,21 @@ A thin wrapper around `file-truename' that handles nil."
       (projectile-project-root)
     (error nil)))
 
+(defun projectile-default-project-name (project-root)
+  "Default function used create project name to be displayed based on the value of PROJECT-ROOT."
+  (file-name-nondirectory (directory-file-name project-root)))
+
 (defun projectile-project-name ()
   "Return project name."
-  (let ((project-root
-         (condition-case nil
-             (projectile-project-root)
-           (error nil))))
-    (if project-root
-        (file-name-nondirectory (directory-file-name project-root))
-      "-")))
+  (if projectile-project-name
+      projectile-project-name
+      (let ((project-root
+             (condition-case nil
+                 (projectile-project-root)
+               (error nil))))
+        (if project-root
+            (funcall projectile-project-name-function project-root)
+          "-"))))
 
 
 ;;; Project indexing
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -207,6 +207,11 @@ Two example filter functions are shipped by default -
   :group 'projectile
   :type 'symbol)
 
+(defcustom projectile-project-name-function 'projectile-default-project-name
+  "A function that receives the project-root and returns the project name to be displayed."
+  :group 'projectile
+  :type 'symbol)
```
bbatsov: This should have a package-version properly.

projectile.el
```diff
@@ -736,6 +741,10 @@ A thin wrapper around `file-truename' that handles nil."
       (projectile-project-root)
     (error nil)))
 
+(defun projectile-default-project-name (project-root)
+  "Default function used create project name to be displayed"
```
bbatsov: `M-x checkdoc` won't like this. `project-root` must appear in all-uppercase in the docstring and it should end with a `.`.

bbatsov:
This should be extended a bit further. If there's a buffer-local variable - `projectile-project-name` it should take precedence over calling this function IMO. The ultimate flexibility.

fvaresi:
Thanks for the feedback. I'll implement these changes as soon as I get a moment.

bbatsov:
You'll also have to rebase on top of the current `master`.

fvaresi:
I started implementing the precedence of a buffer-local-variable but saw some issues while testing this. I'm guessing this local variable would be for referencing projects in an external directory. Is this the use case you were thinking about?

I think using this would also require to use the variable proposed for customizing project-root (#329), otherwise it won't be able to find the project root.

Thanks again for your efforts and sorry about being so slow (these are my first baby steps in contributing to OS).

bbatsov:
> I think using this would also require to use the variable proposed for customizing project-root (#929), otherwise it won't be able to find the project root.

It's totally unrelated. The idea is to be able to set the project name via `.dir-locals.el`, without having to deal with custom name functions. Extremely useful if you don't really want to derive the project name from some directory, but would rather set it explicitly.

fvaresi:
I now see what you meant and implemented the suggested changes. I think tests are currently giving an error cause helm-projectile was moved to its own repo.

Also, since persp-projectile was also moved I'll send a new pull request for changes related to this one, once this pull request is approved.

bbatsov:
The changes look good. Just squash the 3 commits together and we're good to go.

fvaresi:
It's done, thanks.

bbatsov:
And drop that `.` from the commit message - titles shouldn't end with `.`.

fvaresi:
Ok, modified commit message.

bbatsov:
:+1: Thanks!

ReneFroger:
@fvaresi could you tell more how you could specifice a custom project name inside a project? I couldn't figure it out from your code and the changes.

fvaresi:
@ReneFroger if the custom project name is a function depending on the project root, you can set `projectile-project-name-function` as shown in the first comment of this pull request.

If you want to customize the project name just for one project, it's easier to set the local variable `projectile-project-name` using [directory variables](https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html). This is probably the simplest way, since you just need to add a `.dir-locals.el` file in the project root.

ReneFroger:
Thanks for your answer, but I got the asking message everytime:

    "The local variables in [project folder] contains values that may not be safe (*)

    Do you want to apply it? 

    y
    n
    !

    * projectile-project-name . "my-super-project"

Is there any way to turn that confirmation message off? Typing ! didn't help
after I visit the project again.

fvaresi:
In my setup using ! will just add one safe value. If you want to say that all strings are safe values to be used you can add this to your emacs config:

```emacs-lisp
(put 'projectile-project-name 'safe-local-variable #'stringp)
```

ReneFroger:
@fvaresi thanks for your helpful response. It works like a charm, and integrates well with my yasnippets (https://github.com/bbatsov/projectile/issues/848), thanks for your contribution, it's appreciated!

PR #398, opened 2014-08-03
Allow setting projectile root as a file local variable.

ibrahima:
This is useful, for example, if you keep your OrgMode files in one git repo but associate them with other projects in other directories. It just saves you having to switch to a different project all the time.

This is my first time contributing to a big elisp project so I have no idea if this is following the proper conventions. Any feedback is appreciated! Also I have no idea if anyone else would find this functionality useful, but I just implemented it for myself in my .emacs.d and decided to push it up here too in case it was useful for others.

PR #929, opened 2015-12-15
Allow setting projectile root as a file local variable.

ibrahima:
This is useful, for example, if you keep your OrgMode files in one git
repo but associate them with other projects in other directories.

This is a resubmission of #398 addressing Bozhidar's comments. Any feedback is appreciated, again!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -260,7 +260,8 @@ containing a root file."
   :type '(repeat string))
 
 (defcustom projectile-project-root-files-functions
-  '(projectile-root-bottom-up
+  '(projectile-root-file-local
+    projectile-root-bottom-up
     projectile-root-top-down
     projectile-root-top-down-recurring)
   "A list of functions for finding project roots."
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -678,6 +679,13 @@ which we're looking."
              (setq file nil))))
     (and root (expand-file-name (file-name-as-directory root)))))
 
+(defvar-local projectile-custom-root nil
```
bbatsov: Why not name it `projectile-project-root`?

projectile.el
```diff
@@ -678,6 +679,13 @@ which we're looking."
...
+  "Defines a custom Projectile project root.
+   This is intended to be used as a file local variable.")
+
+(defun projectile-root-file-local (dir)
```
bbatsov: I wouldn't wrap this in a function at all - it should simply be the first check in the `projectile-project-root` function.

bbatsov:
This should be mentioned somewhere in the readme. The commit message should not end with a `.`.

bbatsov:
ping

projectile.el
ibrahima: I wasn't sure if that would be bad practice, to have a variable and a function with the same name.

projectile.el
ibrahima: It seemed like it was easiest/least intrusive to add the function to projectile-project-root-files-functions rather than add a check to projectile-project-root. As I wrote it initially, it was possible for me to contain my changes within my init.el rather than actually requiring modifications to projectile.

ibrahima:
Thanks for merging!

projectile.el
aaronjensen: This ended up pretty severely regressing performance if `projectile-project-root` is nil because it does a `file-truename` on `dir` every time.

PR #935, opened 2015-12-24
ag already supports git ignore files, so no need to recalculate it.

fommil:


fommil:
Updated

bbatsov:
One more nitpick - commit message titles should not end with a `.`.

fommil:
updated

Issue #699, opened 2015-04-01
Error in find tag in project (projectile 20150329.7)


krzysztof-magosa:
Hi,
I've pressed C-c p j and below error appeared:

Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  expand-file-name(nil)
  tags-expand-table-name(nil)
  visit-tags-table-buffer(t)
  #[0 "\303\211C\304\30\"\210\212\306\242\205\307!\2038  \262\211\242\2031\310\311\312\313\314\315!\316\"\317\320%\"\210\202    \211\240\210\202    )\304\32\"\210\211\242\211\207" [buffer-file-name tags-completion-table-function tags-completion-table nil message "Making tags completion table for %s..." visit-tags-table-buffer t mapatoms make-byte-code 257 "\301\302!\300\242\"\207" vconcat vector [intern symbol-name] 4 "\n\n(fn SYM)" "Making tags completion table for %s...done"] 9 "\n\n(fn)"]()
  funcall(#[0 "\303\211C\304\30\"\210\212\306\242\205\307!\2038  \262\211\242\2031\310\311\312\313\314\315!\316\"\317\320%\"\210\202    \211\240\210\202    )\304\32\"\210\211\242\211\207" [buffer-file-name tags-completion-table-function tags-completion-table nil message "Making tags completion table for %s..." visit-tags-table-buffer t mapatoms make-byte-code 257 "\301\302!\300\242\"\207" vconcat vector [intern symbol-name] 4 "\n\n(fn SYM)" "Making tags completion table for %s...done"] 9 "\n\n(fn)"])
  tags-completion-table()
  projectile-find-tag()
  call-interactively(projectile-find-tag nil nil)
  command-execute(projectile-find-tag)

krzysztof-magosa:
One thing, it works directly after projectile-regenerate-tags and stops after emacs is restarted.

krzysztof-magosa:
By the way, I've found out that it works okay with 'ggtags' as a backend, but not with ctags.

bbatsov:
I'm closing this ticket due to no recent activity. Feel free to re-open it if you ever come back to it.

Issue #683, opened 2015-03-10
projectile-find-tag missing tags


fommil:
Comparing against http://emacswiki.org/emacs/InteractivelyDoThings#toc11

```elisp
  (defun my-ido-find-tag ()
    "Find a tag using ido"
    (interactive)
    (tags-completion-table)
    (let (tag-names)
      (mapatoms (lambda (x)
                  (push (prin1-to-string x t) tag-names))
                tags-completion-table)
      (find-tag (ido-completing-read "Tag: " tag-names))))
```

the `projectile-find-tag` is missing a lot of entries on a Java/Scala project. e.g. classnames don't appear to be searched with projectile.

rootzlevel:
I can confirm this issue.

To reproduce, try to jump to the tag `Dispatcher` with projectile and [this TAGS file](http://wwwcip.cs.fau.de/~lu03pevi/TAGS). It is not offered as one of the alternatives.

`M-x find-tag` on the same tag works just fine.

My emacs version is 24.5.1 and projectile is of version 20150428.158

fommil:
I was looking into this again and found out that the `ido` completions do not find the tag, but if I type out what I'm searching for in full, then it does seem to be able to jump to the definition. Perhaps `ido` is just aggressively pruning and dropping valid completions?

rootzlevel:
It's not `ido`, setting `projectile-completion-system` to `'default` has the same problem.

fommil:
ok, well I'm just using `find-tag` as the binding until this is fixed so that's a reasonable workaround.

base698:
I'm having this problem too with a javascript project.

jaycotton:
I'm having the same problem in both Perl and Python projects. `projectile-find-tag` does not include many function names in the completion list. Looking into it a bit more, I see that the completion list is being generated by the function etags-tags-completion-table (on my setup at least), from the etags emacs package.

I can manually complete the function I'm looking for, but if there are multiple instances of that function definition in the project, it just jumps to the first one rather than requesting disambiguation.

jaycotton:
The problem is that the completion table returned by `etags-tags-completion-table` has a hard-coded length of 511. Only the last 511 tags in the TAGS file will be returned. My file has around 11,000 tags.

It is surprising to me that `find-tag`, which is in the same package as `etags-tags-completion-table`, does not share this limitation. It happily provides me with 11,000 items in the completion buffer.

base698:
[...]

Issue #668, opened 2015-03-01
projectile-find-tag very slow


fommil:
Even just hacking in my `.emacs.d` the `TAGS` file is about 8MB, which is fine, and `find-tags` is pretty instant.

However, when using `projectile-find-tags` there is a couple second delay every time. It looks like it's rebuilding the table every time, is that correct? Is there no better way to do this?

fommil:
btw, in elisp code it would also be amazing to drop down to find-function-at-point if the TAGS table returns negative.

fommil:
```elisp
(defun elisp-find-tag-or-find-func ()
  ;;; Could be a lot smarter for non-function symbols (variables, faces, packages, etc)
  "Use `find-tag' to find symbol at point, falling back to `find-func' features."
  (interactive)
  (let* ((fun (function-called-at-point)))
    (if (find-function-noselect fun 'no-c-sources)
        (find-function-do-it fun nil 'switch-to-buffer)
      (find-tag (symbol-name (symbol-at-point))))))
```

fommil:
In response to my own comment, I cooked this up for elisp

```elisp
(defun elisp-find-tag-or-find-func ()
  ;;; Could be a lot smarter for non-function symbols (variables, faces, packages, etc)
  "Use `find-tag' to find symbol at point, falling back to `find-func' features."
  (interactive)
  (let* ((fun (function-called-at-point)))
    (if (find-function-noselect fun 'no-c-sources)
        (find-function-do-it fun nil 'switch-to-buffer)
      (find-tag (symbol-name (symbol-at-point))))))
```

jpkotta:
If you're using etags, it's probably rebuilding the completion table (https://github.com/bbatsov/projectile/blob/master/projectile.el#L1796).  I commented out those lines and it sped up a lot.  I don't know why it needs to do that; maybe it could be cached for each project.  BTW, I'm using ido for completion.

fommil:
I'm using exuberant ctags and it's also rebuilding the completion table every time... which is very wasteful. I almost prefer emacs' basic `find-tag`.

colegleason:
This is really hurting. I have a large TAGS file and it sucks to wait for it to build every time. Any good way to speed it up?

fommil:
I replace it with built in find-tag

fommil:
[...]

Issue #557, opened 2014-11-27
etags-tags-completion-table: Stack overflow in regexp matcher


chalmagean:
I've ran regenerate tags file and then `C-c p j` and I'm getting this:

```elisp
Debugger entered--Lisp error: (error "Stack overflow in regexp matcher")
  re-search-forward("^\\(\\([^]+[^-a-zA-Z0-9_+*$:]+\\)?\\([-a-zA-Z0-9_+*$?:]+\\)[^-a-zA-Z0-9_+*$?:]*\\)\\(\\([^\n]+\\)\\)?\\([0-9]+\\)?,\\([0-9]+\\)?\n" nil t)
  etags-tags-completion-table()
  #[0 "\303\211C\304\305\"\210\212\306\242\205\307!\2038	 \262\211\242\2031\310\311\312\313\314\315!\316\"\317\320%\"\210\202	\211\240\210\202	)\304\321\"\210\211\242\211\207" [buffer-file-name tags-completion-table-function tags-completion-table nil message "Making tags completion table for %s..." visit-tags-table-buffer t mapatoms make-byte-code 257 "\301\302!\300\242\"\207" vconcat vector [intern symbol-name] 4 "\n\n(fn SYM)" "Making tags completion table for %s...done"] 9 "\n\n(fn)"]()
  funcall(#[0 "\303\211C\304\305\"\210\212\306\242\205\307!\2038	 \262\211\242\2031\310\311\312\313\314\315!\316\"\317\320%\"\210\202	\211\240\210\202	)\304\321\"\210\211\242\211\207" [buffer-file-name tags-completion-table-function tags-completion-table nil message "Making tags completion table for %s..." visit-tags-table-buffer t mapatoms make-byte-code 257 "\301\302!\300\242\"\207" vconcat vector [intern symbol-name] 4 "\n\n(fn SYM)" "Making tags completion table for %s...done"] 9 "\n\n(fn)"])
  tags-completion-table()
  projectile-find-tag()
  funcall-interactively(projectile-find-tag)
  call-interactively(projectile-find-tag nil nil)
  command-execute(projectile-find-tag)
```

GNU Emacs 24.4.50.1 (x86_64-apple-darwin13.3.0, NS appkit-1265.21 Version 10.9.4 (Build 13E28)) of 2014-09-15

marcinant:
:+1:

fommil:
I'm still getting this. So the the external etags-select is not taking over tab completion of the tags table. Damn.

[...]

PR #936, opened 2015-12-24
support etags-select close #910 and #683 #668 #557 #699

### LINKED ISSUE #910: support for etags-select
I recently discovered etags-select, which is the best way yet that I've seen to search etags without installing a large framework like helm.

I've found that the following is enough to get the support integrated with the tag table built by projectile, incase anybody wants to do similar:

```elisp
(defun projectile-etags-select-find-tag ()
  "Run `etags-select-find-tag' in the current projectile context."
  (interactive)
  (projectile-visit-project-tags-table)
  (etags-select-find-tag))
```

of course it would be great if this was the default behaviour for `C-c p j`

fommil:
Could somebody who uses ggtags please test this? I'm guessing the behaviour there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,11 @@
 * Consider `TAGS` and `GTAGS` root markers.
 * Add relation between the `.h`, `.cxx`, `.ixx` and `.hxx` files in `projectile-other-file-alist`.
 * Add support to specify project name either via `.dir-locals.el` or by providing a customized `projectile-project-name-function'.
-* Prefer ag's internal .gitignore parsing
+
+### Changes
+
+* Prefer ag's internal .gitignore parsing.
+* Always use external find-tag implementations.
 
 ## 0.13.0 (10/21/2015)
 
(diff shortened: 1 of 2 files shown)


bbatsov:
You'll have to rebase this.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Consider `TAGS` and `GTAGS` root markers.
 * Add relation between the `.h`, `.cxx`, `.ixx` and `.hxx` files in `projectile-other-file-alist`.
 * Add support to specify project name either via `.dir-locals.el` or by providing a customized `projectile-project-name-function'.
+* Use external find-tag implementations.
```
bbatsov: To some extent we've done this in the past well, so here you should probably say "always use"...

This should be in changes, not new features. Guess you can add two entries - one about the etags-select here and one about dropping projectile's tag parsing in changes.

fommil:
rebasing...

fommil:
Thanks! Can you please close all the linked tickets?

PR #937, opened 2015-12-27
Check prefix *and* suffix when finding test files

mdorman:
When adding `projectile-find-matching-test` support for emacs projects,
I found emacs projects have no standard regarding prefix or suffix---but
`projectile-test-affix` meant a prefix would always hide a suffix if you
listed both.

I believe the fact the result of `projectile-test-affix` was always
tested as both a prefix or a suffix speaks to its lack of value: better
to simply use the result of `projectile-test-prefix` as a prefix or the
result of `projectile-test-suffix` as a suffix directly.

`projectile-find-matching-test` and `projectile-find-matching-file` were
the only users of `projectile-test-affix`, so I simply eliminated it,
after modifying both of them to use the prefixes and suffixes properly.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1697,12 +1697,6 @@ It assumes the test/ folder is at the same level as src/."
   (find-file
    (projectile-find-implementation-or-test (buffer-file-name))))
 
-(defun projectile-test-affix (project-type)
-  "Find test files affix based on PROJECT-TYPE."
-  (or (funcall projectile-test-prefix-function project-type)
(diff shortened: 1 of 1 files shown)


mdorman:
Ping?

bbatsov:
Sorry, I'm super busy at work. I don't even remember why we introduced the concept of the affix. Your changes look sound. Right now the bigger problem is the test prefices and suffices are not part of the dynamic project registration.

PR #938, opened 2015-12-27
Add very basic emacs Cask project support.

mdorman:
This doesn't enable much more than switching between implementation and test files, but every little bit would seem to help.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Drop and the `.` in the commit message (titles are not sentences) and add a changelog entry.

mdorman:
Done.  I should mention that this is somewhat dependent on #937 to be most useful.

bbatsov:
Thanks. I'll look into the other ticket more closely later.

PR #944, opened 2016-01-10
Edit .dir-locals.el file of a project.

akermu:
Add a command `projectile-edit-dir-locals' to quickly edit or create a
.dir-locals.el file for the project.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2640,6 +2640,41 @@ is chosen."
 
 (projectile-commander-bindings)
 
+(defun projectile-read-variable ()
+  "Prompt for a variable and return its name."
+  (completing-read "Variable: "
+                   obarray
+                   '(lambda (v)
+                      (and (boundp v) (not (keywordp v))))
+                   t))
+
+(define-skeleton projectile-skel-variable-cons
+  "Insert a variable-name and a value in a cons-cell."
+  "Value: "
+  "("
+  (projectile-read-variable)
+  " . "
+  str
+  ")")
+
+(define-skeleton projectile-skel-dir-locals
+  "Insert a .dir-locals.el template."
+  nil
+  "((nil . (("
+  ("" '(projectile-skel-variable-cons) \n)
+  resume:
+  "))))")
+
+(defun projectile-edit-dir-locals ()
+  "Edit or create a .dir-locals.el file of the project."
+  (interactive)
+  (let ((file (expand-file-name ".dir-locals.el" (projectile-project-root))))
+    (find-file file)
+    (when (not (file-exists-p file))
+      (unwind-protect
+          (projectile-skel-dir-locals)
+        (save-buffer)))))
+
 ;;; Minor mode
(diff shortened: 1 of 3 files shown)


bbatsov:
Such functionality should definitely be part of the readme.

projectile.el
```diff
@@ -2640,6 +2640,41 @@ is chosen."
 
 (projectile-commander-bindings)
 
+(defun projectile-skel-read-variable ()
```
bbatsov: Why `skel`?

projectile.el
```diff
@@ -2640,6 +2640,41 @@ is chosen."
...
+  "Prompt for a variable and return its name."
+  (completing-read "Variable: "
+                   obarray
+                   '(lambda (v)
+                      (and (boundp v) (not (keywordp v))))
+                   t))
+
+(define-skeleton projectile-skel-variable-cons
```
bbatsov: I've never seen this `define-skeleton`.

projectile.el
```diff
@@ -2640,6 +2640,41 @@ is chosen."
...
+  nil
+  "((nil . (("
+  ("" '(projectile-skel-variable-cons) \n)
+  resume:
+  "))))")
+
+(defun projectile-edit-dir-locals ()
+  "Edit or create a .dir-locals.el file of the project."
+  (interactive)
+  (let ((file (concat (projectile-project-root) ".dir-locals.el")))
```
bbatsov: Instead of concat use expand-file-name.

projectile.el
```diff
@@ -2640,6 +2640,41 @@ is chosen."
...
+    (find-file file)
```
bbatsov: This code is a bit obscure - I'd create the file explicitly if I see it hasn't been created.

projectile.el
```diff
@@ -2658,6 +2693,7 @@ is chosen."
     (define-key map (kbd "d") #'projectile-find-dir)
     (define-key map (kbd "D") #'projectile-dired)
     (define-key map (kbd "e") #'projectile-recentf)
+    (define-key map (kbd "E") #'projectile-edit-dir-locals)
```
bbatsov: This should also be added to the menu.

bbatsov:
The commit message's title should not end with a `.`.

projectile.el
akermu: Because it is used in the skeleton below, but it can be used generally, so i should probably rename it.

projectile.el
akermu: See https://www.gnu.org/software/emacs/manual/html_node/autotype/Skeleton-Language.html
it is a builtin template language.

projectile.el
akermu: How would you do that, elisp file operations are a little bit obscure?

akermu:
I updated the PR, thanks for the review!

bbatsov:
Thanks!

PR #946, opened 2016-01-14
backport emacs 25 etags-tags-completion-table

fommil:
Workaround for https://github.com/bbatsov/projectile/issues/557 (Emacs 20703)

It now opens up the way for #947

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 ### Bugs fixed
 
 * [#871](https://github.com/bbatsov/projectile/issues/871): Stop advice for `compilation-find-file` to override other advices.
+* [#557](https://github.com/bbatsov/projectile/issues/557): stack overflow in `projectile-find-tag'.
 
 ## 0.13.0 (10/21/2015)
 
--- a/projectile.el
+++ b/projectile.el
@@ -80,6 +80,35 @@ attention to case differences."
         (and (>= start-pos 0)
              (eq t (compare-strings suffix nil nil
                                     string start-pos nil ignore-case))))))
+
+  ;; Improved (no more stack overflows) in Emacs 24.5
+  (eval-after-load 'etags
+    '(when (< emacs-major-version 25)
+       (defvar etags--table-line-limit 500)
+       (defun etags-tags-completion-table ()
+         (let (table
+               (progress-reporter
+                (make-progress-reporter
+                 (format "Making tags completion table for %s..." buffer-file-name)
+                 (point-min) (point-max))))
+           (save-excursion
+             (goto-char (point-min))
+             (while (not (eobp))
+               (if (not (re-search-forward
+                         "[\f\t\n\r()=,; ]?\177\\\(?:\\([^\n\001]+\\)\001\\)?"
+                         (+ (point) etags--table-line-limit) t))
+                   (forward-line 1)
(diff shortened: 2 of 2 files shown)


fommil:
argh, this isn't working. My emacs-lisp foo is not strong enough. This function isn't overriding the one in `etags.el` unless I explicitly `C-x C-e` on it.

fommil:
ok, worked it out. I need to quote the whole function that runs in the `eval-after-load` because macro. I can't modify my commit right now but I will do so when I am at home tonight. Remind me if I forget (also, note to self: typo in `nill` and also rename the `defcustom` to be consistent.)

fommil:
ok, fixed. @bbatsov this should now be good if it passes your review. It now opens up the way for #947

projectile.el
```diff
@@ -502,7 +507,35 @@ just return nil."
         version)
     (error "Cannot determine version without package pkg-info")))
 
-
+;;; Workarounds
+
+(eval-after-load 'etags
+  '(when projectile-etags-tags-completion-table-hack
```
bbatsov: Why don't you just add the version check here?

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 ### Bugs fixed
 
 * [#871](https://github.com/bbatsov/projectile/issues/871): Stop advice for `compilation-find-file` to override other advices.
+* [#557](https://github.com/bbatsov/projectile/issues/557): stack overflow in `projectile-find-tag'
```
bbatsov: This should end with a `.`.

bbatsov:
The commit message should be capitalized.

projectile.el
```diff
@@ -502,7 +507,35 @@ just return nil."
         version)
     (error "Cannot determine version without package pkg-info")))
 
-
+;;; Workarounds
+
+(eval-after-load 'etags
```
bbatsov: I think there should be a comment here explaining the need for this. Btw, we already have a place in the code where we're backporting stuff from newer Emacsen.

projectile.el
```diff
@@ -502,7 +507,35 @@ just return nil."
         version)
     (error "Cannot determine version without package pkg-info")))
 
-
+;;; Workarounds
```
bbatsov: This should be in the already existing compatibility section.

projectile.el
fommil: Because the way I've done it the user can turn it off, or even forward port the impl in the future. Or we can Toggle it during testing.

CHANGELOG.md
fommil: You'll have to wait for that change as I'm not near a computer.

projectile.el
fommil: Which section is that?

bbatsov: See the beginning of the file.

fommil: ok, I'll try to update this PR tonight / tomorrow. Ping me if I forget (it would be very useful if github allowed one to mark a PR / ticket as "unread")

dgutov:
So the new code improves the stack overflow situation? Cool.

fommil:
Bloody github, totally forgot this PR.

fommil:
@bbatsov updated. I removed the `defcustom` and just added the backport for old emacsen.

PR #954, opened 2016-01-26
Make `projectile-find-matching-{test/file}` respect `projectile-test-*-function`

jguenther:
These two functions did not use the value of `projectile-test-prefix/suffix-function`. This changes `projectile-find-matching-test/file` to call the function stored in `projectile-test-prefix/suffix-function`, which default to `projectile-test-prefix/suffix`, respectively.

I'd like to add tests for this as well but don't have time right now, but I might be able to get around to it this week or next.


bbatsov:
:+1:

Issue #955, opened 2016-01-30
Error while toggling between test and source file.


sandor-juhasz:
I have a git project with the following files:
```
test/demo-test.el
Cask
demo.el
demo.org
```

After opening `test/demo-test.el`, I want to switch to the `demo.el` using `C-c p t` but I get the following error message "Wrong type argument: stringp, nil"

The error is caused in the `projectile-dirname-matching-count` as the relative path candidates `demo.el` and `demo.org` does not have directory parts. Modifying the expression inside the `split-string`  and defaulting to empty string using a simple `or` expression solves the issue. 

I'm creating a pull request to fix this.

PR #956, opened 2016-01-30
 Error while toggling between test and source file.

sandor-juhasz:
As described in https://github.com/bbatsov/projectile/issues/955 I rand into difficulties when I was switching from the test file to the source file using `C-c p t`. 

When a matching count is calculated for files in the project root, the parameter `b` in the `projectile-dirname-matching-count` function will not have directory part, just the relative filename. The `file-name-directory` returns nil what the `split-string` cannot interpret, leading to the error "Wrong type argument: stringp, nil". Please see the added test case.

Solution: defaulting to empty string in this case solves the problem. I have added this safety check to the parameter `a` just in case although it seems always to be an absolute path at the moment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 
 * [#871](https://github.com/bbatsov/projectile/issues/871): Stop advice for `compilation-find-file` to override other advices.
 * [#557](https://github.com/bbatsov/projectile/issues/557): stack overflow in `projectile-find-tag'.
+* [#955](https://github.com/bbatsov/projectile/issues/955): Error while toggling between test and source file.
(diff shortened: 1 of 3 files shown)


bbatsov:
The change looks fine. Just squash your two commits into one.

sandor-juhasz:
Done, thanks for the review! Please let me know if any other modifications are needed, this is my first pull request ever. ;-)

bbatsov:
:+1: Thanks!

> Done, thanks for the review! Please let me know if any other modifications are needed, this is my first pull request ever. ;-)

Happy to hear this! Hopefully it will be the first of many! :-)

PR #957, opened 2016-02-01
Do not error inside of `projectile-cache-current-file` when opening a file inside a project.

jbenden:
When opening a file from the terminal, it is possible for the
file to be a project and have the `switch-to-buffer` hook run,
with a new buffer that has no associated file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 * [#955](https://github.com/bbatsov/projectile/issues/955): Error while toggling between test and source file.
 * [#952](https://github.com/bbatsov/projectile/issues/952): VCS submodules brought in even thought not descendent of project root.
 * [#576](https://github.com/bbatsov/projectile/issues/576): `projectile-replace` stomps regular expressions.
+* [#957](https://github.com/bbatsov/projectile/pull/957): When opening a specified file from the terminal, do not error inside of `projectile-cache-current-file`.
 
 ## 0.13.0 (10/21/2015)
 
--- a/projectile.el
+++ b/projectile.el
@@ -666,8 +666,8 @@ The cache is created both in memory and on the hard drive."
   "Add the currently visited file to the cache."
   (interactive)
   (let ((current-project (projectile-project-root)))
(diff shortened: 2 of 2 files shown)


bbatsov:
I'm not quite certain what you mean. Some exact repro steps would be useful.

jbenden:
It's a nil condition with the line patched. Simple fix.

-Joe

bbatsov:
When I looked closer at the code I understood the problem, but the description of the commit and the changelog entry are misleading - in reality we simply don't want to attempt to cache a buffer which is not backed by a file.

projectile.el
```diff
@@ -657,7 +657,7 @@ The cache is created both in memory and on the hard drive."
   "Add the currently visited file to the cache."
   (interactive)
   (let ((current-project (projectile-project-root)))
-    (when (gethash (projectile-project-root) projectile-projects-cache)
+    (when (and (gethash (projectile-project-root) projectile-projects-cache) (buffer-file-name (current-buffer)))
```
bbatsov: I'd move the file check before the cache check. Also `(buffer-file-name (current-buffer))` can be replaced with simply `buffer-file-name`. No need for two function calls. I see the same was used a line below, which can be optimized.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 
 * Prefer ag's internal .gitignore parsing.
 * Always use external find-tag implementations.
+* When opening a specified file from the terminal, do not error inside of `projectile-cache-current-file`.
```
bbatsov: This should be in bugs fixed.

bbatsov:
ping

bbatsov:
There are two entries in the changelog now. Remove the old one, rebase on top of the current `master` and squash the commits into one.

jbenden:
Please review rebase.

bbatsov:
I still see two entries in the changelog and the branch is still outdated compared to the current upstream `master`.

jbenden:
Sorry. Hopefully ok now.

PR #958, opened 2016-02-02
Add a simple command for forgetting old projects.

### LINKED ISSUE #950: Allow editing projectile-known-projects
I use `projectile-switch-project` heavily, but it eventually accumulates test copies of projects that I don't want to reuse.

Poking through the code, I can there's `projectile-ignored-projects` and `projectile-ignored-project-function`. These only seem to apply before a project is added to `projectile-known-projects`.

However, I can't see any interactive way of saying 'forget about project foo_copy'. Ideally projectile would offer something similar to `recentf-edit-list` for interactive removal of old projects from `projectile-known-projects`.

Have I missed some clever way of doing this? Would you be interested in such a feature?

Wilfred:
Fixes #950

This is a very simple implementation, but it meets my immediate needs. What do you think?

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -477,6 +477,12 @@ The saved data can be restored with `projectile-unserialize'."
   "List of locations where we have previously seen projects.
 The list of projects is ordered by the time they have been accessed.")
 
+(defun projectile-forget-project (project)
+  "Remove a project from `projectile-known-projects'."
+  (interactive (list (completing-read "Forget project: "
+                                      projectile-known-projects)))
+  (delete project projectile-known-projects))
+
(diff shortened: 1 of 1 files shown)


bbatsov:
This command is totally redundant. There's already `projectile-remove-known-project`. I thought the point of #950 was some more complex UI to edit the known projects.

Wilfred:
Oh, I had no idea about `projectile-remove-known-project`. I was thinking that it would be nice to have a sophisticated UI, but I thought there was currently no way of doing this.

I'll send a docs fix to make `projectile-remove-known-project` more discoverable.

PR #961, opened 2016-02-06
Make projectile work with fossil ls

Gei0r:
Hi,

projectile-find-other-file did not work with fossil. After investigating, I figured out that `fossil ls` gives a file list separated by `\n`, but projectile expects them separated by `\0`.

This changes the default projectile-fossil-command to give a zero-separated file list, just like the git and svn commands.

Unfortunately, `fossil ls` doesn't have a `-0` switch like `hg` and `bzr`.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -868,7 +868,7 @@ Files are returned as relative paths to the project root."
   :group 'projectile
   :type 'string)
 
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1: Thanks!

PR #963, opened 2016-02-09
Mention projectile-remove-known-project.

Wilfred:
This handy function isn't mentioned anywhere, so add it to the docstring
to aid discoverability.

(Following on from #950)

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -475,7 +475,10 @@ The saved data can be restored with `projectile-unserialize'."
 
 (defvar projectile-known-projects nil
   "List of locations where we have previously seen projects.
-The list of projects is ordered by the time they have been accessed.")
+The list of projects is ordered by the time they have been accessed.
+
+See also `projectile-remove-known-project',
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -475,7 +475,10 @@ The saved data can be restored with `projectile-unserialize'."
 
 (defvar projectile-known-projects nil
   "List of locations where we have previously seen projects.
-The list of projects is ordered by the time they have been accessed.")
+The list of projects is ordered by the time they have been accessed.
+
+See also `projectile-remove-known-project' to interactively
```
bbatsov: Guess you should also mention here the clear and cleanup functions as well.

Wilfred: Aha, I didn't know those existed either! `projectile-cleanup-known-projects` is what I want most of the time, so this PR has been worthwhile just to learn about that :)

Wilfred:
@bbatsov updated, are you happy to merge?

PR #968, opened 2016-02-20
fix test of `projectile-cleanup-known-projects'

spwhitton:
This test can fail if user-emacs-dir does not exist.  So make a
temporary known projects file, as is done in the test of
`projectile-merge-known-projects' in projectile-known-project-test.el.

I am currently packaging projectile for Debian, and the Debian package autobuild infrastructure sets $HOME to /nonexistent, causing this test to fail.

### DIFF
--- a/test/projectile-test.el
+++ b/test/projectile-test.el
@@ -182,7 +182,8 @@
       (should (equal (projectile-relevant-known-projects) '("/path/to/project2"))))))
 
 (ert-deftest projectile-test-projects-cleaned ()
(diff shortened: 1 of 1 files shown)


bbatsov:
> I am currently packaging projectile for Debian

Sweet!

Can you just capitalize the commit message?

spwhitton:
On Sun, Feb 21, 2016 at 12:43:33AM -0800, Bozhidar Batsov wrote:
> Can you just capitalize the commit message?

Done.

-- 
Sean Whitton

bbatsov:
Thanks!

PR #969, opened 2016-02-21
Ignore projectile-globally-ignored-suffixes in projectile-grep.

### LINKED ISSUE #727: Files in projectile-globally-ignored-file-suffixes not ignored by projectile-grep
I have many build files ignored by projectile by using `projectile-globally-ignored-file-suffixes`. Many of these are minified JavaScript files: `.min.js`. When using `projectile-grep` these files are grepped and produce false positives in the grep buffer. Further, because these files have all text on one line, it can slow Emacs down to a crawl.

I think it would be great if `projectile-grep` ignored files using `project-ignored-file-suffixes`.

jdufresne:
Fixes #727

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Add a command to edit the .dir-locals.el file of the project (`projectile-edit-dir-locals`).
 * Add file local variable `projectile-project-root`, which allows overriding the project root on a per-file basis. This allows navigating a different project from, say, an org file in a another git repository.
 * Add `projectile-grep-finished-hook`.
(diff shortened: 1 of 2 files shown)


bbatsov:
I'm guessing this is also an issue with `ag` as well.

jdufresne:
@bbatsov I added the same support for `ag`. I'm not very familiar with this tool nor do I ever use it. I have tested the `grep` version thought. So please let me know if the `ag` change looks sane. Thanks!

bbatsov:
I guess this should work. :-)

bbatsov:
I've merged this manually.

PR #970, opened 2016-02-23
Allow user the choice of not using file true names

kaushalmodi:
- Allow user the choice to use or not use true names of file/dir everywhere (when finding project root, when caching files, etc)
- Also deleted an orphan function projectile-file-truename; wasn't used anywhere.

(Reference: https://github.com/bbatsov/projectile/commit/924d73120bca6adc5b9bf3d79ad53075a63b5f1e )

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -408,6 +408,16 @@ Any function that does not take arguments will do."
   :group 'projectile
   :type 'function)
 
+(defcustom projectile-use-true-name t
+  "When non-nil, the symbolic links are followed through to get the true name of
+the file/directory.
+
+For example, when non-nil, true name of the current directory is used to find
+the project root (in `projectile-project-root') and the true name of the visiting
+file is added to the project cache (in `projectile-cache-current-file')."
+  :group 'projectile
+  :type 'boolean)
+
 
 ;;; Idle Timer
 (defvar projectile-idle-timer nil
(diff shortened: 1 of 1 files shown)


bbatsov:
What's the point of this change?

kaushalmodi:
As explained in that linked reference, using file-truename is a misfeature for me. The VC system we use makes all the checked-in project files as symlinks to read-only files in a cache. So the true names end up like this: `../../../sos_cache/prj_name/prj_name_sos_cache/prj_name/FILES/dir1#dir1_1#dir1_1_1#file_name_v_1.23_456789`.

So I would have to have projectile not try to get the file/dir true names for the purpose of file caching and finding project roots.

We often have symlinks to a read-only project in a different root altogether and tracing the truename of that dir results in "Not a projectile project error". So for my application, it is good to turn of truename searching altogether.

> The earlier behavior was a feature for me and this change is a bug.
> 
> The revision control system we use saved all file revisions in a central cache and a user, if not intending to modify a file, simply accesses a symbolic link to that file in cache.
> 
> In that case, the truename is horrendous: ../../../sos_cache/prj_name/prj_name_sos_cache/prj_name/FILES/dir1#dir1_1#dir1_1_1#file_name_v_1.23_456789
> 
> I would like to have a defvar to switch to the earlier behavior.

projectile.el
```diff
@@ -789,12 +805,6 @@ The current directory is assumed to be the project's root otherwise."
             (error "You're not in a project")
           default-directory))))
 
-(defun projectile-file-truename (file-name)
```
bbatsov: I'd really have the `if` check just here and use this wrapper everywhere. A self-contained change is the best possible change.

bbatsov:
See my inline remark. Also - this should be mentioned in the changelog and probably in the README.

kaushalmodi:
Finally this PR is not needed any more. I found a much cleaner solution .. simply advising `file-truename`.

```el
;; Do not return truenames for files checked into SOS.
;; This causes inconvenience in packages like `projectile' which store the
;; truenames of files for storing in project file caches.
(defun advice/sos-cache-no-file-truename (orig-fun &rest args)
  "Do not return truename for sos_cache files, because they are typically
very useful to the end user. An sos_cache file name typically looks like this:
/sos_cache/PRJ/PRJ_sos_cache/PRJ/FILES123/dir1#dir2#dir3#file.ext_branch.version_hash."
  (let ((fname (when args (car args)))
        (fname-true (apply orig-fun args)))
    (if (string-match-p "\\`/sos_cache" fname-true)
        fname
      fname-true)))
(advice-add 'file-truename :around #'advice/sos-cache-no-file-truename)
```

Issue #943, opened 2016-01-07
projectile-shell command


jamienicol:
I find it really useful having a command to open a shell in the root dir of a project, or if a shell is already open for the project, jump to it. I quickly threw this together and it works, though I'm no elisp expert so I'm sure it could be improved:

    (defun projectile-shell ()
      (interactive)
      (when (projectile-project-root)
        (let* ((default-directory (projectile-project-root))
               (buffer-name (format "*shell*<%s>" (projectile-project-name)))
               (buffer (or (get-buffer buffer-name)
                           (shell buffer-name))))
          (switch-to-buffer buffer))))

Silex:
:+1: 

This could be later improved to support `eshell` and `term`, similar to like the search interface supports `grep` and `ag`.

bbatsov:
This can be simplified using `projectile-with-default-dir`.

bbatsov:
Fixed by #973.

PR #973, opened 2016-03-04
Add a command to run shell in project's root directory

danskarda:
When working with multiple projects, I find it useful to have different shells for each project. This patch adds a command to launch shell in current root directory.

The shell buffer will be called `*shell project-name*`.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -48,12 +48,17 @@
   (defvar ggtags-completion-table)
   (defvar tags-completion-table)
   (defvar tags-loop-scan)
-  (defvar tags-loop-operate))
+  (defvar tags-loop-operate)
+  (defvar eshell-buffer-name)
+  (defvar explicit-shell-file-name))
 
 (declare-function ggtags-ensure-project "ggtags")
 (declare-function ggtags-update-tags "ggtags")
 (declare-function pkg-info-version-info "pkg-info")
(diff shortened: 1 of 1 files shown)


bbatsov:
That's a good idea. I'd add 2 more similar commands - one that starts `term` and one that starts `eshell` (probably using a common keyboard prefix). You should also mention this in the changelog.

danskarda:
Add term / eshell. There are two options:

1.  add customization using dispatch var, where user makes choice between three alternatives
2.  add keyboard prefix (`C-c p x t`, `C-c p x e` and `C-c p x s`)

I am conservative user and 100% time stick to one shell, so I would prefer 1. What do you think?

Add changelog : good catch, thanks.

Once you think about solution for var vs dispatch key, I will prepare the commit and squash everything to original patch for final review.

bbatsov:
I'm leaning towards option two, as I'm not sure what would be the name of a generic command governed by a defcustom in this case.

bbatsov:
Or maybe we can have just one command invoking a shell or an eshell (with a prefix arg).

bbatsov:
P.S. This will solve #943.

bbatsov:
ping :-)

danskarda:
pong:-)

Implemented. Rebased. Squashed.

PR #974, opened 2016-03-04
Associate .cc and .hpp files for C++

holguinj:
Some C++ projects use the .hpp suffix for headers and .cc for source
files, and `projectile-find-other-file` function previously would not
work in this case. This commit updates the alist of associated file
extensions to link .cc and .hpp files in both directions.


bbatsov:
:+1: Thanks!

PR #975, opened 2016-03-06
RFC: Use `find` for listing files in subversion repositories, for a massive speedup

Wilfred:
`svn list` queries the remote server to find out what files are
included. This is extremely slow for repositories with thousands of
files. It also requires a network connection.

For example, given a checkout of the LLVM source code:

    $ svn co http://llvm.org/svn/llvm-project/llvm/trunk llvm
    ... lots of files
     U   llvm
    Checked out revision 262801.
    $ cd llvm

Using `svn` takes almost two minutes, whereas `find` takes under a
second:

    $ time svn list -R . > files.txt
    svn list -R . > files.txt  0.79s user 0.13s system 0% cpu 1:50.73 total

    $ time find . -type f > find_files.txt
    find . -type f > find_files.txt  0.02s user 0.02s system 98% cpu 0.040 total

Let me know what you think. I appreciate it's less elegant (it doesn't respect svn ignores) but the speedup is massive. Even with caching, waiting 2 minutes for projectile to build an index is painful.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -890,7 +890,7 @@ Files are returned as relative paths to the project root."
   :group 'projectile
   :type 'string)
 
-(defcustom projectile-svn-command "svn list -R . | grep -v '$/' | tr '\\n' '\\0'"
+(defcustom projectile-svn-command "find . -type f | grep -v '/.svn/' | tr '\\n' '\\0'"
   "Command used by projectile to get the files in a svn project."
   :group 'projectile
   :type 'string)


bbatsov:
I don't like this approach, as it deviates from all other defaults. The command is configurable for a reason. I understand the default is slow, but I also think most users would expect their file list to be built using `svn` (especially if their not on a Unix OS).

Wilfred:
Fair enough, thanks for your feedback. It's no problem for me to set this variable just within my environment.

I'll open a fresh PR if there's a flag to `svn ls` that prevents network access, but it looks like no such flag exists.

Wilfred:
Turns out I set the default behaviour in #285! I'd totally forgotten. At the time some other users encountered similar issues (see the comments on 1304890963efd568b74f80622c2a6ab289d6f5d1).

PR #977, opened 2016-03-08
[Fix #576] Implement `projectile-replace-regexp'

### LINKED ISSUE #576: projectile-replace stomps regular expressions
I'll start with a question, because I'm a little unclear: Is `projectile-replace` supposed to support regular expressions?

This issue mostly has to do with improvements introduced in #269 by @shosti. `projectile-replace` uses `tags-query-replace`, which uses Emacs regular expressions. However, `projectile-files-with-string` stomps entered regexp under certain conditions. 

I'm working on Mac OS X (`(projectile-unixy-system-p)` → `t`). I have `ag`, `ack`, `git grep`, and `grep` (henceforth "grep tools") installed. Under these conditions, `ag` is selected, and then passed the `--literal` flag, removing meaning from my regexp query. That's what led me to the problem, but that's not the only issue.

Summary:

- Each of the grep tools has their own interpretation of regular expressions, and
- None of these match how Emacs (i.e., `tags-query-replace`) interprets them [AFAIK]. Furthermore,
- All of the grep tools can handle literal matches more or less the same, but
- `tags-query-replace` doesn't support literal matches.

It might be possible to somewhat accurately use the grep tools to speed up literal matches, providing a literal version of `tags-query-replace` could be created. However, I don't see a clear way that the file list could be narrowed for a regular expression replace without finding a tool that exactly emulates Emacs' regular expressions.

Though filtering files through a faster tool is a significant benefit, it comes at the cost of accuracy. This might not seem to be a big deal, but it's very unexpected when matches you think should be replaced aren't being replaced. In this case, I think accuracy trumps speed.

I'm willing to implement a solution if there is agreement on the desired behavior. Projectile rocks and I love using it! Thanks!

seanfisk:
This pull request implements `projectile-replace-regexp` [as discussed and suggested by](https://github.com/bbatsov/projectile/issues/576#issuecomment-66892727) @shosti, and closes #576. The new `projectile-replace-regexp` replaces text by regexp within project files using `tags-query-replace`. `projectile-replace` has been refactored with a modified version of `tags-query-replace` to replace only literal matches (no regexp at all). This makes the behavior of these functions consistent with Emacs' `query-replace` and `query-replace-regexp`.

As it is very difficult to match the behavior of Emacs' regexps with grep tools, `projectile-files-with-string` is not used with `projectile-replace-regexp`, and has been adapted to support literal strings only.

I am definitely not the best Elisp hacker, so if you have any suggestions, I'd be glad to implement them. Thanks to @shosti for the suggestions and @d12frosted for the prodding ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Add file local variable `projectile-project-root`, which allows overriding the project root on a per-file basis. This allows navigating a different project from, say, an org file in a another git repository.
 * Add `projectile-grep-finished-hook`.
 * Ignore file suffixes listed in `projectile-globally-ignored-file-suffixes` when using `projectile-grep` and `projectile-ag`.
+* Add `projectile-replace-regexp`, which supports replacement by regexp within a project. `projectile-replace` is now used solely for literal replacements.
 
 ### Changes
 
@@ -30,6 +31,7 @@
 * [#557](https://github.com/bbatsov/projectile/issues/557): stack overflow in `projectile-find-tag'.
 * [#955](https://github.com/bbatsov/projectile/issues/955): Error while toggling between test and source file.
 * [#952](https://github.com/bbatsov/projectile/issues/952): VCS submodules brought in even thought not descendent of project root.
+* [#576](https://github.com/bbatsov/projectile/issues/576): `projectile-replace` stomps regular expressions.
 
 ## 0.13.0 (10/21/2015)
 
--- a/projectile.el
+++ b/projectile.el
@@ -46,7 +46,9 @@
 (eval-when-compile
   (defvar ag-ignore-list)
   (defvar ggtags-completion-table)
-  (defvar tags-completion-table))
+  (defvar tags-completion-table)
+  (defvar tags-loop-scan)
+  (defvar tags-loop-operate))
 
 (declare-function ggtags-ensure-project "ggtags")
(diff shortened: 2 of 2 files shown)


bbatsov:
At a glance it looks ok. Thanks! :+1:

d12frosted:
@seanfisk thank you for implementing this. I will check it once new version of projectile is released.

Meanwhile I am using following workaround:

```
(defun projectile-replace-regexp ()
  "Replace a string in the project using `tags-query-replace'.
Less efficient than `projectile-replace' but at least allows
usage of regular expressions. See
https://github.com/bbatsov/projectile/issues/576 for more details
on `projectile-replace' issue with regexps."
  (interactive "P")
  (let* ((old-text (read-string
                    (projectile-prepend-project-name "Replace: ")
                    (projectile-symbol-or-selection-at-point)))
         (new-text (read-string
                    (projectile-prepend-project-name
                     (format "Replace %s with: " old-text))))
         (files (-map (lambda (f) (concat (projectile-project-root) f)) (projectile-current-project-files))))
    (tags-query-replace old-text new-text nil (cons 'list files))))
```

seanfisk:
@bbatsov You're welcome; thank you for creating Projectile! :)

@d12frosted Your code is essentially the same as mine, minus the directory prefix argument and Git submodules workaround. It should work the same way for most purposes, I would think. Thanks for reminding me about this issue!

PR #980, opened 2016-03-14
Add find-other-file support for indirect buffer

qqldd:
This change add `projectile-find-other-file` for indirect buffer that is created by `C-x 4 c`

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1345,7 +1345,7 @@ https://github.com/abo-abo/swiper")))
 With FLEX-MATCHING, match any file that contains the base name of current file.
 Other file extensions can be customized with the variable `projectile-other-file-alist'."
   (interactive "P")
-  (-if-let (other-files (projectile-get-other-files (buffer-file-name) (projectile-current-project-files) flex-matching))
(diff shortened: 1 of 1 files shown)


bbatsov:
Hmm, isn't the file name the same for the original and the indirect buffers?

qqldd:
@bbatsov Yes, it is the same. However, `buffer-file-name` cannot get the file name for indirect buffer. For example, the original buffer named A, we cloned a new buffer named AC. When using `buffer-file-name` on AC, it returns `nil`. Only the original buffer has the associated file name.

bbatsov:
So, I'm guessing this is a problem in other places of the codebase as well. Probably we should create a simple helper function that works on indirect buffers?

bbatsov:
Ping :-)

qqldd:
Just saw this pending request. I didn't use indirect buffer for a while. I will try to see if I can make an update to this.

bbatsov:
I'm closing this PR due to no recent activity. Feel free to re-open it if you ever come back to it.

PR #984, opened 2016-03-22
Ignore cached project roots that point to nonexistent directories.

Wilfred:
This fixes an awkward bug when the current project is a symlink that
changes. Suppose we have the following filesystem layout:

    /tmp/projectile_project -> /tmp/first_project

If we now delete /tmp/first_project and change the symlink:

    /tmp/projectile_project -> /tmp/second_project

then `projectile-project-root-cache` will contain a key
`"projectile-root-bottom-up-/tmp/projectile-project"` with value
`"/tmp/first_project"`. This will stop `projectile-find-file` from working
as it can't cd to /tmp/first_project.

Instead, we enforce that the string from `projectile-project-root`
points to a directory that currently exists.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -790,7 +790,7 @@ The current directory is assumed to be the project's root otherwise."
   (let ((dir default-directory))
     (or (--some (let* ((cache-key (format "%s-%s" it dir))
                        (cache-value (gethash cache-key projectile-project-root-cache)))
-                  (if cache-value
+                  (if (and cache-value (file-exists-p cache-value))
                       cache-value
                     (let ((value (funcall it (file-truename dir))))
                       (puthash cache-key value projectile-project-root-cache)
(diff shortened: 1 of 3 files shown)


bbatsov:
The change looks good. Note it in the changelog and drop the `.` from the commit's title.

projectile.el
```diff
@@ -790,7 +790,7 @@ The current directory is assumed to be the project's root otherwise."
   (let ((dir default-directory))
     (or (--some (let* ((cache-key (format "%s-%s" it dir))
                        (cache-value (gethash cache-key projectile-project-root-cache)))
-                  (if cache-value
+                  (if (and cache-value (file-exists-p cache-value))
```
bbatsov: Btw, this change might be problematic for tramp users, as such checks are very slow there.

Wilfred:
@bbatsov thanks, changelog and commit message updated.

PR #993, opened 2016-03-31
Clarify initial usage instructions

### LINKED ISSUE #992: Unclear instructions about enabling projectile-global-mode
I read through the whole readme, googled and googled and *googled*, but I could not figure out how to add a project to Projectile.  No matter what I did, how many times I opened a file in a git repo, deleted the save file, made sure no Projectile settings were customized, Projectile never remembered any projects that I opened files in.

Nowhere could I find a simple statement saying, "To add a project to Projectile..." or, "To make Projectile remember a project...".

Then finally, through trial and error, I figured out that the key is to enable `projectile-global-mode`.  Apparently, if a buffer has `projectile-mode` enabled, then Projectile will add its repo to the project list.  

Since several `projectile-` commands are autoloaded, users may expect that running one of them will activate Projectile, but this is not the case.

Please add a statement to the readme explaining this.  It's not currently clear that `projectile-mode` must be explicitly enabled for a buffer for Projectile to automatically recognize and remember projects, that running one of the autoloaded commands is not sufficient.

alphapapa:
Explain that projectile-mode must be activated for Projectile to remember projects.  Fixes #992.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,13 +88,14 @@ Users of Debian 9 or later or Ubuntu 16.10 or later may simply
 
 ### Basic setup
 
-You can enable Projectile globally like this:
+To add a project to Projectile, open a file in the project and enable `projectile-mode` in that buffer.  
+
+To make Projectile automatically remember projects that you access files in, enable Projectile globally:
 
 ```el
 (projectile-global-mode)
 ```
-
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1:

PR #995, opened 2016-04-07
[Fix #987] Add the ability to omit ignored patterns/paths

### LINKED ISSUE #987: projectile-ag ignores ag-ignore-list when projectile-project-vcs is git
## Expected behavior

`M-x projectile-ag query` ignores files that match ag-ignore-list
## Actual behavior

When the vcs for the project is git, `M-x projectile-ag query` ignores the ag-ignore-list

```
mkdir /tmp/foo
cd /tmp/foo
git init
echo waffle > waffle
echo waffle > waffle.min.js
```

in emacs - go /tmp/foo; eval (setq ag-ignore-list "*.min.js"); run projectile-ag waffle
get two results, one of which is waffle.min.js. 

to contrast

```
mkdir /tmp/bar
cd /tmp/bar
hg init
echo waffle > waffle
echo waffle > waffle.min.js
```

here, opening /tmp/bar, with the same ag-ignore-list "*.min.js"
only has one result

Projectile version: 20160321.2307
GNU Emacs 25.1.50.2 (x86_64-pc-linux-gnu, GTK+ Version 3.10.8) of 2016-03-13
## Suggested fix

The problem is in projectile-ag
`ag-ignore-list` is let-bound to nil by `(unless (eq (projectile-project-vcs) 'git)`, which is the problem.
Instead, in the function `projectile-ag` in `projectile.el`  (~ line 1970)

```
(let (ag-ignore-list (unless (eq (projectile-project-vcs) 'git)
                              ;; ag supports git ignore files
                              (-union ag-ignore-list
                                      (append
                                       (projectile-ignored-files-rel) (projectile-ignored-directories-rel)
                                       (projectile--globally-ignored-file-suffixes-glob)
                                       grep-find-ignored-files grep-find-ignored-directories)))))
```

should probably be something like

```
(ag-ignore-list      (-union ag-ignore-list
                                      (append
                                       (projectile-get-ignored-vcs-files-if-such-a-thing-exists)
                                       (projectile-ignored-files-rel) (projectile-ignored-directories-rel)
                                       (projectile--globally-ignored-file-suffixes-glob)
                                       grep-find-ignored-files grep-find-ignored-directories)))
```

laz2:

Thanks!

Add the ability to omit ignored patterns/paths for `projectile-ag` (controlled via the `projectile-ag-use-ignored` defcustom).

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -417,6 +417,11 @@ Any function that does not take arguments will do."
   :type 'hook
   :package-version '(projectile . "0.14.0"))
 
+(defcustom projectile-ag-use-ignored t
+  "If true, use ignored paths/patterns in search."
+  :group 'projectile
+  :type 'boolean)
+
 (defcustom projectile-test-prefix-function 'projectile-test-prefix
   "Function to find test files prefix based on PROJECT-TYPE."
   :group 'projectile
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -417,6 +417,11 @@ Any function that does not take arguments will do."
   :type 'hook
   :package-version '(projectile . "0.14.0"))
 
+(defcustom projectile-ag-use-ignored t
+  "If true, use ignored paths/patterns in search."
```
bbatsov: Probably you should expand this docstring to mention that we're talking about projectile's ignores and not VC ignores.

bbatsov:
I'll have to look at the original ticket first, as I'm not 100% what the intention of this change is. I also wonder if something like this doesn't make sense for other search tools as well.

PR #996, opened 2016-04-09
Support use of recompile

dougm:
In order to use the `recompile' command, both `compilation-directory'
and `compile-command' variables need to be set.  This is taken care of
when using the `compile' function, but not `compilation-start'.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2534,7 +2534,7 @@ Should be set via .dir-locals.el.")
   "Run external or Elisp compilation command CMD."
   (if (functionp cmd)
       (funcall cmd)
-    (compilation-start cmd)))
+    (compile cmd)))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


bbatsov:
I wonder if I had something in mind when I used `compilation-start` initially. Are the variables you mention some buffer-local variables?

dougm:
They are not buffer local (doc quoted below).  Use of `compilation-start` directly only seems useful if you want to set the optional vars (mode name-function highlight-regexp).  Which projectile doesn't do currently.

> compile-command is a variable defined in `compile.el'.
> ...
> Documentation:
> Last shell command used to do a compilation; default for next compilation.
> 

> compilation-directory is a variable defined in `compile.el'.
> ...
> Documentation:
> Directory to restore to when doing `recompile'.

PR #1004, opened 2016-04-25
Improve error message when file not found

phillord:

Thanks!

The project type is now displayed when a file a matching test or
implementation is not found, which is useful when the project type
has been mis-identified.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1946,14 +1946,17 @@ It assumes the test/ folder is at the same level as src/."
       (let ((impl-file (projectile-find-matching-file file-name)))
         (if impl-file
             (projectile-expand-root impl-file)
-          (error "No matching source file found")))
+          (error
+           (format "No matching source file found for project type: %s"
+                   (projectile-project-type)))))
     ;; find the matching test file
     (let ((test-file (projectile-find-matching-test file-name)))
       (if test-file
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -1946,14 +1946,17 @@ It assumes the test/ folder is at the same level as src/."
       (let ((impl-file (projectile-find-matching-file file-name)))
         (if impl-file
             (projectile-expand-root impl-file)
-          (error "No matching source file found")))
+          (error
+           (format "No matching source file found for project type: %s"
+                   (projectile-project-type)))))
```
bbatsov: Or we can simply display the project type at the modeline, which seems like a better idea to me.

phillord:
Nothing wrong with doing both. The mode-line is fine, but mine is overloaded and so I suspect is everyone elses. I don't know how many people diminish projectile, but I bet it's quite a few.

yveszoundi:
Like @phillord, I also doubt that most people diminish projectile. In my configs, I do use `diminish` or similar approaches to avoid cluttering the mode-line, for pretty much everything **but** projectile.

bbatsov:
OK, let's do both things then.

bbatsov:
Ping :-)

PR #1014, opened 2016-05-30
[Fix #1013] Path comparison on Windows

### LINKED ISSUE #1013: projectile-project-buffer-p may return incorrect result on Windows
`projectile-project-buffer-p` indicates whether a buffer is part of the current project. On Windows, it can sometimes (or always?) return a false negative due to path capitalization. This happens because it uses case-sensitive `string-prefix-p` to check whether the project root is a prefix of the full path to the currently visited file. As a result, functions like `helm-projectile` and `projectile-switch-to-buffer` will not let the user switch to a buffer of the same project.

## Expected behavior
`projectile-project-buffer-p` always returns `t` when a buffer is part of a project.

## Actual behavior
It returns `nil` on Windows.

1. Create a directory `C:\Home`.
2. Create a repository under `C:\Home\repo`.
3. Visit a file in that repository.
4. `M-: (projectile-project-buffer-p (current-buffer) (projectile-project-root)`

The last step will return `nil` when it should be `t` (this buffer is surely part of the project). This happens because in `projectile-project-buffer-p`, `project-root` will evaluate to `c:/Home/repo/demo.txt`, while `(file-truename default-directory)` will be `c:/home/repo/demo.txt`. 

Projectile version: 20160526.832

### Emacs version

GNU Emacs 24.5.1 (x86_64-w64-mingw32)

### Operating system

Windows 7

cshtarkov:
Use case-insensitive comparison on Windows to avoid false negatives from `projectile-project-buffer-p`. See #1013.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 * [#576](https://github.com/bbatsov/projectile/issues/576): `projectile-replace` stomps regular expressions.
 * [#957](https://github.com/bbatsov/projectile/pull/957): When opening a specified file from the terminal, do not error inside of `projectile-cache-current-file`.
 * [#984](https://github.com/bbatsov/projectile/pull/984): Error when a project is a symlink that changes target.
+* [#1013](https://github.com/bbatsov/projectile/issues/1013): projectile-project-buffer-p may return incorrect result on Windows.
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #1015, opened 2016-06-05
Register elixir project type

asok:
Makes possible to alternate between the test and the implementation files.


bbatsov:
Thanks!

PR #1017, opened 2016-06-07
Add projectile-discover-projects-in-directory

justbur:
This is a helper function to go through the subdirectories of a
directory, check which ones are valid projects and add them to the cache
file. I didn't make it recursive although it could be. 

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -721,6 +721,23 @@ The cache is created both in memory and on the hard drive."
                                           projectile-cache-file))
     (projectile-invalidate-cache nil)))
 
+;;;###autoload
+(defun projectile-discover-projects-in-directory (directory)
+  "Discover any projects in DIRECTORY and add them to the projectile cache.
+This function is not recursive and only adds projects with roots
+at the top level of DIRECTORY."
+  (interactive
+   (list (read-directory-name "Starting directory: ")))
+  (let ((subdirs (directory-files directory t)))
+    (mapcar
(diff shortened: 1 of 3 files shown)


justbur:
I triggered a new CI build because it looks like the last one failed because a server was down. 

By the way, a bunch of tests fail on windows, at least some of them can be fixed by using canonical paths, but I'm not sure if you care.

bbatsov:
This should probably be added at least to the menu, otherwise no one is going to find out about it.

bbatsov:
A mention in the manual would be useful as well.

justbur:
@bbatsov ok done

projectile.el
```diff
@@ -721,6 +721,22 @@ The cache is created both in memory and on the hard drive."
...
+at the top level of DIRECTORY."
+  (interactive
+   (list (read-directory-name "Starting directory: ")))
+  (let ((subdirs (directory-files directory t)))
+    (mapcar
+     (lambda (dir)
+       (when (and (file-directory-p dir)
+                  (not (member (file-name-nondirectory dir) '(".." "."))))
+         (let ((default-directory dir))
+           (projectile-cache-projects-find-file-hook))))
```
bbatsov: This is not correct. You should simply check whether you're in a project dir and if so add this dir to the list of known projects.

justbur:
Thanks, please check it now

PR #1018, opened 2016-06-07
Add check for dirty projects

muahah:
Add two functions to check if Git projects are dirty
(uncommitted changes and/or unpushed commits).
Use [vc-check-status](https://github.com/thisirs/vc-check-status) package.

`projectile-check-project-dirty` check if the project associated with the given directory is dirty.
`projectile-see-dirty-projects` list all dirty projects (very useful before leaving Emacs after a day of hard work)

I used it for a while now and it work flawlessly for me.
If the community is interested, I can work on it a little more 
(add tests, maybe support for others vcs... ).

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 * New command `projectile-run-term` (<kbd>C-c p x t</kbd>).
 * Let user unignore files in `.projectile` with the ! prefix.
 * Add a command to add all projects in a directory to the cache (`projectile-discover-projects-in-directory`).
+* Add a command to list dirty projects (for version controlled projects)
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@
 * New command `projectile-run-term` (<kbd>C-c p x t</kbd>).
 * Let user unignore files in `.projectile` with the ! prefix.
 * Add a command to add all projects in a directory to the cache (`projectile-discover-projects-in-directory`).
-* Add a command to list dirty projects (for version controlled projects)
+* Add a command to list dirty version controlled projects (`projectile-browse-dirty-projects`).
 
 ### Changes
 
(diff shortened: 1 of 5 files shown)


bbatsov:
Seems like a useful addition, but it should definitely support multiple VCS. I avoid adding features that are specific just for a certain VC (even if it's as popular as git).

bbatsov:
I'm also a bit concerned about having to depend on another package just for this one command. I prefer to limit the number of dependencies.

muahah:
I will check if its possible to rely on `vc` built-in functions instead of `vc-check-status`.
It should solve both issues.

bbatsov:
Any progress here?

muahah:
Unfortunately no,
I keep it in mind but don't have time to work on it right now.

muahah:
In fact I took a look to the available functions of `vc`, 
it seems pretty straightforward to implement this functionality with `vc-dir`.
I should be able to come with something soon.

bbatsov:
Excellent!

muahah:
It needs some cleaning, but it's working for git.
It should also work with any VCS supported by `vc`, but I only use git so I was not able to test that.

muahah:
I cleaned up a little.
Some remarks concerning this version :
- It seems to be slower than the previous version specific to git.
- It's doesn't check for unpulled/unpushed commits (`vc` doesn't seem to check that...).
- It should work with every VCS supported by `vc`, but I only tested it with git.

Any suggestions of improvement ?

muahah:
Note : 
the first use of this function can be (really) slow,
It's mainly because `vc-dir` check for every file of the repository.
I don't see any (easy) way of making this faster.
Maybe checking the repository status asynchroneously could be a solution if it become a real problem.

muahah:
I made the proposed modifications.
`projectile-browse-dirty-project` should now appear in the menu and in the manual.
Also added `C-c p V` as key-binding for `projectile-browse-dirty-projects` ('V' for version control).
And made the list of checked states customizable (`projectile-vcs-dirty-states`).

bbatsov:
👍

PR #1023, opened 2016-06-30
Inform the user once TAGS has been regenerated

Wilfred:
This is a small change, but I prefer getting feedback on when TAGS generation is done. At the moment, I end up pressing keys thinking 'is Emacs done yet?' as some of my projects take a few seconds to generate TAGS.


bbatsov:
👍

PR #1034, opened 2016-07-26
Do not consider backup files for projectile-get-other-file.

jchmrt:
A minor annoyance for me while using projectile is that when I use `projectile-find-other-file`, it will often prompt me which file to use because it counts backup files as candidates. For example you have a project with 3 files, test.c, test.h and test.h~. If you are then editing test.c and use `projectile-find-other-file`, it will present both test.h and test.h~ as options, while the user probably wouldn't want to edit a backup file. So I think it would be better if this function ignored backup files.

The function didn't already ignore backup files, because it uses `file-name-extension`, which ignores backup suffixes. This pull request fixes this issue by filtering the candidates on not being backup files. I have also added a backup file to the files in the `projectile-test-get-other-files`, so that it also tests that the backup files aren't included.

I added a line to the changelog about the change, but I didn't edit the README, because I don't think this change should be detailed there.

Thanks!

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* Ignore backup files in `projectile-get-other-files`.
+
 ### Bugs fixed
 
 * [#1024](https://github.com/bbatsov/projectile/issues/1024): Do not cache ignored project files.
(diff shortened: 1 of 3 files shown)


bbatsov:
👍 Thanks!

PR #1035, opened 2016-07-27
[Fix #1033]  Move defcustom projectile-go-function above its usage.

### LINKED ISSUE #1033: projectile-compile-project: Symbol's function definition is void: projectile-go-function
Likely introduced by c2785d636804cd23b90eec4388cf4e5624f5e003.

## Expected behavior
No error, the prompt for the compilation command should appear.

## Actual behavior
Error.

Call `M-x projectile-compile-project`.

```
Projectile version: 20160723.622
```

### Emacs version

```
$ emacs-snapshot --version
GNU Emacs 25.0.50.1
```

### Operating system

Ubuntu 14.04

etanzapinsky:
I found out that the sharp quote is necessary and indicates to the compiler that `projectile-go-function` is variable referencing a function.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [#1024](https://github.com/bbatsov/projectile/issues/1024): Do not cache ignored project files.
 * [#1022](https://github.com/bbatsov/projectile/issues/1022): Scan for Fossil's checkout DB, not its config DB.
 * [#1007](https://github.com/bbatsov/projectile/issues/1007): Make use of `projectile-go-function`.
+* [#1033](https://github.com/bbatsov/projectile/issues/1033): projectile-compile-project: Symbol's function definition is void: projectile-go-function
 
 ## 0.14.0 (2016-07-08)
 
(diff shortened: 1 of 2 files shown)


shwsun:
Yes, it solved the ``symbol variable is void`` error. Thx

tee3:
Thanks for this!  I believe that only the `defcustom` of the `projectile-go-function` variable needs to be moved earlier since the `projectile-go` function is only used later.  Does that makes sense?

Note that without this change, the byte compilation of `ibuffer-projectile` as well as `projectile` itself fails when installing from MELPA.  Note that I only saw this when reinstalling the `projectile` package, not when updating via `package`.

etanzapinsky:
@tee3 I agree. I think it's a good idea to move the `projectile-cabal` function above for semantic consistency, that all the functions that find file type groups stay together.

tee3:
@etanzapinsky okay, that makes sense to me too

bbatsov:
I didn't see this PR and committed pretty much the same fix myself. Sorry about this.

PR #1046, opened 2016-08-26
Replace dash with cl-lib

abo-abo:
Using e.g. the standard `cl-remove-if` in place of 4 aliases (`-remove`, `--remove`, `-reject`, `--reject`) improves code quality and eases new contributions and refactoring.

I've also rewritten `Makefile`, because any Cask-based `make test` stopped working on my system some time ago (and it fails intermittently on Travis CI as well).

I have more stuff on top of this PR in my develop branch, I'll open a new PR if this gets merged.

-----------------

Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 emacs ?= emacs
 bemacs = $(emacs) -batch -l test/elpa.el
 
-elpa: compile test
+elpa: update compile test
 
 update:
 	$(emacs) -batch -l test/make-update.el
--- a/Makefile
+++ b/Makefile
@@ -1,39 +1,18 @@
-EMACS ?= emacs
-EMACSFLAGS =
-CASK = cask
+emacs ?= emacs
+bemacs = $(emacs) -batch -l test/elpa.el
 
-OBJECTS = projectile.elc
+all: compile test
 
-elpa:
-	$(CASK) install
-	$(CASK) update
-	touch $@
+update:
+	$(emacs) -batch -l test/make-update.el
 
-.PHONY: build
-build : elpa $(OBJECTS)
+compile:
+	$(bemacs) -l test/make-compile.el
 
-.PHONY: byte-compile-strict
-byte-compile-strict : elpa
-	$(CASK) exec $(EMACS) --no-site-file --no-site-lisp --batch \
-		--directory "."                          \
-		$(EMACSFLAGS)                            \
-		--eval "(progn                           \
-			(setq byte-compile-error-on-warn t)  \
-			(batch-byte-compile))" projectile.el
+test:
+	$(bemacs) -l test/run-tests
 
-.PHONY: test
-test : byte-compile-strict
-	$(CASK) exec $(EMACS) --no-site-file --no-site-lisp --batch \
-		$(EMACSFLAGS) \
-		-l test/run-tests
+clean:
+	rm -f *.elc
 
-.PHONY: clean
-clean :
-	rm -f $(OBJECTS)
-	rm -f elpa
-	rm -rf .cask # Clean packages installed for development
-
-%.elc : %.el
-	$(CASK) exec $(EMACS) --no-site-file --no-site-lisp --batch \
-		$(EMACSFLAGS) \
-		-f batch-byte-compile $<
+.PHONY: all update compile test clean
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 emacs ?= emacs
 bemacs = $(emacs) -batch -l test/elpa.el
 
-all: compile test
+elpa: compile test
 
 update:
 	$(emacs) -batch -l test/make-update.el
--- a/doc/contributing.md
+++ b/doc/contributing.md
@@ -63,12 +63,11 @@ You can support the development of Projectile via
 
 ## Running the tests in batch mode
 
-Install [cask](https://github.com/cask/cask) if you haven't
-already, then:
-
 ```
 $ cd /path/to/projectile
-$ cask
+$ make update
+$ make compile
+$ make test
 ```
 
 Run all tests with:
@@ -77,5 +76,4 @@ Run all tests with:
 $ make test
 ```
 
-(Note: tests may not run correctly inside Emacs' `shell-mode` buffers. Running
-them in a terminal is recommended.)
+Tests should run fine in `shell-mode` or `term-mode`. It's also possible to use <kbd>M-x</kbd> `compile` (or `helm-make`).
--- a/projectile.el
+++ b/projectile.el
@@ -1845,10 +1845,10 @@ With a prefix ARG invalidates the cache first."
 
 (defun projectile-test-file-p (file)
   "Check if FILE is a test file."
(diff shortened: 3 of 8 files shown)


bbatsov:
While I do agree with the spirit of the change I do wonder if it wouldn't be a better idea to use the new `seq.el` instead of `cl-lib`. `seq.el` is available in ELPA as well and will be part of the upcoming Emacs release. Its API is more consistent IMO.

test/elpa.el
```diff
@@ -0,0 +1,5 @@
+(setq package-user-dir
+      (expand-file-name (format ".cask/%s/elpa" emacs-version)))
```
bbatsov: If the idea is to drop the cask dependency, why do we use a folder `.cask` something?

abo-abo:
> While I do agree with the spirit of the change I do wonder if it wouldn't be a better idea to use the new seq.el instead of cl-lib. seq.el is available in ELPA as well and will be part of the upcoming Emacs release. Its API is more consistent IMO. 

Well, `seq.el` is just another API for the uninitiated (myself included) to learn before hacking starts. I've been using `cl-lib` for years, and it's used in core Emacs as well.

In any case, I made the translation firstly for myself. I needed to make some functional changes, and the code wasn't readable to me before I switched to `cl-lib`.

But just now, looking at `seq-difference` I like the idea that it uses `equal` by default, unlike `cl-set-difference`. Then again, last time I checked `cl-lib` was faster than `seq`.

test/elpa.el
abo-abo: I like the idea of Cask. The problem with it is it often fails to do its job (current error is `there's no package s- when downloading dependencies`).

No need to invent another dotfile to put `elpa/` in, `.cask` will do, it's familiar.

bbatsov: > I like the idea of Cask. The problem with it is it often fails to do its job (current error is there's no package s- when downloading dependencies).

I'm reasonably sure this was some bug in `package.el` on Emacs's master branch. I thought this was resolved in newer builds.

abo-abo: Does Cask still have problem being used from `ansi-term`? That was another issue I had earlier.

In any case, I think it's more flexible to tweak a few lines of an Elisp build target rather than relying on a separate Python script and passing data from Makefile through shell interface.

For instance, see `(setq byte-compile--use-old-handlers nil)`. Looks better in an Elisp file rather than being passed through `--eval` in a Makefile.

Makefile
```diff
@@ -1,39 +1,18 @@
-EMACS ?= emacs
-EMACSFLAGS =
-CASK = cask
+emacs ?= emacs
+bemacs = $(emacs) -batch -l test/elpa.el
```
bbatsov: bemacs? :-)

test/elpa.el
bbatsov: > Does Cask still have problem being used from ansi-term? That was another issue I had earlier.

Guess @lunaryorn or @rejeep can answer this.

I'm fine with dropping cask in general, but I liked the idea of having some solution that makes it easy to have an uniform test setup across different projects.

lunaryorn: I am not aware that Cask had any issues in ansi-term nor can I remember any such issues reported.

But then I do not use ansi-term and am not in a position to answer.

If you could report the ansi-term issue to our tracker we might be able to try and fix it.

lunaryorn: I'm aware of the current issue with Emacs snapshot but that's an issue in Emacs (Emacs 25 rc works fine) and beyond our control. Unfortunately that happens from time to time 😒

lunaryorn: Finally I'd prefer if you did not use .cask to install packages manually.  We make no guarantees about the layout of that folder, and I'd not like to see users being confused by a name that refers to a tool that's not being used.

If anything breaks in your code it given the name it might end up in our tracker and generate additional work for us.

Makefile
abo-abo: batch emacs, yes.

abo-abo:
Any more comments on this? Just saw the new C `mapcan` added to Emacs and thought about this PR.

projectile.el
```diff
@@ -1250,10 +1250,10 @@ Only buffers not visible in windows are returned."
 
 (defun projectile-normalise-paths (patterns)
   "Remove leading `/' from the elements of PATTERNS."
-  (-non-nil (--map (and (string-prefix-p "/" it)
-                        ;; remove the leading /
-                        (substring it 1))
-                   patterns)))
+  (-non-nil (mapcar (lambda (it) (and (string-prefix-p "/" it)
```
bbatsov: It'd rename all those `it` arguments to something meaningful. This name was required for `--something` macros but we can certainly do better.

bbatsov:
I've added just one bit of comment - rename the `it` args from `--something` macros to something more meaningful. I'm also guessing we can replace `projectile-difference` with the `seq.el` version of the function, but that's not a big deal.

test/elpa.el
bbatsov: @abo-abo I guess it'd be best to use a different folder for this indeed.

abo-abo:
> I'm also guessing we can replace projectile-difference with the seq.el version of the function, but that's not a big deal.

This is a big deal when thinking about which minimal version of Emacs Projectile wants to support, and if Projectile wants to depend on seq for Emacs-24 versions.

doc/contributing.md
```diff
@@ -63,12 +63,11 @@ You can support the development of Projectile via
 
 ## Running the tests in batch mode
 
-Install [cask](https://github.com/cask/cask) if you haven't
-already, then:
-
 ```
 $ cd /path/to/projectile
-$ cask
+$ make update
```
abo-abo: This could also be a single `make elpa`, if you prefer.

lunaryorn:
Just FYI, I'm unsubscribing from this thread again, so please mention me again if need be.

bbatsov:
👍  I think that's good for now.

bbatsov:
> This is a big deal when thinking about which minimal version of Emacs Projectile wants to support, and if Projectile wants to depend on seq for Emacs-24 versions.

I think it was 24.1 if memory serves. This has to be checked and documented one of these days...

Fuco1:
> Well, seq.el is just another API for the uninitiated (myself included) to learn before hacking starts.

Just like people not using `cl-lib` will now have to learn that.  Looking at some diffs the new cl code is hardly more readable than the old code.  CL is an archaic api most people writing in "new" languages won't understand or know.  I don't understand this change at all.

Fuco1:
-  (--sort (> (car it) (car other))
    -          (--group-by (projectile-dirname-matching-count file it) candidates)))
    +  (cl-sort (lambda (it other) (> (car it) (car other)))
    +           (copy-sequence
    +            (let (value result)
    +              (while (setq value (pop candidates))
    +                (let* ((key (projectile-dirname-matching-count file value))
    +                       (kv (assoc key result)))
    +                  (if kv
    +                      (setcdr kv (cons value (cdr kv)))
    +                    (push (list key value) result))))
    +              (mapcar (lambda (x)
    +                        (cons (car x) (nreverse (cdr x))))

How is this an improvement?

Issue #523, opened 2014-11-11
projectile-global-mode prevents new processes from working over TRAMP


Silex:
I don't have time to investiguate it right now but I should be able to this week.

Silex:
Okay, I found interesting stuffs:

``` elisp
(setq projectile-mode-line '(:eval (format " Projectile[%s]" (file-truename default-directory))))
(setq projectile-mode-line '(:eval (format " Projectile[%s]" default-directory)))
```

The first line creates the problem, while the 2nd one doesn't.
Apparently, `file-truename` doens't work properly when the TRAMP connection is in progress...

@thomasf, @tarsius: I can't find a way to detect wether TRAMP is ready to be used... I tried the following but it still exhibits the bug:

``` elisp
(setq projectile-mode-line '(:eval (if (or (not (tramp-tramp-file-p default-directory)) (tramp-connectable-p default-directory))
                                         (format " Projectile[%s]" (file-truename default-directory))
                                       " Projectile[NOOOO]")))
```

It looks like `tramp-connectable-p` returns true when it should not.

Silex:
Hum, `with-timeout` seems to work, but it's a bit hacky:

``` elisp
(setq projectile-mode-line '(:eval (with-timeout (0.2 " Projectile[NOOO]")
                                     (format " Projectile[%s]" (file-name-nondirectory (f-parent default-directory))))))
```

Silex:
I'm trying to find answers at http://lists.gnu.org/archive/html/bug-gnu-emacs/2015-01/msg00596.html for information.

bgvianyc:
It probably worth pointing out that projectile was working normally with TRAMP until I installed tern.

Silex:
If `tern` spawns a process then it happens because of `:lighter` because of a TRAMP bug.

bgvianyc:
disabled tern still an issue. For the uninitiated, what is :lighter?

thomasf:
The lighter is the `projectile-mode-line` customization which produces the text that is displayed in the mode line.

Silex:
@bgvianyc: `(setq projectile-mode-line " Projectile")` fixes all of these issues for me.

timvisher:
Adding a :+1: for this. Anything I can do to help diagnose the issue, just let me know. I've applied @Silex's workaround and it works so I'm at least back up off the floor.

Silex:
[...]

PR #1047, opened 2016-08-30
Unify modifications to find-file-hook

abo-abo:
Currently, `projectile-mode` adds 3 functions to `find-file-hook`. Moreover, it does so locally, while most people use `projectile-global-mode`. This means it's really hard to suppress the modifications to the hook, since it would have to be done for many buffers separately.

I propose to add only a single function to `find-file-hook` that calls all 3. Moreover, consider modifying `find-file-hook` globally, not locally.

One more thing: projectile's modification to `find-file-hook` makes it really slow to open a file on two unrelated remotes that I have: around 20 second delay difference. Should there be a variable to disable the hook functions on a remote file, or do you want to do it uniformly?

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3300,6 +3300,13 @@ entirely."
   :risky t
   :package-version '(projectile "0.12.0"))
 
+(defun projectile-find-file-hook-function ()
+  "Called by `find-file-hook' when `projectile-mode' is on."
+  (unless (file-remote-p default-directory)
+    (projectile-cache-files-find-file-hook)
+    (projectile-cache-projects-find-file-hook)
+    (projectile-visit-project-tags-table)))
+
 ;;;###autoload
 (define-minor-mode projectile-mode
   "Minor mode to assist project management and navigation.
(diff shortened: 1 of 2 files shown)


Silex:
About the remote part, please check #523

The main issue with projectile and remote files seems to be the ticket, but I guess that a way to disable the ticker and the hooks for remote files would solve a lot of problems.

projectile.el
```diff
@@ -3300,6 +3300,13 @@ entirely."
   :risky t
   :package-version '(projectile "0.12.0"))
 
+(defun projectile-find-file-hook-function ()
+  "Called by `find-file-hook' when `projectile-mode' is on."
+  (unless (file-remote-p default-directory)
```
bbatsov: As mentioned in the ticket - the lighter causes the remote delay. It has nothing to do with those hooks.

abo-abo: I got a nice speedup with just this change. Maybe fixing the lighter would speed it up further, but this change is needed anyway, I think.

Of course, some people might prefer for the hook to work anyway on the remote, if their remote is fast. So a custom var can be added for this.

But adding 1 function to `find-file-hook` is in any case better than adding 3. Much simpler to understand and customize.

bbatsov:
@abo-abo ping :-)

abo-abo:
@bbatsov I agree that the root of the problem that I'm trying to solve is indeed `projectile-global-mode`. Thanks for pointing that out, @Silex.

The easiest solution that I can think of is to not use this:

```elisp
(define-globalized-minor-mode projectile-global-mode
  projectile-mode
  projectile-mode)
```

The way I see it, the only shape in which `projectile-mode` is useful is `projectile-global-mode`. I suggest to make `projectile-mode` already global by default and have `projectile-global-mode` be an alias to it.

The current PR is orthogonal to the above proposed change. I still think it's sound and can be merged.

Issue #1005, opened 2016-04-26
current buffer is incorrectly considered part of project to switch


ericdanan:
## Expected behavior

When I am in a buffer and switch to a project to which this buffer does not belong, with `projectile-switch-project-action` bound to `projectile-kill-buffers` (or using `projectile-commander` and selecting this action), this buffer should not be killed.

## Actual behavior

The buffer is killed.

## Steps to reproduce the problem

As stated above:
1. Start emacs
2. Bind `projectile-switch-project-action` to `projectile-kill-buffers`
3. Open a buffer A belonging to some project P and a buffer B not belonging to it.
4. Select buffer B
5. Call `projectile-switch-project` and select project P
Result: both buffers A and B are killed, but only A should be.

## Comments

This seems to happen because `projectile-switch-project-by-name` let-binds the current buffer's `default-directory` to the root of the project to switch, so this buffer is incorrectly considered as belonging the project to switch when `projectile-switch-project-action` is called.

The simplest solution I can think of would be to modify `projectile-switch-project-by-name` by:
1. Applying the change I proposed in #413 (call `projectile-switch-project-action` from a temporary buffer instead of the current buffer),
2. Additionally, no longer let-binding `default-directory` at the beginning of the function.

Note that this would cause `projectile-before-switch-project-hook` to be run with `default-directory` being that of the current buffer, not the root of the project to switch as it is now. I don't know if this is good or bad (seems natural to me since it is called "before switch"), but if it is a problem one could perhaps run `projectile-before-switch-project-hook` in the temp buffer as well or let-bind `default-directory` around it only.

As I said in #413, I can prepare a PR if you like.

### Projectile version information

```
Projectile version: 0.14
```

### Emacs version

```
GNU Emacs 24.5.1 (x86_64-unknown-cygwin)
```

### Operating system

```
Windows 7
```

PR #1051, opened 2016-09-12
load dir-locals before switching project

### LINKED ISSUE #413: `projectile-switch-project-action` unsettable by .dir-locals.el
Recently I tried to use `.dir-locals.el` to set `projectile-switch-project-action` to `magit-status`. However, the dir-locals in that project are not read or enabled until *after* a file/directory within that project has been opened.

I see a few ways of addressing this. The simplest way would be to open dired inside a project *before* invoking `projectile-switch-project-action`. However, a user might not want a dired of their project root.

Another way would be to manually invoke the dir-locals functionality. This seems a little tricky, because it seems as though it would require creating another buffer of some sort. We might be able to directly call the functions that apply dir-local variables, but since these are all buffer-locals, would they then apply to the minibuffer? Maybe a `with-temp-buffer` that has been CD'd into the project directory would be sufficient.

A third and much more drastic solution would be to implement a projectile-specific dir-locals equivalent. I almost don't want to do this, but my hunch is that we will keep running into places where dir-locals is strange/doesn't work.

ericdanan:
Modified function `projectile-switch-project-by-name`. Instead of calling `switch-project-action` from the current buffer, this is done in a temporary buffer after calling `hack-dir-local-variables-non-file-buffer`.

I think this fixes #413 and #1005 . It is the same solution I proposed there, but maybe easier to test for you with a PR.

Sorry I don't know how to add tests or `make test`. I don't think the changelog or readme should be updated since the function is not "visible".

Thanks

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2843,12 +2843,14 @@ With a prefix ARG invokes `projectile-commander' instead of
 Invokes the command referenced by `projectile-switch-project-action' on switch.
 With a prefix ARG invokes `projectile-commander' instead of
 `projectile-switch-project-action.'"
-  (let* ((default-directory project-to-switch)
-         (switch-project-action (if arg
+  (let ((switch-project-action (if arg
                                     'projectile-commander
                                   projectile-switch-project-action)))
     (run-hooks 'projectile-before-switch-project-hook)
-    (funcall switch-project-action)
+    (with-temp-buffer
+      (setq default-directory project-to-switch)
+      (hack-dir-local-variables-non-file-buffer)
+      (funcall switch-project-action))
     (run-hooks 'projectile-after-switch-project-hook)))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -2843,12 +2843,14 @@ With a prefix ARG invokes `projectile-commander' instead of
...
 With a prefix ARG invokes `projectile-commander' instead of
 `projectile-switch-project-action.'"
-  (let* ((default-directory project-to-switch)
-         (switch-project-action (if arg
+  (let ((switch-project-action (if arg
                                     'projectile-commander
                                   projectile-switch-project-action)))
     (run-hooks 'projectile-before-switch-project-hook)
-    (funcall switch-project-action)
+    (with-temp-buffer
```
bbatsov: I'd add some comments explaining here the need for this.

projectile.el
```diff
@@ -2843,12 +2843,14 @@ With a prefix ARG invokes `projectile-commander' instead of
...
+      (setq default-directory project-to-switch)
```
bbatsov: Messing with the `default-directory` is a very bad idea. Just let-bind it.

ericdanan:
Done, thanks.

PR #1055, opened 2016-09-17
Change symbol to function.

Fuco1:
by the way, `make test` is broken on master

    [16:28:08]matus@herakleitos:~
    > cd .emacs.d/dev/projectile/
    [22:41:57]matus@herakleitos:~/.emacs.d/dev/projectile
    > make test
    emacs -batch -l test/elpa.el -l test/run-tests
    Cannot open load file: no such file or directory, noflet
    make: *** [test] Error 255

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -263,7 +263,7 @@ Two example filter functions are shipped by default -
 `projectile-buffers-with-file' and
 `projectile-buffers-with-file-or-process'."
   :group 'projectile
-  :type 'symbol)
+  :type 'function)
 
 (defcustom projectile-project-name nil
   "If this value is non-nil, it will be used as project name.
@@ -278,7 +278,7 @@ It has precedence over function `projectile-project-name-function'."
 
 If variable `projectile-project-name' is non-nil, this function will not be used."
   :group 'projectile
-  :type 'symbol
+  :type 'function
   :package-version '(projectile . "0.14.0"))
 
 (defcustom projectile-project-root-files
@@ -425,7 +425,7 @@ it for functions working with buffers."
 
 Any function that does not take arguments will do."
   :group 'projectile
-  :type 'symbol)
+  :type 'function)
 
 (defcustom projectile-find-dir-includes-top-level nil
   "If true, add top-level dir to options offered by `projectile-find-dir'."


bbatsov:
> by the way, make test is broken on master

@abo-abo can you look into this?

abo-abo:
@bbatsov Still runs fine for me. And for Travis CI, since this request passed.

    make update && make test

Fuco1:
Why are we doing this instead of just using cask?

```elisp
(setq package-user-dir
      (expand-file-name (format ".elpa/%s/elpa" emacs-version)))
(message "installing in %s ...\n" package-user-dir)
(package-initialize)
(setq package-archives
      '(("melpa" . "http://melpa.org/packages/")
        ("gnu" . "http://elpa.gnu.org/packages/")))
(package-refresh-contents)

(defconst dev-packages
  '(helm noflet ag))

(dolist (package dev-packages)
  (unless (package-installed-p package)
    (ignore-errors
     (package-install package))))

(save-window-excursion
  (package-list-packages t)
  (condition-case nil
      (progn
        (package-menu-mark-upgrades)
        (package-menu-execute t))
    (error
     (message "All packages up to date"))))
```

Further, if calling `make update` is necessary it should be added to documentation.

abo-abo:
> Why are we doing this instead of just using cask?

These are 15 lines of Elisp that always work, are easy to understand and customize. I think it's really cool that automated package management can be achieved with so little code.

Using Cask means relying on a dependency in order to do basic things like compiling and testing.  A dependency that's proven to be unreliable in the past: I've had Cask breakage many times both on my own computers and on Travis CI.

> Further, if calling make update is necessary it should be added to documentation.

I agree. I did not want to make the `test` target depend on `update` target to speed things up, so that you can compile and test without contacting internet all the time.

Fuco1:
It's 22 lines of imperative code which can also randomly break when package.el changes versus 3 lines of declarative code.

This code is not obvious to many including me and I would have to look at some documentation to understand it.  Declarative code like "depends on a,b,c" is readily understandable.

If you had a bad experience with Cask I'm sorry but you might be the only one.

Pretty much every language nowadays uses some declarative package management system, and Cask was a huge leap forward for Emacs package development.  Going back to spagetti code is not an answer.  Go and make Cask better instead.

abo-abo:
> Going back to spagetti code is not an answer.

Insulting my code won't get you anywhere. This code has proven to work without error for the last year, unlike Cask.

> This code is not obvious to many including me and I would have to look at some documentation to understand it.

This is a ridiculous and faulty reasoning. The code conforms to package.el API and it's obvious at a glance what it does.

> Pretty much every language nowadays uses some declarative package management system, and Cask was a huge leap forward for Emacs package development. Going back to spagetti code is not an answer. 

Disagree.

> Go and make Cask better instead.

I've raised a couple issues with Cask years ago, they're still unfixed. From the code above, it's obvious how little extra value Cask adds to what package.el already provides.

Fuco1:
Fine.

Issue #1063, opened 2016-10-04
Fossil checkouts on Windows are not supported.


jaccarmac:
## Expected behavior

Projectile adds Fossil checkouts to its list of projects.

`projectile-find-file` in a Windows fossil checkout provides a list of files tracked by the repository.

## Actual behavior

Fossil checkouts on Windows are not added to the list.

`projectile-find-file` provides a list of files, but each name has a `^M` character appended to it.

## Steps to reproduce the problem

Use Projectile in a Fossil checkout on Windows.

### Projectile version information

```
Projectile version: 20160830.138
```

### Emacs version

25.1.1

### Operating system

Windows 10

bbatsov:
Fixed by #1064.

PR #1028, opened 2016-07-15
[Fix #1022] Fix Fossil detection

michaelbartnett:
The fossil checkout database is named `.fslckout` or `_FOSSIL_`, not
`.fossil`. Searching for a dominating .fossil file would just find
`~/.fossil`, which is Fossil's user config database.

I made the changes `projectile-project-vcs` such that when it searches for a Fossil repo checkout in the current directory or looks for the dominating project file, it tries `.fslckout`, then `_FOSSIL_` for the reasons described above.

@npostavs identified the needed changes in #1022's comment thread.

There were no existing tests for `projectile-project-vcs`, and figured having Fossil as a test dependency would be no good, so I did not add or change any tests for this.

I did a manual:

1. Created a Fossil repo and checked it out, and ran `touch .projectile`.

    ```
    testfossil1/
      test.fossil
      .fslckout
      .projectile
    ```

2. Ran `fossil add .projectile; fossil commit -m "add projetcile file"; touch test.txt`

    ```
    testfossil1/
      test.fossil
      .fslckout
      .projectile
      test.txt <- new file
    ```
  
3. Verified that `projectile-find-file` listed `.projectile` but not `test.txt`.

-----------------

Before submitting a PR make sure the following things have been done (and denote this
by checking the relevant checkboxes):

- [X] The commits are consistent with our [contribution guidelines](../CONTRIBUTING.md)
- [X] You've added tests (if possible) to cover your change(s)
- [X] All tests are passing (`make test`)
- [X] The new code is not generating bytecode or `M-x checkdoc` warnings
- [X] You've updated the changelog (if adding/changing user-visible functionality)
- [X] You've updated the readme (if adding/changing user-visible functionality)

Thanks!

PR #1064, opened 2016-10-04
Fix Fossil support on Windows.

jaccarmac:
See #1063 for the problems I faced using Projectile in a Windows Fossil repository and the commit messages below for my thought process re: fixes.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -952,7 +952,11 @@ Files are returned as relative paths to the project root."
   :group 'projectile
   :type 'string)
 
-(defcustom projectile-fossil-command "fossil ls | tr '\\n' '\\0'"
+(defcustom projectile-fossil-command (concat "fossil ls | "
+                                             (when (string-equal system-type
+                                                                 "windows-nt")
+                                               "dos2unix | ")
+                                             "tr '\\n' '\\0'")
   "Command used by projectile to get the files in a fossil project."
   :group 'projectile
   :type 'string)
--- a/projectile.el
+++ b/projectile.el
@@ -317,6 +317,7 @@ The topmost match has precedence."
     ".git"        ; Git VCS root dir
     ".hg"         ; Mercurial VCS root dir
     ".fslckout"   ; Fossil VCS root dir
+    "_FOSSIL_"    ; Fossil VCS root DB on Windows
     ".bzr"        ; Bazaar VCS root dir
     "_darcs"      ; Darcs VCS root dir
     )
@@ -371,6 +372,7 @@ containing a root file."
     ".git"
     ".hg"
     ".fslckout"
+    "_FOSSIL_"
     ".bzr"
     "_darcs"
     ".tox"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Bugs fixed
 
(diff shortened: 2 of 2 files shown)


projectile.el
```diff
@@ -950,7 +951,11 @@ Files are returned as relative paths to the project root."
   :group 'projectile
   :type 'string)
 
-(defcustom projectile-fossil-command "fossil ls | tr '\\n' '\\0'"
+(defcustom projectile-fossil-command (concat "fossil ls | "
+                                             (when (string-equal system-type
```
bbatsov: This `when` will return `nil` on non-windows systems. Don't recall if `concat` handles `nil` or not.

projectile.el
```diff
@@ -317,6 +317,7 @@ The topmost match has precedence."
     ".git"        ; Git VCS root dir
     ".hg"         ; Mercurial VCS root dir
     ".fslckout"   ; Fossil VCS root dir
+    "_FOSSIL_"    ; Fossil VCS root DB on Windows
```
jaccarmac: I'll go back through the history and find those bugfix commits to make sure; This may be a case for another `when` as below. What I do know is that `_FOSSIL_` is the traditional name for the checkout DB and the default on Windows today. The Fossil docs note that the convention may change in the future, but as of now the latest progress in that direction is a change from August 2012 that made `.fslckout` the default on Unix only (See version 1.23 in the changelog).

Something I didn't realize until now was that `_FOSSIL_` can be safely renamed to `.fslckout` even on Windows, giving Windows users an alternate workaround. This seems inelegant to me, however, and I'd prefer to get some form of this change merged, maybe with special logic for Windows systems.

https://www.fossil-scm.org/xfer/doc/trunk/www/tech_overview.wiki
https://www.fossil-scm.org/index.html/doc/trunk/www/changes.wiki

jaccarmac: The only commit involving `_FOSSIL_` (according to `git log -S'_FOSSIL'`) is the one resulting from #1028, which replaced `.fossil` with `.fslckout` and `_FOSSIL_` cases in a few limited cases. The existing code that allows Fossil repos with `.fslckout` to work is from 8576b3a776f7e8554d5bf3865f03d9a1de23951e, but that change didn't take into account `_FOSSIL_`.

What it looks like to me is that Fossil support was implemented incompletely and incorrectly in 2013, and earlier this year the incorrect parts were fixed while leaving the support not quite right on Windows. If I'm missing anything, please let me know!

projectile.el
bbatsov: Another idea to make the code a bit clear would be to have an `if` returning the whole command string for Windows and non-windows.

jaccarmac: If you think that would be better I can change it. I decided on `concat` to avoid duplicating the lion's share of the string.

jaccarmac:
Missed one more place where `.fslckout` was used without `_FOSSIL_` beside it. This can probably be amended onto 93c37c59666cee41fe572cf440ecc4bd9e45e8f1 when the PR is ready to be merged.

bbatsov:
OK, guess we're more or less ready here. Just squash the related commits together and I'll merge this PR.

jaccarmac:
Just force-pushed a cleaned-up version of the original sequence of commits? The related change helped make the first commit message summary a bit more readable as well.

bbatsov:
Thanks!

PR #1065, opened 2016-10-05
Extend the completion interface

abo-abo:
The `projectile-completing-read` interface now gets an optional `:action` arg that can be passed to the completion system (Ivy or Helm). For other completion systems, the behavior stays as before.

Now, functions like `projectile-find-file` or `projectile-switch-project` support advanced completion features such as:

- `ivy-next-line-and-call` or `helm-follow-action-forward`.
- `ivy-resume` or `helm-resume`.
- `ivy-occur` or the Helm equivalent.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1489,36 +1489,37 @@ Never use on many files since it's going to recalculate the
 project-root for every file."
   (expand-file-name name (projectile-project-root)))
 
-(defun projectile-completing-read (prompt choices &optional initial-input)
+(cl-defun projectile-completing-read (prompt choices
+                                      &key initial-input action)
   "Present a project tailored PROMPT with CHOICES."
   (let ((prompt (projectile-prepend-project-name prompt)))
     (cond
-     ((eq projectile-completion-system 'ido)
-      (ido-completing-read prompt choices nil nil initial-input))
-     ((eq projectile-completion-system 'default)
-      (completing-read prompt choices nil nil initial-input))
-     ((eq projectile-completion-system 'helm)
-      (if (fboundp 'helm-comp-read)
-          (helm-comp-read prompt choices
-                          :initial-input initial-input
-                          :candidates-in-buffer t
-                          :must-match 'confirm)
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -1489,36 +1489,45 @@ Never use on many files since it's going to recalculate the
 project-root for every file."
   (expand-file-name name (projectile-project-root)))
 
-(defun projectile-completing-read (prompt choices &optional initial-input)
+(cl-defun projectile-completing-read (prompt choices
```
bbatsov: Do we really need to use `cl-defun` here? I get why you did it, but I generally dislike introducing interfaces that are not common for Elisp.

projectile.el
```diff
@@ -2821,9 +2838,10 @@ With a prefix ARG invokes `projectile-commander' instead of
   (interactive "P")
   (let (projects)
     (if (setq projects (projectile-relevant-known-projects))
-        (projectile-switch-project-by-name
-         (projectile-completing-read "Switch to project: " projects)
-         arg)
+        (projectile-completing-read
+         "Switch to project: " projects
+         :action (lambda (x)
```
bbatsov: I hate param names like `x`.

bbatsov:
Seems to me the second commit should be squashed together with the third.

projectile.el
abo-abo: `cl-defun` allows to have keyword arguments in Elisp, which makes function calls a lot more readable.

Additionally, keyword args allow more flexibility in extending the original interface:

- Adding new keyword args is very easy and breaks nothing.
- Changing keyword args order breaks nothing.
- Removing a keyword arg breaks loudly and is easy to fix.

abo-abo:
Updated.

projectile.el
bbatsov: I'm well-aware of the benefits, I just dislike using features not present in Elisp by default (they tend to be confusing to people not familiar with Common Lisp). Anyways, I can live with this.

PR #1076, opened 2016-10-25
Use helm-build-sync-source

cruegge:
This PR fixes a problem I'm having on Emacs 25 in combination with Helm from a recent Melpa installation.

The way `helm` is called in `projectile-completing-read` is apparently problematic: the `:sources` argument gets passed a plain alist instead of a `helm-source` instance, thereby missing some slots which the constructor usually initializes. This leads to unusual matching behaviour, e.g. the input is not split on spaces properly, as Helm usually does.

The PR adds calling `helm-build-sync-source` instead. Moreover, it adds passing the `:prompt`, `:input` and `:buffer` arguments to `helm`.
---

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1597,12 +1597,15 @@ project-root for every file."
            ((eq projectile-completion-system 'helm)
             (if (fboundp 'helm)
                 (helm :sources
-                      `((name . ,prompt)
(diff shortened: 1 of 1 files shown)


bbatsov:
👍

PR #1092, opened 2016-12-06
Remove package.json from root file list

azer:
`package.json` should not be a part of the root files list because;

* A lot of projects keep JS code in a subfolder. [example 1](https://github.com/olebedev/go-starter-kit) [example 2](https://github.com/olebedev/go-starter-kit)
* A lot of projects have multiple packages. See [learna](https://github.com/lerna/lerna)

I think `package.json` is not needed in the root file list. People could either initialize a version system or add a projectile file.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -300,7 +300,6 @@ If variable `projectile-project-name' is non-nil, this function will not be used
     "requirements.txt"   ; Pip file
     "setup.py"           ; Setuptools file
     "tox.ini"            ; Tox file
-    "package.json"       ; npm package file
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Shouldn't these files also be removed for consistency (since they're typically used in node modules)?

```
     "gulpfile.js"        ; Gulp build file
      "Gruntfile.js"       ; Grunt project file
      "bower.json"         ; Bower project file
```

azer:
@nickmccurdy 👍

nickmccurdy:
#1137 ;)

PR #1094, opened 2016-12-08
[Fix #1019] Generalizing test/implementation lookups

### LINKED ISSUE #1019: Is there a way to jump to a test named the same way but in a different directory?
I've been looking through the code but it seems like the only way to influence the test file checker only considers the file base name. I'm currently in the situation where the project I'm working on arranges tests like this:
- src/some-file.js
- test/some-file.js

Is there any way to teach projectile about this?

jordonbiondo:
This PR puts the existing suffix and prefix based test/implementation lookup feature behind another layer of abstraction. The purpose of this is to allow users greater control over how projectile handles it's knowledge of test files and their relationship to implementation files.

While this PR creates the possibility for more extensive customization it does not include any, and oes not change the current behavior of projectile in any way. It simply allows others, or potentially other packages, to use projectile for projects that don't fit into the prefix/suffix naming convention assumed by projectile. Refer to #1019 for an example of why this is needed. It would be my goal after acceptance of this PR to create a new PR or new package that uses this new customization layer to support `ember-cli` projects in projectile.

This is a work in progress, there are currently no tests, or example implementations. I have opened this PR now to begin the discussion about whether or not this feature can/would be accepted into projectile, or if a different solution to the same problem would be desired.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -460,6 +460,21 @@ Any function that does not take arguments will do."
   :group 'projectile
   :type 'function)
 
+(defcustom projectile-test-file-p-function 'projectile-test-file-suffix-prefix-p
+  "Function to determine if a FILE is a test file."
+  :group 'projectile
+  :type 'function)
+
+(defcustom projectile-find-matching-file-function 'projectile-find-matching-file-default
+  "Function to compute the name of a test matching FILE-PATH."
+  :group 'projectile
+  :type 'function)
+
+(defcustom projectile-find-matching-test-function 'projectile-find-matching-test-default
+  "Function to compute the name of a file matching TEST-FILE-PATH."
+  :group 'projectile
+  :type 'function)
+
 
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm fine with the proposed changes.

bbatsov:
@jordonbiondo Any progress here?

bbatsov:
@jordonbiondo ping :-)

Qquanwei:
@jordonbiondo Have any progress ? Orz

jordonbiondo:
Sorry all, I stopped working on the project that prompted me to do this so I lost track of it, I'll try to rebase this up on the latest code and see if I can take it from there.

dakra:
@jordonbiondo another ping ;)

jordonbiondo:
Yeah to be realistic, I'm never going to pick this back up

Issue #893, opened 2015-10-28
Projectile Modeline lags scrolling on TRAMP buffers


almendez:
When viewing a file remotely through a TRAMP buffer, emacs will lag heavily when scrolling.

I ran the emacs profiler and the problem seems to be caused by (projectile-project-name). Please note that I disabled powerline and other packages and the same behaviour persists.
```
- command-execute                                               15005  59%
 - call-interactively                                           15005  59%
  - mwheel-scroll                                               11896  47%
   - funcall                                                    11883  46%
    - scroll-up                                                 11310  44%
     - eval                                                     11310  44%
      - let*                                                    11310  44%
       - list                                                   11283  44%
        - powerline-minor-modes                                  9935  39%
         - format-mode-line                                      9918  39%
          - eval                                                 9918  39%
           - format                                              9918  39%
            - projectile-project-name                            9917  39%
             - funcall                                           9917  39%
              + #<compiled 0x141a2df>                            9917  39%
         + powerline-raw                                            5   0%
         + mapconcat                                                3   0%
        + powerline-raw                                          1333   5%
        + funcall                                                  11   0%
          powerline-major-mode                                      1   0%
       + concat                                                    14   0%
       + powerline-raw                                              8   0%
   + byte-code                                                      7   0%
   + select-window                                                  3   0%
   + run-with-timer                                                 3   0%
  + byte-code                                                    3103  12%
  + helm-M-x                                                        6   0%
+ ...                                                           10290  40%
+ redisplay_internal (C function)                                   4   0%
+ evil-repeat-post-hook                                             3   0%
  evil-normal-post-command                                          3   0%
+ company-post-command                                              2   0%
+ evil-repeat-pre-hook                                              1   0%
```
The cause of the lag seems to be that (projectile-project-name) is being called repeatedly to update the modeline. I was able to work around it by modifying the projectile mode line to be a static value.
```elisp
(setq projectile-mode-line "XXX")
```
This problem seems related to issue #523 and #657.

Silex:
Yeah, it's because of the ticker which updates whenever the buffer is redisplayed... Adding a simple cache should fix _this_ issue I guess.

almendez:
Here is a better workaround.

```lisp
(setq projectile-mode-line
      '(:eval
        (if (file-remote-p default-directory)
            " Projectilej[*remote*]"
          (format " Projectile[%s]" (projectile-project-name)))))
```

[...]

PR #1096, opened 2016-12-13
Cache the name of the current Projectile project

raxod502:
This improves performance slightly in many cases, and massively in a
few cases. The cache is invalidated whenever `buffer-file-name`
changes, so it should be pretty safe.

From #893:

> I've disabled the dynamic modeline for remote buffers as a temporary
  workaround. I don't use tramp and I don't plan to fix this, so it's
  up to someone else.

How does this solution look?

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -895,17 +895,44 @@ A thin wrapper around `file-truename' that handles nil."
   "Default function used create project name to be displayed based on the value of PROJECT-ROOT."
   (file-name-nondirectory (directory-file-name project-root)))
 
+(defvar-local projectile-cached-project-name nil
+  "Cached name of the current Projectile project. If non-nil, it
+is used as the return value of `projectile-project-name' for
+performance (unless the variable `projectile-project-name' is
+also set). If nil, it is recalculated the next time
+`projectile-project-name' is called.
+
+This variable is reset automatically when Projectile detects that
+the `buffer-file-name' has changed. It can also be reset manually
+by calling `projectile-reset-cached-project-name'.")
+
+(defvar-local projectile-cached-buffer-file-name nil
+  "The last known value of `buffer-file-name' for the current
+buffer. This is used to detect a change in `buffer-file-name',
+which triggers a reset of `projectile-cached-project-name'.")
+
+(defun projectile-reset-cached-project-name ()
+  "Resets the value of `projectile-cached-project-name' to nil,
+so that it is automatically recalculated the next time
+`projectile-project-name' is called."
+  (interactive)
+  (setq projectile-cached-project-name nil))
+
 (defun projectile-project-name ()
   "Return project name."
-  (if projectile-project-name
-      projectile-project-name
-    (let ((project-root
-           (condition-case nil
-               (projectile-project-root)
-             (error nil))))
-      (if project-root
-          (funcall projectile-project-name-function project-root)
-        "-"))))
+  (or projectile-project-name
+      (and (equal projectile-cached-buffer-file-name buffer-file-name)
+           projectile-cached-project-name)
+      (progn
+        (setq projectile-cached-buffer-file-name buffer-file-name)
+        (setq projectile-cached-project-name
+              (let ((project-root
+                     (condition-case nil
+                         (projectile-project-root)
+                       (error nil))))
+                (if project-root
+                    (funcall projectile-project-name-function project-root)
+                  "-"))))))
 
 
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -895,17 +895,44 @@ A thin wrapper around `file-truename' that handles nil."
...
+This variable is reset automatically when Projectile detects that
+the `buffer-file-name' has changed. It can also be reset manually
+by calling `projectile-reset-cached-project-name'.")
+
+(defvar-local projectile-cached-buffer-file-name nil
+  "The last known value of `buffer-file-name' for the current
+buffer. This is used to detect a change in `buffer-file-name',
+which triggers a reset of `projectile-cached-project-name'.")
+
+(defun projectile-reset-cached-project-name ()
```
bbatsov: I don't see any usages of this function.

projectile.el
```diff
@@ -895,17 +895,44 @@ A thin wrapper around `file-truename' that handles nil."
   "Default function used create project name to be displayed based on the value of PROJECT-ROOT."
   (file-name-nondirectory (directory-file-name project-root)))
 
+(defvar-local projectile-cached-project-name nil
+  "Cached name of the current Projectile project. If non-nil, it
```
bbatsov: Run `M-x checkdoc` on your docstrings.

bbatsov:
Another (perhaps simpler) approach would be to just set the `projectile-project-name` for a buffer on `find-file` (using the provided hook) and assume the project name won't change afterwards. This won't work for special project buffers (not backed by files), but your approach won't work for them as well.

projectile.el
raxod502: Indeed, it was intended to be used by the user if something went wrong with the caching mechanism. Should I remove it?

projectile.el
raxod502: Done.

raxod502:
Yes, I noticed that was what you did before [this commit](https://github.com/bbatsov/projectile/commit/5dafafa600348942ac00fbc6cb15c34b93b5890b). I think it is nice to allow for the possibility of `buffer-file-name` being changed unexpectedly. You never know!

I couldn't think of any situation where the buffer could change projects without `buffer-file-name` being changed, though—why do you say this won't work for buffers not backed by files?

bbatsov:
> I couldn't think of any situation where the buffer could change projects without buffer-file-name being changed, though—why do you say this won't work for buffers not backed by files?
>  Show all checks

`buffer-file-name` is `nil` for buffers like `*magit*`, `*dired*` and so on.

raxod502:
Well, sure. But `projectile-cached-project-name` starts off at `nil`, so it'll always be calculated at least once (including for those buffers—note that the cache is invalidated when `projectile-cached-project-name` is `nil` *or* `projectile-cached-buffer-file-name` has changed). The only way we could end up with a problem is if the project name *changes* for those buffers *after* they are created. Is that possible? The only situation I could think of was if the project directory is renamed, but I think that would mess up enough other things to not be worth considering (just a guess).

bbatsov:
Guess it's not a big issue. Btw, won't it be simpler to simply set the project name as a buffer-local variable to every project buffer and dispense with the second variable you've introduced? After all - changing a buffer's name doesn't remove the buffer's local variables.

raxod502:
The point is that if you change `buffer-file-name` to a totally different path, then the project name might need to be recalculated. I can't think of any way to detect that without storing the last known value of `buffer-file-name`.

bbatsov:
> The point is that if you change buffer-file-name to a totally different path, then the project name might need to be recalculated. I can't think of any way to detect that without storing the last known value of buffer-file-name.

I really don't understand what's the point you're trying to make. A buffer-local-variable survives a buffer rename. Some example would be useful - right now it seems to me this patch is more complex than it needs to be.

raxod502:
Naturally; I wouldn't want to introduce unneeded complexity. Suppose we've got a file `foo` in the `bar` Projectile project. The mode line will show `[bar]` for the project. But then suppose we use `C-x C-w` to write the file to a different directory, inside the `baz` Projectile project. If we don't include `projectile-cached-buffer-file-name`, then the mode line will keep showing `[bar]` instead of updating to show `[baz]` for the new file.

bbatsov:
Now, I finally get it. :-) I guess it makes some sense - it's simply not something I'd ever do myself, which it made it hard for me to understand why was the extra check needed.

raxod502:
Thanks!

PR #1100, opened 2016-12-24
[Fix #1067] Added hook that "knows" the current project name.

### LINKED ISSUE #1067: projectile-project-name not returning current project name when called from projectile-after-switch-project-hook
## Expected behavior

Would think that is should return the project name chosen from list supplied by C-c p p.
## Actual behavior

When switching projects with C-c p p projectile-project-name when called from projectile-after-switch-project-hook seems to return the previous project, the first time it's invoked it returns "-".

```
(defun my-projectile-after-hook ()
  (message (concat "project:" (projectile-project-name))))

(add-hook 'projectile-after-switch-project-hook 'my-projectile-after-hook)
```

Just use C-c p p to switch back and forth between projects

```
 projectile-20161005.847
```
### Emacs version

Built from source.

```
GNU Emacs 25.1.2 (x86_64-unknown-linux-gnu, GTK+ Version 3.14.13)
 of 2016-08-06
Copyright (C) 2016 Free Software Foundation, Inc.
```
### Operating system

Centos 7

art-solopov:
Added the `projectile-before-project-action-hook` hook that is triggered right before running `switch-project-action`. This hook is aware of the new `default-directory` and therefore is aware of the new project directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Consider Ensime configuration file as root marker, `.ensime`.
 * [#1057](https://github.com/bbatsov/projectile/issues/1057): Make it possible to disable automatic project tracking via `projectile-track-known-projects-automatically`.
+* [#1067](https://github.com/bbatsov/projectile/issues/1067): Add a hook that runs right before `switch-project-action` within the new project directory.
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


bbatsov:
As mentioned in the ticket you've misunderstood how `default-directory` works. This hook is redundant (and confusing IMO).

PR #1103, opened 2017-01-05
projectile-after-switch-project-hook with arg PROJECT-TO-SWITCH

zenkiezhu:
Here is my situation: I want to set `flycheck-javascript-eslint-executable` to project local `eslint` after switching project, so I need to know the target project dir to check the existence of `PROJECT-TO-SWITCH/node_modules/.bin/eslint`, while I can't work it out by `projectile-after-switch-project-hook` and `default-directory`.

My solution is: add a hook `projectile-after-switch-which-project-hook` which is passed an arg PROJECT-TO-SWITCH.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3021,7 +3021,8 @@ With a prefix ARG invokes `projectile-commander' instead of
       (with-temp-buffer
         (hack-dir-local-variables-non-file-buffer))
       (funcall switch-project-action))
-    (run-hooks 'projectile-after-switch-project-hook)))
+    (run-hooks 'projectile-after-switch-project-hook)
+    (run-hook-with-args 'projectile-after-switch-which-project-hook project-to-switch)))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


bbatsov:
Just check `default-directory` in the existing after hook. No need to complicate this further.

zenkiezhu:
Tried, but the `default-directory` is the dir of the target file, not the target project root

bbatsov:
It used to be until recently due to a bug, but I fixed it a few days ago on `master`. Not sure whether you're using the newest possible projectile.

zenkiezhu:
I'm using the newest from melpa

bbatsov:
Ah, yeah. I meant to say use `(projectile-current-project)`. Sorry about this.

zenkiezhu:
Thanks, using `(projectile-project-root)` works.
Sorry for bothering, I should have checked that.

PR #1104, opened 2017-01-05
Show correct project name when switching projects

raxod502:
See discussion on #1067.


raxod502:
Note, this still doesn't fix the issue where the same project name is shown in the mode line as is shown in the minibuffer, but it does at least make it so the right project name is shown in the minibuffer without messing up the project name in the mode line when you switch back to the previous buffer.

bbatsov:
Add some comment in the source why we're doing this, otherwise someone is bound to guess.

raxod502:
Agreed, and done.

bbatsov:
👍

PR #1112, opened 2017-02-10
Handle prefix argument to projectile-multi-occur as context lines

### LINKED ISSUE #1111: projectile-multi-occur should allow context lines
the `projectile-multi-occur` command should take a prefix arg for context lines, just like the other `occur` commands.

wbolster:
Fixes #1111.


bbatsov:
👍

PR #1122, opened 2017-03-07
Make projectile-project-type safe local variable

madand:
Prevents Emacs from confirming setting of `projectile-project-type` value
in `.dir-locals.el`.


bbatsov:
👍

PR #1128, opened 2017-03-18
Use expand-file-name before using magit-status-internal

marsam:
`magit-status-internal` fails when is called with a relative path.

```
ELISP> (projectile-vc "relative/path/to/project/")
*** Eval error ***  Wrong type argument: stringp, nil
ELISP>
```

`expand-file-name` is used by [`magit-status`](https://github.com/magit/magit/blob/0344413a/lisp/magit-status.el#L177) before calling `magit-status-internal`

Tell me if is necessary to add tests for this change.

Thanks in advance.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2748,7 +2748,7 @@ For hg projects `monky-status' is used if available."
     (cl-case vcs
       (git
        (cond ((fboundp 'magit-status-internal)
-              (magit-status-internal project-root))
+              (magit-status-internal (expand-file-name project-root)))
              ((fboundp 'magit-status)
               (with-no-warnings (magit-status project-root)))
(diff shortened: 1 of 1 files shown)


bbatsov:
Why would anyone pass to this something that's a project dir? Those are never relative paths.

marsam:
I had a eshell alias `alias g projectile-vc $1`, so I got that error with relative paths.  Checking that `magit-status` calls `expand-file-name` before calling `magit-status-internal`, I thought `projectile-vc` might use it as well.

> […] Those are never relative paths.

I'm closing this PR given that you point that `project-root` is never expected to be a relative path.

PR #1137, opened 2017-05-09
Remove more JavaScript files from root file list

nickmccurdy:
Similarly to package.json, these files may exist in subfolders if repositories have JavaScript packages that aren't at the top level (see #1092).


azer:
👍

PR #1138, opened 2017-05-15
Autoload `(projectile-project-root)`

wyuenho:
Apparently some packages like such as some spaceline themes rely on `(projectile-project-root)` to be forward declarable, but if it isn't autoloaded, the packages will result in all kinds of errors. This function is also quite useful for external elisp configs. I suggest you autoload it, since it doesn't look like it's very heavy anyway.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -860,6 +860,7 @@ topmost sequence of matched directories.  Nil otherwise."
                  (not (projectile-file-exists-p (expand-file-name f (projectile-parent dir)))))))))
    (or list projectile-project-root-files-top-down-recurring)))
 
+;;;###autoload
 (defun projectile-project-root ()
   "Retrieves the root directory of a project if available.
 The current directory is assumed to be the project's root otherwise."


bbatsov:
Auto-loading non-interactive functions doesn't make a lot of sense, as if you use them in your config you'll immediately load the entire package anyways. What exactly are you trying to solve?

wyuenho:
I'm trying to solve the use case where it doesn't make sense to require the entire package. This is a useful function for packages that want to get the project root without requiring projectile mode to be activated.

https://github.com/domtronn/spaceline-all-the-icons.el/blob/master/spaceline-all-the-icons-segments.el#L497-L511

I think elpy does something similar as well, they also hide the void function error under an `(ignore-errors)`, which conflates void function error with not finding a project root.

bbatsov:
The problem is autoloading doesn't work the way you think it does - the moment this related function gets called the entire package would be loaded anyways. You can always conditionally require projectile within it or simply add an autoload in the file that refers to the function is question. autoload cookies mostly make sense for interactive commands that might load the package "on demand".

PR #1139, opened 2017-05-16
Improve performance of unignore directive for directories.

smallid:
I have changed the way unignored directories are handled to make use of the optional directory argument provided by 'git ls-files'. This improves performance in the case where there are many files ignored by git in other directories (note that most of the gain is lost if a specific file is unignored since the entire list of ignored files must be retrieved in that case).

No user-visible functionality has been changed, and tests are passing after a tweak to support the new projectile-get-repo-ignored-directory function.

Feedback on this PR would be welcome, as this is my first attempt at contributing :-)

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1138,6 +1138,12 @@ they are excluded from the results of this function."
     (when cmd
       (projectile-files-via-ext-command cmd))))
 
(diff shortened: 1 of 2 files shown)


bbatsov:
The changes look good to me! Thanks!

PR #1140, opened 2017-05-16
Add command project-ag-regexp.

czipperz:
This simply adds a command named `project-ag-regexp`.  This is nice because it allows the user to map this command to a keypress instead of having to remember to C-u to get regular expression behavior.


fommil:
This copies the original code, I'm not a fan. Especially since there are bugs there addressed in open PRs. If it's a command you want, then just create an interactive function calling the main function with the universal arg.

bbatsov:
I agree with @fommil.

bbatsov:
@czipperz Ping :-)

bbatsov:

PR #1145, opened 2017-06-23
projectile-ag now respects ag-ignore-list

### LINKED ISSUE #987: projectile-ag ignores ag-ignore-list when projectile-project-vcs is git

fommil:
fixes #987

This is very hard to test (there are no existing `projectile-ag` tests).

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * Cache the name of the current project for mode-line display of the project name.
 * [#1078](https://github.com/bbatsov/projectile/issues/1078): For projectile-grep/ag use default value like grep/rgrep/ag.
 * Don't treat `package.json` as a project marker.
+* [#987](https://github.com/bbatsov/projectile/issues/987): projectile-ag ignores ag-ignore-list when projectile-project-vcs is git
 
 ### Bugs fixed
 
--- a/projectile.el
+++ b/projectile.el
@@ -2523,13 +2523,15 @@ regular expression."
          current-prefix-arg))
   (if (require 'ag nil 'noerror)
       (let ((ag-command (if arg 'ag-regexp 'ag))
-            (ag-ignore-list (unless (eq (projectile-project-vcs) 'git)
-                              ;; ag supports git ignore files
-                              (cl-union ag-ignore-list
+            (ag-ignore-list (cl-union ag-ignore-list
+                                      (projectile--globally-ignored-file-suffixes-glob)
+                                      ;; ag supports git ignore files directly
+                                      (unless (eq (projectile-project-vcs) 'git)
                                         (append
-                                         (projectile-ignored-files-rel) (projectile-ignored-directories-rel)
-                                         (projectile--globally-ignored-file-suffixes-glob)
-                                         grep-find-ignored-files grep-find-ignored-directories))))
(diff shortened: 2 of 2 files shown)


fommil:
is there anything I need to do to push this along?

bbatsov:
> is there anything I need to do to push this along?

I've just been extremely busy lately. Sorry about this.

fommil:
Cool thanks! Beauty of lisp is I don't need you to release, I just add to my config :-)

vincekd:
After this update I get the error `cl-member: Bad keyword argument nil` if `ag-ignore-list` is set to non-nil.

fommil:
probably because your `ag-ignore-list` is now being respected and you have a bug in it, or there is a bug in some other code that uses it.

I kind of think `ag-ignore-list` shouldn't exist... `ag` already uses `~/.agignore`.

vincekd:
I don't think that's it. Calling `ag-files` works as expected, it's just when I do `projectile-ag` that it throws the error. I believe the issue is `cl-union` can only take two lists. Can you replicate?

`ag-ignore-list` allows you to set project-specific ignored dirs though.

fommil:
You can put agignore files in projects too. Hmm, if cl-union only takes two lists then that's a bug and a PR to fix would be great. Ugly though.

Issue #1003, opened 2016-04-24
projectile-project-root is called often and is slow


aaronjensen:
Plenty of things call `projectile-project-root` against lists of files such as `projectile-ignored-directories`, `projectile-unignored-files`, etc. Because it is called so often, it is sensitive to performance issues. 

So far, I've found at least two performance issues. The first is from:

https://github.com/bbatsov/projectile/pull/929

if the `projectile-project-root` var is not set, `file-truename` (which is slow) will be called on `dir` every time `projectile-project-root` is invoked. This can be fixed by implementing the suggestion here: https://github.com/bbatsov/projectile/pull/929#discussion_r47745040

The second is that `file-exists-p` is slow as well and is called even if the cache is hit. One solution here would be to simply omit this check. Probably an ideal fix would be something like checking the project root once at the beginning of an operation instead of once (or more) for each file in a project.

ibrahima:
Oh man, didn't realize I introduced a performance regression :( What functions should I test to see the regression? I didn't personally experience an issue myself. But it makes sense to remove the wrapper function if that's a performance issue. I just didn't like the idea of introducing a special case for this feature when there was a projectile-project-root-files-functions list to add to.

Silex:
Most of the performance issues are usually TRAMP related... see https://github.com/bbatsov/projectile/issues/523 for example.

aaronjensen:
@ibrahima It's a somewhat subtle pause coming from `projectile-current-project-files`. Here's a profile from me running it a few times. This comes from a fairly small directory with a `.projectile` in it.

```
                - projectile-current-project-files               1772  41%
                 - -mapcat                                       1711  40%
                  - mapcar                                       1711  40%
                   - #<compiled 0x444f465b>                      1711  40%
                    - projectile-dir-files                       1711  40%
                     - projectile-adjust-files                   1456  34%
                      - projectile-add-unignored                  923  21%
                       - projectile-remove-ignored                456  10%
                        - projectile-ignored-directories-rel                294   6%
                         - projectile-ignored-directories                279   6%
                          - -map                                  173   4%
                           - mapcar                               173   4%
                            - projectile-expand-root                173   4%
                             - projectile-project-root                173   4%
                              + file-truename                     146   3%
                              + file-exists-p                      18   0%
                          - projectile-unignored-directories                 61   1%
                           - projectile-project-unignored-directories                 61   1%
                            - projectile-project-unignored                 61   1%
[...]

PR #1149, opened 2017-07-19
Cache projectile-project-root for performance

alexmurray:
Cache the result of (projectile-project-root) as a buffer local variable to
increase performance - this is implemented the same as was done
for (projectile-project-name) in PR 1906.

Should help resolve issue #1003

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -860,29 +860,55 @@ topmost sequence of matched directories.  Nil otherwise."
                  (not (projectile-file-exists-p (expand-file-name f (projectile-parent dir)))))))))
    (or list projectile-project-root-files-top-down-recurring)))
 
+(defvar-local projectile-cached-project-root nil
+  "Cached root of the current Projectile project. If non-nil, it
+is used as the return value of `projectile-project-root' for
+performance (unless the variable `projectile-project-root' is
+also set). If nil, it is recalculated the next time
+`projectile-project-root' is called.
+
+This variable is reset automatically when Projectile detects that
+the `buffer-file-name' has changed. It can also be reset manually
+by calling `projectile-reset-cached-project-root'.")
+
+(defvar-local projectile-cached-buffer-file-name nil
+  "The last known value of `buffer-file-name' for the current
+buffer. This is used to detect a change in `buffer-file-name',
+which triggers a reset of `projectile-cached-project-root' and
+`projectile-cached-project-name'.")
+
 (defun projectile-project-root ()
   "Retrieves the root directory of a project if available.
 The current directory is assumed to be the project's root otherwise."
   ;; The `is-local' and `is-connected' variables are used to fix the behavior where Emacs hangs
   ;; because of Projectile when you open a file over TRAMP. It basically prevents Projectile from trying
   ;; to find information about files for which it's not possible to get that information right now.
-  (let* ((dir default-directory)
-         (is-local (not (file-remote-p dir)))      ;; `true' if the file is local
-         (is-connected (file-remote-p dir nil t))) ;; `true' if the file is remote AND we are connected to the remote
-    (or (when (or is-local is-connected)
-          (cl-some
-           (lambda (func)
-             (let* ((cache-key (format "%s-%s" func dir))
-                    (cache-value (gethash cache-key projectile-project-root-cache)))
-               (if (and cache-value (file-exists-p cache-value))
-                   cache-value
-                 (let ((value (funcall func (file-truename dir))))
-                   (puthash cache-key value projectile-project-root-cache)
-                   value))))
-           projectile-project-root-files-functions))
-        (if projectile-require-project-root
-            (error "You're not in a project")
-          default-directory))))
+  (or (cl-subst nil 'none
+                (or (and (equal projectile-cached-buffer-file-name buffer-file-name)
+                         projectile-cached-project-root)
+                    (progn
+                      (setq projectile-cached-buffer-file-name buffer-file-name)
+                      (setq projectile-cached-project-root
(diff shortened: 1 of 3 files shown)


fommil:
I'd like to give this a try because I've been hit by perf in the past.

projectile.el
```diff
@@ -860,29 +860,55 @@ topmost sequence of matched directories.  Nil otherwise."
...
-        (if projectile-require-project-root
-            (error "You're not in a project")
-          default-directory))))
+  (or (cl-subst nil 'none
+                (or (and (equal projectile-cached-buffer-file-name buffer-file-name)
+                         projectile-cached-project-root)
+                    (progn
+                      (setq projectile-cached-buffer-file-name buffer-file-name)
+                      (setq projectile-cached-project-root
+                            (or (let* ((dir default-directory)
```
bbatsov: You forgot the move the comment explaining this code here.

projectile.el
```diff
@@ -911,10 +937,13 @@ This variable is reset automatically when Projectile detects that
 the `buffer-file-name' has changed. It can also be reset manually
 by calling `projectile-reset-cached-project-name'.")
 
-(defvar-local projectile-cached-buffer-file-name nil
```
bbatsov: Why did you delete this?

projectile.el
```diff
@@ -860,29 +860,55 @@ topmost sequence of matched directories.  Nil otherwise."
...
-               (if (and cache-value (file-exists-p cache-value))
-                   cache-value
-                 (let ((value (funcall func (file-truename dir))))
-                   (puthash cache-key value projectile-project-root-cache)
-                   value))))
-           projectile-project-root-files-functions))
-        (if projectile-require-project-root
-            (error "You're not in a project")
-          default-directory))))
+  (or (cl-subst nil 'none
```
bbatsov: I don't understand what exactly is this supposed to do. Seems like some unnecessary complication of the code.

alexmurray: To handle the case where there is no project root (and (projectile-project-root) returns nil) we can't just store nil as the cached value - as then we would end up reevaluating it every time and never using the cache. So instead we store the symbol 'none there instead - however we don't want to return 'none from the function since this invalidates the expected contract (return nil if not in a project) so we use cl-subst to replace any 'none value with nil.

projectile.el
alexmurray: Ah thanks.

projectile.el
alexmurray: I just moved it up earlier in the file so it is declared before we now first reference it.

projectile.el
bbatsov: Can't you simply use `'none` as the default value for the buffer-local-variable?

alexmurray: But then we will never evaluate it - the code is:

```
(or (and (equal projectile-cached-buffer-file-name buffer-file-name)
               projectile-cached-project-root)
```

So a non-nil value means the buffer-local cached value will be used - so it needs to be nil initially, then it gets evaluated once, and either set to a path or `'none` and then it can be used as a cached value without reevaluation in the future.

fommil: Isn't it easier if nil just cache misses? So what, projectile is slightly slower on directories that are not source managed or have a manual .projectile :-/

There is an elisp package for caching btw, if you'd like to reuse it.

alexmurray: I disagree - I often use emacs on directories which are not part of a project and having EVERYTHING slow down (say when using [spaceline-all-the-icons](https://github.com/alexmurray/spaceline-all-the-icons.el) which calls `(projectile-project-root)` on every redisplay) is very annoying. Why go to the effort of caching in a project and not also add support for caching outside of a project - also since when outside a project to determine this each time is expensive since we have to exhaust every possible project test each time - it MORE of a win to cache in this case than in the case of being in a project.

Which elisp package are you referring to?

bbatsov: Just explain in a comment what this code is doing for future reference and I'll merge it. No need for more 3rd party deps.

alexmurray: How about c473d17 ? Do you want me to squash these all into a single commit?

projectile.el
bbatsov: No need for this. I can squash from the web ui.

bbatsov:
Thanks!

projectile.el
```diff
@@ -911,10 +945,13 @@ This variable is reset automatically when Projectile detects that
 the `buffer-file-name' has changed. It can also be reset manually
 by calling `projectile-reset-cached-project-name'.")
 
-(defvar-local projectile-cached-buffer-file-name nil
-  "The last known value of `buffer-file-name' for the current
-buffer. This is used to detect a change in `buffer-file-name',
-which triggers a reset of `projectile-cached-project-name'.")
+(defun projectile-reset-cached-project-root ()
```
aaronjensen: Would it make sense to call this during `projectile-invalidate-cache`?

Silex:
I wonder if it still makes sense to cache `projectile-cached-project-name`

fommil:
that lib: https://github.com/skeeto/emacs-memoize

fommil:
(in hindsight, I agree it makes sense to cache non-projects. I have worked in environments with network drives and ... short of having an extensive exclusion list for projectile... a cache for buffers in those folders would have surely helped)

JeanMax:
Hey!

This merge broke (projectile-all-project-files) for me.
It now only returns the files of the current project... so (projectile-find-file-in-known-projects) acts as (projectile-find-file); 'projectile-known-projects is set correctly.

I could reproduce the behaviour with 'emacs -q' and 2 random projects, but it seems that it doesn't affect everybody.
linux / emacs 25.2.1

Issue #1150, opened 2017-07-24
projectile-switch-project breaks after recent caching commit


kaushalmodi:
## Expected behavior

`projectile-switch-project` should list files from the project to be switched **to** (default projectile-find-file action).

## Actual behavior

`projectile-switch-project` instead behaves just like `projectile-find-file`.. shows files from the current project.

## Steps to reproduce the problem

1. Be on commit https://github.com/bbatsov/projectile/commit/7951b17e6d4adef8fcb4e15ef6c466ef53b59ccf
2. Open a file in project A.
3. `M-x projectile-switch-project` -> Select a project B -> Files are still listed from project A!

---

**Temp fix**: I have confirmed that reverting to the one-prior commit https://github.com/bbatsov/projectile/commit/5dea114a59e0648561c00aeb2ceaba876bf04cae fixes this issue.

### Projectile version information

- Latest projectile build from Melpa (uses commit https://github.com/bbatsov/projectile/commit/7951b17e6d4adef8fcb4e15ef6c466ef53b59ccf)

> *Include here the version string displayed by `M-x
projectile-version`. Here's an example:*

`projectile-version` seems to have a bug. I was on the one-prior commit to test that the last commit actually introduced this bug. And `M-x projectile-version` still printed:

> Projectile 0.15.0snapshot (package: 20170724.442)

even when I haven't loaded the elpa version.. I have evaluated the one-prior commit:

![image](https://user-images.githubusercontent.com/3578197/28529672-2992fa22-705f-11e7-8665-51394a6734d6.png)

### Emacs version

Latest emacs build from master (26.0.50)

### Operating system

RHEL 6.6

kaushalmodi:
@alexmurray

manuel-uberti:
I can confirm this.
```
Debugger entered--Lisp error: (error "You’re not in a project")
  signal(error ("You’re not in a project"))
  error("You're not in a project")
  projectile-project-root()
  projectile-dirconfig-file()
  projectile-maybe-invalidate-cache(nil)
  projectile-find-file()
  projectile-switch-project-by-name("~/.emacs.d/" nil)
  #f(compiled-function (dir) #<bytecode>)("~/.emacs.d/")
  ivy-call()
[...]

PR #1151, opened 2017-07-25
Fix GH issue #1150

alexmurray:
When switching projects need to ignore the cached value of the project root otherwise the old project is shown, not the new project.

I wonder if a similar approach could be taken for projectile-project-name here as well but I left it so as to keep the changeset as minimal as possible.


bbatsov:
> I wonder if a similar approach could be taken for projectile-project-name here as well but I left it so as to keep the changeset as minimal as possible.

I wonder if we need this caching at all now. The project name calculation was slow mostly due to the slow computation of the project root anyways. Maybe we can drop this caching completely with minimum performance impact.

Silex:
I second @bbatsov

PR #1159, opened 2017-08-09
Add `nix' project type

matthewbauer:
I've been using this personally for a while but I think it makes sense to go in Projectile. This just registers "nix" projects as anything containing a "default.nix" file. "nix-build" actually both compiles and tests (if doCheck=true), should it be set for both?

-----------------

Thanks!


bbatsov:
> "nix-build" actually both compiles and tests (if doCheck=true), should it be set for both?

I'd say - yes.

You should also mention this in the changelog.

matthewbauer:
Ok, looks like it's mergeable now. /cc @bbatsov

PR #1160, opened 2017-08-10
Fix `(projectile-all-project-files)`

alexmurray:
Need to ignore current cached project when iterating across all projects

See https://github.com/bbatsov/projectile/pull/1149#issuecomment-321337673

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3215,7 +3215,8 @@ This command will first prompt for the directory the file is in."
   (cl-mapcan
    (lambda (project)
      (when (file-exists-p project)
-       (let ((default-directory project))
+       (let ((default-directory project)
+             (projectile-cached-project-root nil))
          (mapcar (lambda (file)
                    (expand-file-name file project))
                  (projectile-current-project-files)))))


projectile.el
```diff
@@ -3215,7 +3215,8 @@ This command will first prompt for the directory the file is in."
   (cl-mapcan
    (lambda (project)
      (when (file-exists-p project)
-       (let ((default-directory project))
+       (let ((default-directory project)
+             (projectile-cached-project-root nil))
```
bbatsov: Don't we also have a version of `projectile-project-files` that takes a project as a parameter? Seems it'd be better to use it instead, as opposed to meddling with dynamic bindings and using just `projectile- current-project-files`?

alexmurray: I can't see one - I was just going for the lightest touch to fix this...

bbatsov: I'll think about this fix. It seems to me that we're not fixing the real problem - if such a function doesn't exist it should be trivial to add it and simply implement `projectile-current-project-files` in terms of it. Basically the implicit project-root would be become some parameter.

ericdanan: Perhaps this approach could also be used to switch project? That is, functions used to as switch project actions (`projectile-find-file` ...) could take the project root as optional parameter and `projectile-switch-project-by-name` could then simply do something like this:

```
(run-hooks 'projectile-before-switch-project-hook)
(funcall action project-to-switch)
(run-hooks 'projectile-after-switch-project-hook)
```

without needing to let-bind `default-directory`? I'm not sure how this would affect the caching mechanism for project roots and names or if it may not be suitable for all potential actions.

Speaking about `projectile-switch-project-by-name`, I think the `with-temp-buffer` call inside it is useless now that that the action is no longer called from the temp buffer (commit 72a8be5 ). Unless I'm mistaken, it just creates a temp buffer, loads the project-to-siwtch's dir-local variables into this buffer and then immediately kills it.

And speaking about this commit, it reintroduced issue #1005 (for me at least) that had been solved by calling the action from the temp buffer (but that had created other issues). Perhaps this issue would be solved again if `projectile-switch-project-by-name`no longer let-binds `default-directory`?

bbatsov: Yeah, definitely. I think the approach with dynamic vars was an epic mistake on my part and I regret taking this shortcut in the early days of the project. I would have changed this a while ago if I had any time for OSS work.

bbatsov:
Seems you're not interested in taking this further, so I'll just merge this fix as is. Maybe we'll revisit this horrible design decision down the road.

alexmurray:
@bbatsov - sorry I haven't had time to look at this in more detail - will try and post a PR if I get some time in the future - what did you have in mind? Something like what @ericdanan posted earlier https://github.com/bbatsov/projectile/pull/1160#discussion_r132819888 ?

ericdanan:
I am also happy to help if I can. But I guess getting rid of the let-bindings for `default-directory` is quite a lot of change so I'm not sure what is the best way to proceed. What do you think?

PR #1164, opened 2017-08-13
Implement 'projectile-find-all-references'

myTerminal:
## Summary

The change is to implement the function `projectile-find-all-references`, that prompts the user for a text to be searched for all references in files within the current project.

## Motivation

The available functions `projectile-ag` and `projectile-grep` are good at what they do already, but have their own limitations: first relies on an external tool and the other though it works on Unix based systems, gives a hard time to get it to work on Windows workstations. I spent a lot of time to make it work in my Windows workstation at office but failed as there are apparently different versions of those dependencies out of which only some work on Windows.

The function utilizes some of the architecture that is already available and adds some code to achieve the functionality without relying on any external tools.

## Shortcomings

- The code may not look clean as I managed to implement it with my shallow knowledge of Elisp
- The implementation assumes that a line within a file will not hold more than one occurrence of the text being searched. In case it does, only the first occurrence is listed, the rest are all ignored.

---
This change is [Reviewable](https://reviewable.io/reviews/bbatsov/projectile/1164)

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2598,6 +2598,46 @@ regular expression."
         (funcall ag-command search-term (projectile-project-root)))
     (error "Package 'ag' is not available")))
 
+;;;###autoload
+(defun projectile-find-all-occurrences (&optional arg)
+    "List all occurrences of the text being searched using completion.
+With a prefix argument ARG prompts you for a directory on which to run search in."
+    (interactive "P")
+    (let* ((text-to-search (read-string
+                            (projectile-prepend-project-name "Find all: ")
+                            (projectile-symbol-or-selection-at-point)))
+           (directory (if arg
+                          (file-name-as-directory
+                           (read-directory-name "Find in directory: "))
+                        (projectile-project-root)))
+           (file-to-search (projectile-files-with-string text-to-search directory)))
+      (projectile-completing-read
+       "All occurrences: "
+       (reduce #'append
+               (mapcar (lambda (file)
+                         (with-temp-buffer
+                           (insert-file-contents file)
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -2598,6 +2598,46 @@ regular expression."
...
+                            (projectile-symbol-or-selection-at-point)))
+           (directory (if arg
+                          (file-name-as-directory
+                           (read-directory-name "Find in directory: "))
+                        (projectile-project-root)))
+           (file-to-search (projectile-files-with-string text-to-search directory)))
+      (projectile-completing-read
+       "All occurrences: "
+       (reduce #'append
+               (mapcar (lambda (file)
```
bbatsov: Probably you can extract this and the action into some private functions which you can actually test and would be easier to understand.

bbatsov:
There are a few problems with this:

* for bigger projects it would certainly freeze your Emacs for a while, so I'd add some configurable file threshold, after which (e.g. project has over 100 files) you get a warning confirming you really want to use this command. 
* on unix machines I'd also add some warning to users to try something else instead
* this should probably be added the the menu, manual etc

myTerminal:
Hi @bbatsov,

Thanks for your comments, I agree.
Even though the code has changed since I initiated this pull-request, I think my code is not yet ready to be added in here. I have a feeling it would 'pollute' projectile. It definitely needs more work.

I have created a separate [project](https://github.com/myTerminal/projectile-extras) for the time-being for at least myself to use in my [dotfiles](https://github.com/myTerminal/dotfiles).

I should say you've done a great job with projectile. I use it every single day at home/work.

PR #1166, opened 2017-08-21
Add -other-frame versions of commands that had -other-window versions

mattiasb:
Some commands in projectile comes with an `-other-window` version with keybindings under the `4` prefix.
Sometimes I want to open a `dired` instance or a file in another frame and wonder why there is not `5`-prefix for frames. Well now there is, if this PR is merged. :)

-----------------

- [ x ] The commits are consistent with our [contribution guidelines](../CONTRIBUTING.md)
- [ - ] You've added tests (if possible) to cover your change(s)
  - I don't believe it applies here. No new units to tests really.
- [ x ] All tests are passing (`make test`)
- [ x ] The new code is not generating bytecode or `M-x checkdoc` warnings
  - No, but the old code sure is. ;)
- [ x ] You've updated the changelog (if adding/changing user-visible functionality)
- [ x ] You've updated the readme (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
+* [#1166](https://github.com/bbatsov/projectile/pull/1166): Add `-other-frame` versions of commands that had `-other-window` versions.
 * Consider Ensime configuration file as root marker, `.ensime`.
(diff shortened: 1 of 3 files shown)


bbatsov:
Nice improvements! Thanks! 👍

mattiasb:
Woop! :)

Issue #1162, opened 2017-08-11
delete-dups call in projectile-ag causes "Attempt to modify read-only object" error


nwidger:
I get the following error when I try to run `projectile-ag`: 

```
Debugger entered--Lisp error: (error "Attempt to modify read-only object" ("SCCS" "RCS" "CVS" "MCVS" ".src" ".svn" ".git" ".hg" ".bzr" "_MTN" "_darcs" "{arch}"))
  delete-dups(("*.svn-base" "*.o" "*.pyc" ".gitignore" "TAGS" ".svn/" ".idea/" ".eunit/" ".git/" ".hg/" ".fslckout/" ".bzr/" "_darcs/" ".tox/" "Godeps/" ".#*" "*~" "*.bin" "*.lbin" "*.so" "*.a" "*.ln" "*.blg" "*.bbl" "*.elc" "*.lof" "*.glo" "*.idx" "*.lot" "*.fmt" "*.tfm" "*.class" "*.fas" "*.lib" "*.mem" "*.x86f" "*.sparcf" "*.dfsl" "*.pfsl" "*.d64fsl" "*.p64fsl" "*.lx64fsl" "*.lx32fsl" "*.dx64fsl" "*.dx32fsl" "*.fx64fsl" "*.fx32fsl" "*.sx64fsl" "*.sx32fsl" "*.wx64fsl" ...))
  projectile-ag("" nil)
  funcall-interactively(projectile-ag "" nil)
  call-interactively(projectile-ag nil nil)
  command-execute(projectile-ag)
```

If I remove the `delete-dups` call from `projectile-ag`, the error goes away.  I believe commit https://github.com/bbatsov/projectile/commit/916c87f2ffce267b770a3121aa261a4d083a401c is to be the culprit.

I am running Emacs 25.2.1 on CentOS 7, projectile 20170727.2351 and ag 20170712.1549 (both from MELPA).  Please let me know if I can provide any more information.

czw:
Replicated with Projectile 20170827.2053 as well. Environment: Emacs 25.2.1, macOS 10.3.3.

danielbarbarito:
Replicated with Projectile 0.15.0-cvs. This only happens when I search in non-git projects. As soon as I add a `.git` folder (even if it is completely empty) everything works.

nwidger:
[...]

PR #1172, opened 2017-09-03
Remove delete-dups call in projectile-ag

danielbarbarito:
Removed the `delete-dups` call from the projectile-ag function which caused a "Attempt to modify read-only object" error. This PR solves [this issue](https://github.com/bbatsov/projectile/issues/1162).

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4,6 +4,7 @@
 
 ;; Author: Bozhidar Batsov <bozhidar@batsov.com>
 ;; URL: https://github.com/bbatsov/projectile
+;; Package-Version: 20170827.2053
 ;; Keywords: project, convenience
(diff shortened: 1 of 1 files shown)


runekaagaard:
Would be cool to have this merged.

projectile.el
```diff
@@ -4,6 +4,7 @@
 
 ;; Author: Bozhidar Batsov <bozhidar@batsov.com>
 ;; URL: https://github.com/bbatsov/projectile
+;; Package-Version: 20170827.2053
```
bbatsov: This should be removed.

bbatsov:
Seems the approach in #1173 is better, so I'll close this one.

runekaagaard:
OK. Thx for reply.

PR #1173, opened 2017-09-03
Copy grep-find-ignored-directories in projectile-ag

### LINKED ISSUE #1162: delete-dups call in projectile-ag causes "Attempt to modify read-only object" error

nwidger:
The call to `delete-dups` in `projectile-ag` throws an error if it tries to modify the read-only object `grep-find-ignored-directories`.  This only happens on non-Git projects, since `projectile-ag` only references `grep-find-ignored-directories` if `projectile-project-vcs` does not return `git`.  This PR wraps the reference to `grep-find-ignored-directories` with a call to `copy-sequence`, which prevents `delete-dups` from modifying `grep-find-ignored-directories`.

I'm not sure if this is the best way to solve the problem.  Please let me know if there is a better solution!

Fixes #1162

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2711,15 +2711,16 @@ regular expression."
       (let ((ag-command (if arg 'ag-regexp 'ag))
             (ag-ignore-list (delq nil
                                   (delete-dups
-                                   (append
-                                    ag-ignore-list
-                                    (projectile--globally-ignored-file-suffixes-glob)
-                                    ;; ag supports git ignore files directly
-                                    (unless (eq (projectile-project-vcs) 'git)
-                                      (append (projectile-ignored-files-rel)
-                                              (projectile-ignored-directories-rel)
-                                              grep-find-ignored-files
-                                              grep-find-ignored-directories))))))
+                                   (copy-sequence
+                                    (append
+                                     ag-ignore-list
+                                     (projectile--globally-ignored-file-suffixes-glob)
+                                     ;; ag supports git ignore files directly
+                                     (unless (eq (projectile-project-vcs) 'git)
+                                       (append (projectile-ignored-files-rel)
+                                               (projectile-ignored-directories-rel)
+                                               grep-find-ignored-files
+                                               grep-find-ignored-directories)))))))
             ;; reset the prefix arg, otherwise it will affect the ag-command
             (current-prefix-arg nil))
         (funcall ag-command search-term (projectile-project-root)))
--- a/projectile.el
+++ b/projectile.el
@@ -2711,16 +2711,15 @@ regular expression."
       (let ((ag-command (if arg 'ag-regexp 'ag))
             (ag-ignore-list (delq nil
                                   (delete-dups
-                                   (copy-sequence
-                                    (append
-                                     ag-ignore-list
-                                     (projectile--globally-ignored-file-suffixes-glob)
-                                     ;; ag supports git ignore files directly
-                                     (unless (eq (projectile-project-vcs) 'git)
-                                       (append (projectile-ignored-files-rel)
-                                               (projectile-ignored-directories-rel)
-                                               grep-find-ignored-files
-                                               grep-find-ignored-directories)))))))
+                                   (append
+                                    ag-ignore-list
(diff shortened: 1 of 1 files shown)


bbatsov:
Btw, won't it be much easier to use `cl-union` instead of `append`?

nwidger:
@bbatsov Do you mean replacing `delete-dups` with a call to `cl-union` and getting rid of the `copy-sequence` (since `cl-union` is apparently non-destructive)?

bbatsov:
> @bbatsov Do you mean replacing delete-dups with a call to cl-union and getting rid of the copy-sequence (since cl-union is apparently non-destructive)?

I meant replacing the `append` which result `delete-dups` is modifying. `cl-union` would merge the lists removing all duplicate values.

nwidger:
@bbatsov But `cl-union` already removes duplicates, so wouldn't passing the result from `cl-union` to `delete-dups` be redundant?

```
(cl-union '(one two three) '(one two three four)) -> (one two three four)
```

nwidger:
@bbatsov Any thoughts on this PR?  Issue #1162 is still an issue for me.

bbatsov:
> @bbatsov But cl-union already removes duplicates, so wouldn't passing the result from cl-union to delete-dups be redundant?

Exactly. This was my whole point - if you use `union` you solve the original problem and you don't need `delete-dups`.

projectile.el
```diff
@@ -2719,7 +2719,7 @@ regular expression."
                                       (append (projectile-ignored-files-rel)
```
bbatsov: This should be `cl-union`.

bbatsov: And the preceding `append` as well.

fommil:
I find that a better alternative to this is to let ag see `ag-ignore-list` even if using a VCS... that way you can put whatever you want in there AND the output is not polluted by the grep noise.

I'll raise a PR with this simple approach.

fommil:
oh, it seems others have been hacking in this area already.

bbatsov:
@fommil Where?

fommil:
projectile-ag, is different between stable and master. I was just going to make it use ag-ignore-list even if using git, but the changes are a lot more substantial that that.

ag-ignore-list should be local variable, I sent a request upstream. That way dir-locals can have project specific "don't search the test data" configuration.

bbatsov:
Got it. Thanks!

PR #1175, opened 2017-09-07
Improved support for out-of-tree compilation

### LINKED ISSUE #1169: projectile-compile-project does not prompt for compilation command
In a git project with CMake, when I type `C-c p c`, projectile executes `cmake -E chdir build cmake .. && cmake --build build --target` without prompting me for a compilation command.  Same with `C-u C-c p c`.  `compilation-read-command` is `t`.

I would expect to be prompted for a compilation command.  That was the behavior of the previous version of projectile.

```
Projectile version: 20170827.2053
```

### Emacs version

GNU Emacs 25.1

### Operating system

Ubuntu 17.04

mattiasb:
Make it possible to enforce out-of-tree compilation for build systems that can't build in-tree (eg. meson) while also making out-of-tree or in-tree compilation a preference for the user to set for build-systems that can do both (CMake, autotools etc).

These build systems usually needs to run a configure step before building so this PR adds a new command for that too.

As part of this I factored out the common body of `projectile-{compile,test,run}-project`, but these commands were behaving differently in subtle ways so this PR also makes them work the same. 

Specifically: 
 - All these commands now run inside `(projectile-compilation-dir)`, not
   just the compile command.
 - All these commands cache the compilation command per directory
   instead of per project (again not just the compile command).
 - The extra `dir` parameter to the compile command was dropped.

There are some subtle changes in some of the commits so please
review carefully. I hope I'm not breaking anything.

Also make the CMake- and meson project types use this new functionality.

Fixes #1169 

-----------------

- [ ✓ ] The commits are consistent with our [contribution guidelines](../CONTRIBUTING.md)
- [ ✓ ] You've added tests (if possible) to cover your change(s)
- [ ✓ ] All tests are passing (`make test`)
- [ ✓ ] The new code is not generating bytecode or `M-x checkdoc` warnings
- [ ✓ ] You've updated the changelog (if adding/changing user-visible functionality)
- [ ✓ ] You've updated the readme (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2233,16 +2233,21 @@ With a prefix ARG invalidates the cache first."
   "A hash table holding all project types that are known to Projectile.")
 
 (cl-defun projectile-register-project-type
-    (project-type marker-files &key compile test run test-suffix test-prefix)
+    (project-type marker-files &key configure compile test run test-suffix test-prefix)
   "Register a project type with projectile.
 
 A project type is defined by PROJECT-TYPE, a set of MARKER-FILES,
-and optional keyword arguments COMPILE which specifies a command
-that builds the project, TEST which specified a command that
-tests the project, RUN which specifies a command that runs the
-project, TEST-SUFFIX which specifies test file suffix, and
+and optional keyword arguments:
+CONFIGURE which specifies a command that configures the project
+          `%s' in the command will be substituted with (projectile-project-root)
+          before the command is run,
+COMPILE which specifies a command that builds the project,
+TEST which specified a command that tests the project,
+RUN which specifies a command that runs the project,
+TEST-SUFFIX which specifies test file suffix, and
 TEST-PREFIX which specifies test file prefix."
   (let ((project-plist (list 'marker-files marker-files
+                             'configure-command configure
                              'compile-command compile
                              'test-command test
                              'run-command run)))
@@ -3139,6 +3144,10 @@ For hg projects `monky-status' is used if available."
   "Serializes the memory cache to the hard drive."
   (projectile-serialize projectile-projects-cache projectile-cache-file))
 
+(defvar projectile-configure-cmd-map
+  (make-hash-table :test 'equal)
+  "A mapping between projects and the last configure command used on them.")
+
 (defvar projectile-compilation-cmd-map
   (make-hash-table :test 'equal)
   "A mapping between projects and the last compilation command used on them.")
@@ -3150,6 +3159,11 @@ For hg projects `monky-status' is used if available."
 (defvar projectile-run-cmd-map
   (make-hash-table :test 'equal)
   "A mapping between projects and the last run command used on them.")
+
+(defvar projectile-project-configure-cmd nil
+  "The command to use with `projectile-configure-project'.
+It takes precedence over the default command for the project type when set.
+Should be set via .dir-locals.el.")
 
 (defvar projectile-project-compilation-cmd nil
   "The command to use with `projectile-compile-project'.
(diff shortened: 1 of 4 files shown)


mattiasb:
Tell me if some of the commits should be split up or if this PR should be split up. It's easy to do.

mattiasb:
Meh. These tests doesn't fail locally (I even did a `git rebase -x "git clean -fdx && make" master`). Probably some Emacs 25-specific stuff. :(

mattiasb:
And there! Apparently pcase needs backticks in Emacs 24.

mattiasb:
Re-triggering the build (since it seems to have failed on some weird thing).

@bbatsov I understand that you might be too busy for a full review right now, but are there any major / obvious issues with this PR that I could work on while waiting?

projectile.el
```diff
@@ -213,11 +213,21 @@ Otherwise consider the current directory the project root."
   :group 'projectile
   :type 'string)
 
-(defcustom projectile-build-dir "build"
-  "The directory Projectile will use for build systems that build out of tree."
+(defcustom projectile-compilation-dir "build"
```
bbatsov: Shouldn't this be something project-type specific?

projectile.el
```diff
@@ -213,11 +213,21 @@ Otherwise consider the current directory the project root."
...
+  "The directory Projectile will use when compiling out of tree.
+
+The directory path is relative to the project root.
+
+Overridable per project by setting `projectile-project-compilation-dir' in
+.dir-locals"
   :group 'projectile
   :type 'string)
 
+(defcustom projectile-prefer-out-of-tree-compilation t
```
bbatsov: New defcustoms should all have a `:package-version` property. 

I don't like a lot the name of this `defcustom` and I wonder if it should be something like `projectile-compilation-style` with possible options `'in-project-tree` and `'outside-of-project-tree`. Not a big deal though.

bbatsov:
Your PR seems to be of great quality overall, I've added just a couple of minor comments.

bbatsov:
I'd also suggest adding a section on this to the manual.

bbatsov:
@mattiasb ping :-)

mattiasb:
@bbatsov Hi, what's up? ;) 

Yeah I know, I've been procrastinating this a bit but life's been more than busy. I'll try to get to this on Thursday or Sunday or so.

damienrg:
@mattiasb  Can I do something to help merging this PR?

projectile.el
mattiasb: This only applies sometimes though. For CMake it's possible to both build in- and out-of-tree but for meson `'in-project-tree` is not allowed, and for all other `'outside-of-project-tree` is not allowed. So it really is a preference of sorts.

mattiasb:
Thinking some more about this, I just realized some of the build systems (like cargo) already work like meson (and only build out of dir). The difference is that for meson and CMake the compilation directory is configurable. Also I was trying to make a difference between CMake and meson in that meson can't build in the source dir but that becomes very obvious as soon as you try to build in 

If I let the changes be around just a configurable compilation dir (but with project defaults) it all becomes a lot easier to follow I believe.

mattiasb:
I put up my changes in `fixup!`- and `squash!`- commits since I believe that's easier to review.

Please squash / fixup with `git rebase --autosquash -i master` before merging.

mattiasb:
@damienrg one thing you could do is try it out and see if I've forgotten something.

bbatsov:
I'm fine with merging this in its current state, if you just squash the fixup commits.

mattiasb:
There. Rebased against master as well.

PR #1189, opened 2017-10-24
support relative ignore for alien indexing

### LINKED ISSUE #1184: relative ignore does not work
## Expected behavior
trying projectile on chromium code base, I have a .projectile file:

-*.png
-*.java

.png & .java files should not appear in projectile find file

## Actual behavior
these files still appear

clone chromium's code
inside chromium, under the root of git repo, create a .projectile file containing
```
-*.png
-*.java
```

and try to find file using projectile

```
Projectile version: 20171009.848
```

### Emacs version

```
emacs versin: 27.0.50
```

### Operating system

Mac

fantasticsid:
Projectile relative glob ignore used to work but now it does not work if indexing method is alien (native indexing still works). This PR adds support for glob ignore even if indexing is alien, which is default on non Windows machines.

fixes #1184 
-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -42,6 +42,7 @@
 (require 'ibuf-ext)
 (require 'compile)
 (require 'grep)
+(require 'em-glob)
 
 (eval-when-compile
   (defvar ag-ignore-list)
@@ -1292,7 +1293,8 @@ If ignored directory prefixed with '*', then ignore all
 directories/subdirectories with matching filename,
 otherwise operates relative to project root."
   (let ((ignored-files (projectile-ignored-files-rel))
-        (ignored-dirs (projectile-ignored-directories-rel)))
+        (ignored-dirs (projectile-ignored-directories-rel))
+        (filtering-regexes (mapcar #'eshell-glob-regexp (car (projectile-filtering-patterns)))))
     (cl-remove-if
      (lambda (file)
        (or (cl-some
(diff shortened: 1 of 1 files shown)


bbatsov:
Some unit tests demonstrating what exactly does this solve would be useful.

projectile.el
```diff
@@ -1292,7 +1293,8 @@ If ignored directory prefixed with '*', then ignore all
 directories/subdirectories with matching filename,
 otherwise operates relative to project root."
   (let ((ignored-files (projectile-ignored-files-rel))
-        (ignored-dirs (projectile-ignored-directories-rel)))
+        (ignored-dirs (projectile-ignored-directories-rel))
+        (filtering-regexes (mapcar #'eshell-glob-regexp (car (projectile-filtering-patterns)))))
```
bbatsov: This bit of code puzzles me - why take just the `car` of the filtering patterns?

bbatsov:
I wonder if I didn't change this intentionally - the problem with applying filters to the results returned by something like `find` is that this filtering is very slow on bigger projects and that forces you to use caching. For me it's preferable for people to simply leverage whatever configuration tools like `pt` and `ag` provide, so there would be no slowdowns. For the native indexing that's not a big problems as it's already pretty slow. :-)

bbatsov:
I'm closing this PR due to no recent activity. Feel free to re-open it if you ever come back to it.

Issue #1121, opened 2017-03-01
Toggling between implementation and test fails in large rails codebase


levicole:
## Expected behavior

In a large rails code base, I expect `projectile-toggle-between-implementation-and-test` to toggle between my test file and implementation file.

## Actual behavior

This works fine in a small rails project, but it doesn't in my organizations monolith. When I run the `M-x projectile-toggle-between-implementation-and-test` (I'm using spacemacs so it's bound to `SPC p a`) in both a test file and an implementation file I receive the following message:

> No matching test file found

I suspect this has to do with the fact that this is a large rails app. It seems like (I'm new to emacs list, so forgive me if I'm wrong), that this method searches through every file in the project to find the appropriate associated file. In a large rails app like mine (or the one I link to in steps to reproduce), it seems as if it's timing out while searching.

## Steps to reproduce the problem

I can't give you my repo, but I've cloned the [Discourse](https://github.com/discourse/discourse) repo and I can reproduce my issue.

1. Clone the discourse repository
2. Open app/models/user.rb
3. Try to toggle between test and implementation

Should be noted that I'm using Spacemacs and it includes projectile-rails. I wasn't sure if the issue was there, or here.
 
### Projectile version information
Projectile version: 20170106.606 

### Emacs version

GNU Emacs 25.1.1 (x86_64-apple-darwin13.4.0, NS appkit-1265.21 Version 10.9.5 (Build 13F1911))

### Operating system

OSX 10.11.16

ricardoxvpn:
+1

darrenboyd:
I looked into this issue (as I'm also suffering from it) and discovered the following...

In the discourse repo linked to above, `projectile-project-info` reports the _project type_ as `ruby-test`. However, it should be `rails-rspec`. If the _test_ folder is deleted (i.e., `rm -fr test` from the project dir), then projectile will see the project type as `ruby-rspec`, and will properly switch between test (spec in this case) and implementation.

In the project I'm working on, which is also a Rails with RSpec project, the _project  type_ is reported as `npm`. If I remove the `package.json` file, projectile will see the project as `ruby-rspec`, and the toggle between implementation and spec will work properly.

I found documentation* that seems to suggest certain things can be overridden with a `.dir-locals.el` file in the project root. However, I was unable to get the project type explicitly set, due to insufficient elisp skills :slightly_frowning_face:.

* https://github.com/bbatsov/projectile/blob/master/doc/configuration.md#storing-project-settings

[...]

PR #1191, opened 2017-11-02
Update `projectile-detect-project-type` for detecting order.

kayhide:
This fixes `projectile-toggle-between-implementation-and-test` in recent rails apps.

Recent rails apps typically include `package.json` in the project root.
This messes up the projectile's project-type detection and sets it `npm` despite `rails-spec` or `rails-test`.

This also fixes: https://github.com/bbatsov/projectile/issues/1121#issuecomment-300851052

### Detail

Original `projectile-detect-project-type` is looking into the `projectile-project-types` hash in **reverse** order of registration.

So if the registrations go in the order of this (and currently they do):

```
(projectile-register-project-type 'rails-rspec '("Gemfile" "app" "lib" "db" "config" "spec")
                                  :compile "bundle exec rails server"
                                  :test "bundle exec rspec")
(projectile-register-project-type 'npm '("package.json")
                                  :compile "npm install"
                                  :test "npm test")
```

and  `package.json` exists, npm beats rails-rspec no matter how other files and dirs are.

Seeing the registration order of the current source code, it likely assumes that the latter, the less strict.
It seems more natural for detection to go in the order of registration.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2402,7 +2402,7 @@ Fallsback to a generic project type when the type can't be determined."
                                (if (listp marker)
                                    (and (projectile-verify-files marker) project-type)
                                  (and (funcall marker) project-type))))
-                           (projectile-hash-keys projectile-project-types))
+                           (reverse (projectile-hash-keys projectile-project-types)))
                           'generic)))
     (puthash (projectile-project-root) project-type projectile-project-type-cache)
     project-type))
--- a/projectile.el
+++ b/projectile.el
@@ -2289,12 +2289,6 @@ TEST-PREFIX which specifies test file prefix."
 
 (projectile-register-project-type 'emacs-cask '("Cask")
                                   :compile "cask install")
-(projectile-register-project-type 'rails-rspec '("Gemfile" "app" "lib" "db" "config" "spec")
-                                  :compile "bundle exec rails server"
-                                  :test "bundle exec rspec")
-(projectile-register-project-type 'rails-test '("Gemfile" "app" "lib" "db" "config" "test")
-                                  :compile "bundle exec rails server"
-                                  :test "bundle exec rake test")
 (projectile-register-project-type 'symfony '("composer.json" "app" "src" "vendor")
                                   :compile "app/console server:run"
                                   :test "phpunit -c app ")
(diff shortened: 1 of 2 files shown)


bbatsov:
I'm fine with the change, but we should mention somewhere that the order of registration of project type matters. Frankly, I never thought of the problem you brought up. This can solved in other ways as well, but I prefer the simplicity of this approach.

kayhide:
@bbatsov Hi, thanks for your quick reply.

I looked for a nice place for documentation and found this:
https://github.com/bbatsov/projectile/blob/master/doc/configuration.md#adding-custom-project-types

Referring to this documentation, I noticed a problem.
If this pull request is merged, custom project-type registration will not work when its condition conflicts with any existing project-type.
The documented one above is the case.

So now, we need to rethink the approach.
The simplest solution I think is:

- Make detection go in reverse order (as it was), and reorder the preset registration itself

As this way,  users will still be able to overwrite any project-type as they want.
What do you think?

bbatsov:
Sounds reasonable, unfortunately it also sounds a bit brittle.

kayhide:
Yeah, it still does not solve the problem neatly.
Intrinsically, project detection is not deterministic. `rails` project can be `npm` simultaneously,  as so too `ruby-test` and `ruby-rspec`, `haskell-stack` and `make`, `nix` and whatever...

Now, the obvious problem is `rails` part, because now its official generator produces `package.json`.
So, how about just putting `rails-*` registrations  after `npm`?

Detection of `rails` is very strict and does not likely fail.
Like this way, non-`rails` user would not be affected.

projectile.el
```diff
@@ -2387,6 +2381,12 @@ TEST-PREFIX which specifies test file prefix."
 (projectile-register-project-type 'nix '("default.nix")
                                   :compile "nix-build"
                                   :test "nix-build")
+(projectile-register-project-type 'rails-test '("Gemfile" "app" "lib" "db" "config" "test")
```
bbatsov: I'd also add some comment here that the order matters and they should go after `npm`.

kayhide: OK.
I put a comment along with the link of this.

kayhide:
Thanks!

To1ne:
@bbatsov is the order of traversing `projectile-project-types` reversed, cause I feel there is a regression and my rails-rspec project is again recognized as npm.

I've added ` :from-end t` to `projectile-detect-project-type` and then it's properly detected as rails-rspec.

Shall I file a PR to move the registration more to the top? https://github.com/bbatsov/projectile/blob/master/projectile.el#L2458

PR #1192, opened 2017-11-08
Allow the project files to list to expire after a time period

Wilfred:
`projectile-enable-caching` is important for performance for commands like `projectile-find-file`. However, the cache lives indefinitely (even across Emacs restarts) and can go stale if files are added/removed outside of Emacs.

Instead, allow the user to specify how long the cache may live by setting `projectile-files-cache-expire`, similar to `projectile-file-exists-local-cache-expire`.

-----------------

- [x] The new code is not generating bytecode warnings


Wilfred:
I've written set the default of `projectile-files-cache-expire` to 'never expire', although my personal preference is 24 hours. Let me know if you think we should set the default to a different value.

bbatsov:
I think a much better solution to this would be to monitor for FS changes and update the cache from them, but timed expiration is still better than nothing.

PR #1193, opened 2017-11-08
Proper file path when auto-creating test file for maven projects

yyoncho:
Projectile was generating incorrect file paths for maven projects.


projectile.el
```diff
@@ -2477,7 +2477,10 @@ PROJECT-ROOT is the targeted directory.  If nil, use
   (let* ((test-file (projectile--test-name-for-impl-name impl-file-path))
          (project-root (projectile-project-root))
          (relative-dir (file-name-directory (file-relative-name impl-file-path project-root)))
-         (test-dir (expand-file-name (replace-regexp-in-string "src/" "test/" relative-dir) project-root))
+         (test-dir (expand-file-name (if (eq (projectile-project-type) 'maven)
```
bbatsov: This should actually be made part of the project registration properties. We can't really have a bunch of `if`s for different project types.

yyoncho: Sure. What will be more appropriate to be added in the project registration: a function generating test file name from impl name or a list of regexp to replace in file path?

bbatsov: Seems a function would be a more flexible approach.

PR #1194, opened 2017-11-09
Allow to cache status of dirty projects

galaunay:
### Motivation
With a long list of known projects, `projectile-browse-dirty-projects` (`C-c p V`) can take a while.

### Modifications proposed
Dirty projects status is now cached and can be retrieved using prefix argument (`C-u C-c p V`).

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1248,18 +1248,19 @@ If PROJECT-PATH is a project, check this one instead."
   "Return the list of dirty projects.
 The list is composed of sublists~: (project-path, project-status).
 Raise an error if their is no dirty project."
-  (message "Checking for modifications in known projects...")
-  (let ((projects projectile-known-projects)
-        (status ()))
-    (dolist (project projects)
-      (when (and (projectile-keep-project-p project) (not (string= 'none (projectile-project-vcs project))))
-        (let ((tmp-status (projectile-check-vcs-status project)))
-          (when tmp-status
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -1257,19 +1261,22 @@ Raise an error if their is no dirty project."
...
     (when (= (length status) 0)
       (message "No dirty projects have been found"))
+    (setq projectile-cached-dirty-projects-status status)
     status))
 
-(defun projectile-browse-dirty-projects ()
-  "Browse dirty version controlled projects."
-  (interactive)
-  (let ((status nil)
+(defun projectile-browse-dirty-projects (cached)
```
bbatsov: I'd make this an optional param, so you wouldn't have to pass it explicitly in non-interactive calls. In general I wonder do you need this as a param, or a configuration option.

galaunay: Good point, I passed `cached` as an optional parameter.

I am not sure what you mean by passing this as a configuration option.
You mean making it a customizable variable ?
If it is the case, I am not sure this would be a good idea, cached dirty project list is only useful when you remember what modifications you have made since the previous dirty project list update (and update is only made when `cached` is nil).

galaunay:
The last commit just fix a bug appearing when using the keyboard during the dirty repos check (the check stops and displays a window showing the currently checked repo state).
Let me know if you think it deserves another pull request.

bbatsov:
👍

Issue #949, opened 2016-01-20
projectile-multi-occur should ignore the TAGS buffer


fommil:
it is possible to ignore the tags buffer by doing this

``` elisp
(setq projectile-globally-ignored-buffers '("TAGS<src>"))
```

but it is not scalable because the buffer name is project-specific (and can change when there is a collision).

It would be better if the `projectile-globally-ignored-buffers` took regex or glob patterns, or if `TAGS` were always ignored as a special case.

fommil:
oh, I just seen that this accepts functions! OK, I'll propose a better default than `nil`.

fommil:
actually, I think just aliasing `(setq projectile-globally-ignored-buffers ido-ignore-buffers)` works well for me.

fommil:
I would have expected this to work, since the docs say that if an element is a buffer it will be ignored

```elisp
(push (get-file-buffer tags-file-name) projectile-globally-ignored-buffers)
```

but `TAGS` still show up in results.

fommil:
adding the tags file to the ignored buffers means that tag searching fails, so it needs to be special cased inside `projectile-multi-occur`

PR #1196, opened 2017-11-14
Allow regexp in globally ignored buffers

panaak:
Not sure if this is the right way, but anyway..)

This allows the `projectile-globally-ignored-buffers` customization to contain regular expressions works pretty much the same as with `projectile-globally-ignored-modes`. Is there a reason, why it wasn't implemented the same way from the beginning?

I'm aware of the fact, that all previously defined values which contain any *regex* chars like `*` etc. would have to be escaped.

This would fix: https://github.com/bbatsov/projectile/issues/949

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -419,7 +419,7 @@ it for functions working with buffers."
   :type '(repeat string))
 
 (defcustom projectile-globally-ignored-buffers nil
-  "A list of buffer-names ignored by projectile.
+  "A list of regular expressions for buffer-names ignored by projectile.
 
 If a buffer is in the list projectile will ignore
 it for functions working with buffers."
@@ -1415,7 +1415,11 @@ this case unignored files will be absent from FILES."
 (defun projectile-ignored-buffer-p (buffer)
   "Check if BUFFER should be ignored."
   (or
-   (member (buffer-name buffer) projectile-globally-ignored-buffers)
+   (cl-some
+    (lambda (ignore-buffer)
+      (string-match-p (concat "^" ignore-buffer "$")
+                      (buffer-name buffer)))
+    projectile-globally-ignored-buffers)
    (with-current-buffer buffer
      (cl-some
       (lambda (mode)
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -419,7 +419,7 @@ it for functions working with buffers."
   :type '(repeat string))
 
 (defcustom projectile-globally-ignored-buffers nil
-  "A list of buffer-names ignored by projectile.
+  "A list of regular expressions for buffer-names ignored by projectile.
```
bbatsov: I think this change is a bit misleading given the name of the defcustom. Might make more sense to just add one more defcustom for regular expressions instead.

panaak:
```
(defcustom projectile-globally-ignored-modes
  '("erc-mode"
    "help-mode"
    "completion-list-mode"
    "Buffer-menu-mode"
    "gnus-.*-mode"
    "occur-mode")
  "A list of regular expressions for major modes ignored by projectile.

If a buffer is using a given major mode, projectile will ignore
it for functions working with buffers."
  :group 'projectile
  :type '(repeat string))
```

panaak: IMHO it might seem a bit redundant to have two defcustoms `projectile-globally-ignored-buffers` and `projectile-globally-ignored-buffers-regexp`, but I can do that if you think it's better solution.

shackra:
Does this PR implements what I'm trying to implement in my pull request pull/1177 ?

panaak:
@shackra: Yes, pretty much it does with a slight difference. Your changes are more broad and affects all the ignored stuff. After fixing the tests, I think it should be good to go.

projectile.el
```diff
@@ -1415,7 +1415,11 @@ this case unignored files will be absent from FILES."
 (defun projectile-ignored-buffer-p (buffer)
   "Check if BUFFER should be ignored."
   (or
-   (member (buffer-name buffer) projectile-globally-ignored-buffers)
+   (cl-some
```
bbatsov: OK, so this here is going to be a huge problem on bigger projects. Something that was very fast will become veeery slow. Keep in mind that you need to apply this functions to some pretty big lists of files here and there. People are already complaining that all the filtering makes projectile super slow and we're about to make it much slower.

Issue #1157, opened 2017-08-01
"you're not in a project" via the shell


fdr:
## Expected behavior

If I "cd" into a project directory, I was able to use projectile. I updated and this broke. Regrettably I do not know the old version, but I updated via `el-get` not long ago to try to get a fix for this bug in handling `.agignore` and it was not yet upstream: https://github.com/bbatsov/projectile/issues/987

## Actual behavior

Now, I get the `projectile-project-root: You’re not in a project` message.

## Steps to reproduce the problem

1. Start emacs
2. Use M-x shell
3. `cd` to a git repository
4. Attempt to use projectile search (or anything)

### Projectile version information

```
Projectile version: 0.15.0snapshot
```

### Emacs version

25.2.1

### Operating system

Fedora 26

jellelicht:
This is also the case when using eshell

bbatsov:
Seems we fixed this, but we forgot to close it.

PR #1197, opened 2017-11-16
[Fix #1171 #1157] Don't use cached values when buffer-file-name is nil

### LINKED ISSUE #1171: Cached values shouldn't be used when buffer-file-name nil (eg shell mode)
## Expected behavior
While cd'ing into different projects in shell mode, I expect C-c p p, C-c p f to show the name of the project I am currently into, along with its file (for the f binding).

## Actual behavior
Some other project name/files are displayed. They are the cached values of projectile-cached-buffer-file-name and projectile-cached-project-root. These were set the first time projectile bindings were issued in a project in that shell buffer, and not recomputed since (especially after cd'ing into a different project).

Open shell buffer. cd into a project. First time name and files should match project cd'ed into. Now cd into a different project. Name and files reported by projectile are that of the first project and not that of the project newly cd'ed into.

```
Projectile 20170727.2351
```

### Proposed patch
Only use cached values if non nil (otherwise they match shell buffer's buffer-file-name nil value) :
```
--- a/projectile.el
+++ b/projectile.el
@@ -883,7 +883,8 @@
   ;; cl-subst to replace this 'none value with nil so a nil value is used
   ;; instead
   (or (cl-subst nil 'none
-                (or (and (equal projectile-cached-buffer-file-name buffer-file-name)
+                (or (and projectile-cached-buffer-file-name
+                         (equal projectile-cached-buffer-file-name buffer-file-name)
                          projectile-cached-project-root)
                     (progn
                       (setq projectile-cached-buffer-file-name buffer-file-name)
@@ -962,7 +963,8 @@
 (defun projectile-project-name ()
   "Return project name."
   (or projectile-project-name
-      (and (equal projectile-cached-buffer-file-name buffer-file-name)
+      (and projectile-cached-buffer-file-name
+           (equal projectile-cached-buffer-file-name buffer-file-name)
            projectile-cached-project-name)
       (progn
         (setq projectile-cached-buffer-file-name buffer-file-name)
```

guillaumekoenig:
Only use cached values if non nil (otherwise they match buffer's
buffer-file-name nil value in shell-mode).

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -909,7 +909,8 @@ will return the current directory, otherwise it'd raise an error."
   ;; cl-subst to replace this 'none value with nil so a nil value is used
   ;; instead
   (or (cl-subst nil 'none
-                (or (and (equal projectile-cached-buffer-file-name buffer-file-name)
+                (or (and projectile-cached-buffer-file-name
+                         (equal projectile-cached-buffer-file-name buffer-file-name)
                          projectile-cached-project-root)
                     (progn
                       (setq projectile-cached-buffer-file-name buffer-file-name)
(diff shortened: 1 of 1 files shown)


bbatsov:
👍

PR #1198, opened 2017-11-18
Treat members of projectile-globally-ignored-file-suffixes as file name suffixes

laz2:
Treat members of `projectile-globally-ignored-file-suffixes` as file name suffixes. Previous treat as file extensions. By example, ignore ".min.js" in web project.

-----------------

Thanks!


bbatsov:
That's a good point. 👍 I think I originally thought of this as extensions, but I guess I misnamed it. Anyways, I agree with this name, this should be the behaviour.

PR #1204, opened 2017-12-07
[Fix #1193] Add :src-dir and :test-dir to project-type

aheaume:
Currently, the source directory and test directory are hardcoded
values ("src/" and "test/", cf. `projectile-create-test-file-for`)

This commit adds `:src-dir` and `:test-dir` as arguments to
`projectile-project-type` to make it easier to control where the tests
file are created.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Added ability to specify test files suffix and prefix at the project registration.
 * [#1154](https://github.com/bbatsov/projectile/pull/1154) Use npm install instead of build.
 * Added the ability to expire old files list caches via `projectile-projectile-files-cache-expire`.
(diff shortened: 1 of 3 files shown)


aheaume:
Hi @bbatsov !

I wanted to be able to customize the `src` and `test` directories on per-project basis.
I was considering adding stuff to the README or to the docs, but the README does not mention project-types and the documentation on project-type does not mention some options (like `:test-prefix`). I can update the documentation with the new options if required.

@yyoncho This is mostly the same as #1193. I've added support for maven projects, I'd love to have your input on this one!

Would you consider reviewing this? Thanks

bbatsov:
Nice change! 👍 

> I was considering adding stuff to the README or to the docs, but the README does not mention project-types and the documentation on project-type does not mention some options (like :test-prefix). I can update the documentation with the new options if required.

It would be really nice if the README featured something on the subject.

PR #1213, opened 2018-01-27
Cache root in non-file-backed buffers

aaronjensen:
-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -905,10 +905,10 @@ will return the current directory, otherwise it'd raise an error."
   ;; instead
   (or (cl-subst nil 'none
                 (or (and projectile-cached-buffer-file-name
-                         (equal projectile-cached-buffer-file-name buffer-file-name)
+                         (equal projectile-cached-buffer-file-name (or buffer-file-name 'none))
                          projectile-cached-project-root)
                     (progn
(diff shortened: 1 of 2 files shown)


aaronjensen:
I'm not sure how to add tests for this... let me know if that's a problem. Thanks!

bbatsov:
Looks good to me! Normally the test for this would involve creating some buffer, let-binding a bit of config, running `projectile-project-root` and checking that the cached file name is nil.

projectile.el
```diff
@@ -905,10 +905,10 @@ will return the current directory, otherwise it'd raise an error."
   ;; instead
   (or (cl-subst nil 'none
                 (or (and projectile-cached-buffer-file-name
-                         (equal projectile-cached-buffer-file-name buffer-file-name)
+                         (equal projectile-cached-buffer-file-name (or buffer-file-name 'none))
```
bbatsov: Btw, we could have also wrapped this in a `(when buffer-file-name` which might have been more clear, but this came to my mind only after I merged this. Anyways, not a big deal.

aaronjensen: Wrapped which? I tried wrapping the whole thing in that previously and it broke some functionality for me.

aaronjensen:
Also, thanks for the merge!

PR #1214, opened 2018-01-28
Add known autoconf and cscope files to projectile-project-root-files

vsedach:


bbatsov:
I think you'd normally have those in a make project, so adding them seems redundant to me.

mattiasb:
Autotools-projects generally doesn't have a toplevel `Makefile` but the project instead has one or several `Makefile.am` and then a `configure.ac` at the top.

vsedach:
Correct, you won't get a Makefile until you run configure.

bbatsov:
Roger that! Thanks for this improvement!

PR #1223, opened 2018-03-11
[Fix #1222] Don't pass nil format string to `format`

### LINKED ISSUE #1222: projectile-configure-project fails for generic project type
Noticed the following bug while working with some generic project. I have a proposed patch for this and I'll make a PR soon.

## Expected behavior

projectile-configure-project should show "Configure command: " for the generic project type, but instead ...

## Actual behavior

projectile-configure-project fails with "Wrong type argument: stringp, nil".

1. Open a generic project.
2. Press `C-c p C` or `M-x projectile-configure-project`.

`M-x projectile-version`: 0.15.0snapshot, commit: 40b01b9d61ed310737d87e86b226bb4e1ed2e0cb

### Emacs version

25.1.1

### Operating system

Debian Linux 9

stephan-cr:
The return value of `projectile-default-configure-command` is checked
for nil, since `format` doesn't accept nil format strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 
 ### Bugs fixed
 
+* [#1222](https://github.com/bbatsov/projectile/issues/1222): `projectile-configure-project` fails for generic project type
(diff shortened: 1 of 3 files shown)


bbatsov:
👍

PR #1225, opened 2018-03-12
Add :safe keywords to all the projectile defcustom

arichiardi:
I am adding `:safe` for our function `defcustom` so that we don't get warnings anymore.

-----------------

I get an error here `No such file or directory, noflet` - using the newer emacs

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -292,6 +292,7 @@ It has precedence over function `projectile-project-name-function'."
 If variable `projectile-project-name' is non-nil, this function will not be used."
   :group 'projectile
   :type 'function
+  :safe (lambda (f) (and (functionp f) (not (unsafep f))))
   :package-version '(projectile . "0.14.0"))
 
 (defcustom projectile-project-root-files
@@ -461,11 +462,13 @@ Any function that does not take arguments will do."
 (defcustom projectile-test-prefix-function 'projectile-test-prefix
   "Function to find test files prefix based on PROJECT-TYPE."
   :group 'projectile
+  :safe (lambda (f) (and (functionp f) (not (unsafep f))))
   :type 'function)
 
 (defcustom projectile-test-suffix-function 'projectile-test-suffix
   "Function to find test files suffix based on PROJECT-TYPE."
   :group 'projectile
+  :safe (lambda (f) (and (functionp f) (not (unsafep f))))
   :type 'function)
 
 
@@ -2292,6 +2295,7 @@ TEST-DIR which specifies the path to the tests relative to the project root."
 (defcustom projectile-go-project-test-function #'projectile-go-project-p
   "Function to determine if project's type is go."
   :group 'projectile
+  :safe (lambda (f) (and (functionp f) (not (unsafep f))))
   :type 'function)
 
 (define-obsolete-variable-alias 'projectile-go-function 'projectile-go-project-test-function "0.15")


thblt:
Sorry, but this is a bad idea, since function variables are intrinsically unsafe.  You should probably use:

``` emacs-lisp
:safe (lambda (f) (and (functionp f) (not (unsafep f))))
```

arichiardi:
@thblt sure thanks actually let me try that. Learning how this stuff works thanks for pointing this out!

thblt:
@arichiardi Actually I [asked on reddit](https://www.reddit.com/r/emacs/comments/83z3q9/security_defcustom_var_def_doc_safe_functionp_is/) and Stefan Monnier (former Emacs maintener) suggested that even `unsafep` wasn't to be trusted.  

I then respectfully suggest this PR should not be merged, since there doesn't seem to be a good general solution to protect against attacks, and such attacks would be trivial to implement.

If you still want to push this feature, an alternative may be to consider only some *symbols* as safe, eg:

``` emacs-lisp
(defcustom ... :safe (lambda (f) (memq f '(projectile-acceptable-function-1 projectile-acceptable-function-2)))
```

What do you think?

Regards,

arichiardi:
@thblt I see.

Maybe we should drop functions there and just use strings?

The fact is that an arbitrary function is called anyways so I wonder what is the definition of an "acceptable" function?

thblt:
>The fact is that an arbitrary function is called anyways

Of course, but with the `:safe` clause your PR adds, any random file becomes able to execute any arbitrary function just by being opened.  Without your PR, this would trigger a warning and prompt user for confirmation before any bindings happen and the function is run.

> Maybe we should drop functions there and just use strings?

I'm sorry, I don't understand what you mean by strings?

arichiardi:
Should have explained better: for test prefix and suffix we could return either a fn or the string directly.

This would solve my problem but not the function problem.

I basically wrap a string that I return in a lamdba at the moment.

bbatsov:
Well, I'm puzzled and I don't know how to best proceed here. I understand both arguments and they all have merits, but they are also kind of conflicting, as it's often the case (usability vs security).

thblt:
@bbatsov  FWIW we've [discussed](https://github.com/clojure-emacs/inf-clojure/pull/146#issuecomment-373772464) a few solutions with @arichiardi in my PR on inf-clojure.

slippycheeze:
> Well, I'm puzzled and I don't know how to best proceed here. I understand both arguments and they all have merits, but they are also kind of conflicting, as it's often the case (usability vs security).

@bbatsov FWIW, and I know this is late, but the PR is obviously being kept up to date: I believe that marking them safe is inappropriate.  It is a security vs usability problem, but having it prompt is the least worst choice if you do want to make function calls accessible from here.  (ie: mark it safe if the value is a string, and nothing else.)

You may wish to point users at `safe-local-variable-values`, but ... the problem is most likely best solved by the original user pressing `!` once when prompted, which will forever treat that variable and value pairing as safe, and save that in their custom settings.

If this was made possible, I'd be unable to use projectile: that would be a sufficiently broad gap that I would be out of compliance with our security policies, at least, without patching it to remove the safe flag.  For better or worse, this is not an unjustified level of security concern.

I appreciate that it is ultimately your call, but I'd strongly advocate for letting the user who wants implicit trust do that themselves, rather than making it in any way a feature of projectile itself; Emacs provides the hooks and settings required, as well as a simple interface to let users assert that, yes, this value is trusted both now, and forever.

PR #1228, opened 2018-03-22
Add support for a prefix argument to projectile-vc

knu:
This makes `C-u M-x projectile-vc` prompt you for a project directory to open.

It currently simply ignores a prefix argument that is interactively given.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
+* [#1228](https://github.com/bbatsov/projectile/pull/1228): Add support for a prefix argument to `projectile-vc`.
 * [#1221](https://github.com/bbatsov/projectile/pull/1221): Modify Ruby and Elixir project settings.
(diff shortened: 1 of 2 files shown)


bbatsov:
Seems like a reasonable addition to me. 👍

cocreature:
This PR seems to have resulted in `C-p v` always asking for a project name instead of only when a prefix argument is passed. My elisp foo is sadly quite weak but looking at the source I can’t find anything that checks for the prefix argument while other functions such as [projectile-invalidate-cache](https://github.com/bbatsov/projectile/blob/bafd7d6422c4e06bfc48540a785c36cff835480b/projectile.el#L677) do seem to check explicitely for the argument.

Am I missing something here or is this a bug?

BTW, keep up the great work, projectile is a joy to use!

knu:
@cocreature Thanks for the catch!  I now realized `&optional` and `(interactive ...)` don't really go well.  I'll file another PR to address the problem.

PR #1229, opened 2018-03-23
Fix projectile-vc calling projectile-completing-read regardless of the prefix argument

knu:
This fixes a bug in #1228.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3135,9 +3135,11 @@ If PROJECT-ROOT is given, it is opened instead of the project
 root directory of the current buffer file.  If interactively
 called with a prefix argument, the user is prompted for a project
 directory to open."
-  (interactive (list (projectile-completing-read
+  (interactive (and current-prefix-arg
+                    (list
+                     (projectile-completing-read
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -3135,9 +3135,11 @@ If PROJECT-ROOT is given, it is opened instead of the project
 root directory of the current buffer file.  If interactively
 called with a prefix argument, the user is prompted for a project
 directory to open."
-  (interactive (list (projectile-completing-read
+  (interactive (and current-prefix-arg
```
bbatsov: Won't it better to use `"P"` in interactive and push the optional reading of the the project in the body of the function?

knu: That'll also work.  I just thought it would be intuitive and misfire-free to keep interactive code inside of the `(interactive)` construct, as I saw `(projectile-vc)` (non-interactive use without an argument) in projectile.el.  However, I don't feel strongly.  I'll leave it up to you.

bbatsov: 👍

PR #1236, opened 2018-05-07
[Fix #1235] support for clojure deps.edn projects

### LINKED ISSUE #1235: Support for clojure clj projects with deps.edn
Sometime ago Cognitect introduced  a `clj` tool along with new way of storing project dependencies (based on `deps.edn` files - https://clojure.org/guides/deps_and_cli). Projectile already supports `boot` and `leiningen` - based projects which, among others, allows out-of-the-box to detect project root directory and easily switch between test/impl files. 

Unfortunately, projects based on deps.edn are treated as "generic" ones. I would like to introduce a simple patch which detects clj projects and brings back all sweet things like correctly working `projectile-toggle-between-implementation-and-test` function.

mbuczko:
This adds support for clojure deps.edn - based projects which also fixes switching between test/implementation files with `projectile-toggle-between-implementation-and-test` function.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -298,6 +298,7 @@ If variable `projectile-project-name' is non-nil, this function will not be used
   '("rebar.config"       ; Rebar project file
     "project.clj"        ; Leiningen project file
     "build.boot"         ; Boot-clj project file
+    "deps.edn"           ; Clojure CLI project file
     "SConstruct"         ; Scons project file
     "pom.xml"            ; Maven project file
     "build.sbt"          ; SBT project file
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -2432,6 +2433,8 @@ TEST-DIR which specifies the path to the tests relative to the project root."
                                   :compile "boot aot"
                                   :test "boot test"
                                   :test-suffix "_test")
+(projectile-register-project-type 'clj '("deps.edn")
+                                  :test-suffix "_test")
```
mbuczko: as far as I know clj has no standarized way of invoking `compile` or `test` tasks (actually it has no "task" notion), so I left these undefined here.

projectile.el
```diff
@@ -2432,6 +2433,8 @@ TEST-DIR which specifies the path to the tests relative to the project root."
                                   :compile "boot aot"
                                   :test "boot test"
                                   :test-suffix "_test")
+(projectile-register-project-type 'clj '("deps.edn")
```
bbatsov: Technically speaking this is not a `clj` project, but `clojure-cli` or `clojure-tools` depending on your perspective. In CIDER I've named it `clojure-cli`.

Issue #850, opened 2015-08-27
Idea: instead of showing "You're not in a project", it can instead trigger projectile-switch-project


robert-stuttaford:
I'll often be in a scratch buffer or similar and want to open a file, and I keep having to butt my head against this warning and then do switch-project anyway.

offby1:
+1

bbatsov:
Guess we have the message prompt you for the next action (switch to project) or make this configurable. I'm fine with changing the existing behavior.

arichiardi:
+1

PR #1239, opened 2018-05-21
Make find-file prompt for a project if not in one

nickdrozd:
This PR presents a partial, hacky solution to https://github.com/bbatsov/projectile/issues/850. Currently, executing `projectile-find-file` outside of a project buffer results in an error message saying "You're not in a project". This is annoying, and invariably my next steps are to switch into a project and then rerun the file command. With this change, if `projectile-find-file` is run outside of a project buffer (say, in the scratch buffer on Emacs startup), it will first prompt for a project (using the same prompt as `projectile-switch-project`) and then prompt for files in that project as usual.

The solution, as I said, is hacky: the desired project root is set in a temporary variable, which is saved and then substituted in for the nonexistent real project root at the last minute. It won't do to set `projectile-cached-project-root` directly, since it will immediately get reset by some call to `projectile-project-root` (by my count, this function is called 81 times in the course of a single call to `projectile-find-file`). I don't know if there is a more elegant way to achieve this, but I've been using it without any issue for a few months now.

I say the solution is partial because similar behavior should be implemented for other commands like `projectile-vc` and `projectile-find-dir`.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2098,12 +2098,21 @@ With INVALIDATE-CACHE invalidates the cache first.  With FF-VARIANT set to a
 defun, use that instead of `find-file'.   A typical example of such a defun
 would be `find-file-other-window' or `find-file-other-frame'"
   (interactive "P")
-  (projectile-maybe-invalidate-cache invalidate-cache)
-  (let ((file (projectile-completing-read "Find file: "
-                                          (projectile-current-project-files)))
-        (ff (or ff-variant #'find-file)))
-    (funcall ff (expand-file-name file (projectile-project-root)))
-    (run-hooks 'projectile-find-file-hook)))
+  (let (projectile-cached-project-root-temp)
+    (when (eq projectile-cached-project-root 'none)
+      (setq projectile-cached-project-root
+            (projectile-completing-read "Switch to project: "
+                                        (projectile-relevant-known-projects)
+                                        :action #'identity))
+      (setq projectile-cached-project-root-temp
+            projectile-cached-project-root))
+    (projectile-maybe-invalidate-cache invalidate-cache)
+    (let ((file (projectile-completing-read "Find file: "
+                                            (projectile-current-project-files)))
+          (ff (or ff-variant #'find-file)))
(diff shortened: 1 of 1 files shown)


bbatsov:
I don't like the proposed solution as it makes the code significantly more complex than it needs to be. The cached project root is not something that has to be manipulated directly pretty much anywhere. It's just an optimization, nothing else.

projectile.el
```diff
@@ -2098,12 +2098,21 @@ With INVALIDATE-CACHE invalidates the cache first.  With FF-VARIANT set to a
 defun, use that instead of `find-file'.   A typical example of such a defun
 would be `find-file-other-window' or `find-file-other-frame'"
   (interactive "P")
-  (projectile-maybe-invalidate-cache invalidate-cache)
-  (let ((file (projectile-completing-read "Find file: "
-                                          (projectile-current-project-files)))
-        (ff (or ff-variant #'find-file)))
-    (funcall ff (expand-file-name file (projectile-project-root)))
-    (run-hooks 'projectile-find-file-hook)))
+  (let (projectile-cached-project-root-temp)
```
bbatsov: I'd just do a check whether I'm in a project or not (we've got some predicate for this) and if I'm not I'd issue switch project and rebind `default-directory` to that project's directory.

bbatsov:
> I say the solution is partial because similar behavior should be implemented for other commands like projectile-vc and projectile-find-dir.

Likely we need a generic function to ensure you're in a project that we can reuse everywhere it makes sense.

bbatsov:
@nickdrozd Ping :-)

bbatsov:
Well, for the solution to be generic it probably has to be part of `projectile-project-root` or some wrapper function. E.g. - we can simply have a defcustom which controls whether it should raise an error if not in project dir or prompt for a project dir.

bbatsov:
Closing this in favour of https://github.com/bbatsov/projectile/commit/c6c3094968a256b7adf9b7c078caf762f7c4e2fc

PR #1247, opened 2018-06-13
Ensure project-roots are directories

nickdrozd:
Projectile adds projects as directory names, as in `~/project/`.
Manually adding a project root that is not a directory name, as
in `(projectile-add-known-project "~/project")`, works fine, but
Projectile will then add the project root as a directory name, so that
both `~/project` and `~/project/` will be in the list of known
projects.

This change converts `~/project` to `~/project/` before adding it to
the list, thereby avoiding annoying near-duplicates.

-----------------

Thanks!


bbatsov:
Your change broke a couple of unit tests.

nickdrozd:
@bbatsov Fixed the tests and added a new one. Also updated the changelog, but not any docs (this is a very minor change).

bbatsov:
Thanks!

Issue #1080, opened 2016-11-02
Bad autoload on projectile-mode-line


darkfeline:
## Expected behavior

User can customize `projectile-mode-line`.

## Actual behavior

User customizations get clobbered.

## Steps to reproduce the problem

Put in `init.el` before `(package-initialize)`

    (custom-set-variables '(projectile-mode-line '(:eval (if (file-remote-p default-directory)
                                 " P"
                                (format " P[%s]" (projectile-project-name))))))

More specifically, `defcustom` forms shouldn't be autoloaded:

> > Then I looked at the Elisp files that come with Emacs.  I haven't gone
> > through all of them, but it seems almost *none* of them specify
> > autoloading for their defcustoms.
> 
> Yes.  It's generally not a good idea to autoload variables.
> 
> > What gives?  Are users really expected to load the whole library (and that
> > assumes they know which one) before being able to customize a variable
> > in it?
>   
> Yes.

https://lists.gnu.org/archive/html/help-gnu-emacs/2007-06/msg00216.html
https://lists.gnu.org/archive/html/help-gnu-emacs/2007-06/msg00360.html

### Projectile version information

```
Projectile version: 20161008.47
```

### Emacs version

25.1.1

### Operating system

Arch Linux

PR #1251, opened 2018-06-21
Don't autoload defcustom

aidalgol:
Remove declaration of `projectile-mode-line` as an autoload.

Fixes bbatsov/projectile#1080


bbatsov:
I don't remember why this was done in the first place, so I'm fine with removing it.

PR #1255, opened 2018-07-10
Add support for function symbols as project default commands

andreas-marschke:
This allows for the implementation of IDE-like features where projectile could context-sensitively change the value of the command to execute for a given command type.

An example usage of this feature would go as follows:

```elisp
(defun my/compile-command ()
  "Returns a String representing the compile command to run for the given context"
  (cond
   ((and (eq major-mode 'java-mode)
         (not (string-match-p (regexp-quote "\\.*/test/\\.*") (buffer-file-name (current-buffer)))))
    "./gradlew build")
   ((eq major-mode 'web-mode)
    "./gradlew compile-templates")
   ))

(defun my/test-command ()
  "Returns a String representing the test command to run for the given context"
  (cond
   ((eq major-mode 'js-mode) "grunt test") ;; Test the JS of the project
   ((eq major-mode 'java-mode) "./gradlew test") ;; Test the Java code of the project
   ((eq major-mode 'my-mode) "special-command.sh") ;; Even Special conditions/test-sets can be covered
   ))

(projectile-register-project-type 'has-command-at-point '("file.txt")
                                  :compile 'my/compile-command
                                  :test 'my/test-command)
```

In cases where one projectile project covers multiple different build types and commands to run
subdivisions of the testset this is a major improvement.

I added a subsection in the documentation regarding this option when defining a new project-type.

The CHANGELOG.md now references this changeset as well.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3245,7 +3245,7 @@ to function `funcall's. Return value of function MUST be string to be executed a
       (if (fboundp command)
         (funcall (symbol-function command))))
      (t
-      (error (format "The value for: %s in project-type: %s was neither a function nor a string." command-type project-type))))))
+      (user-error "The value for: %s in project-type: %s was neither a function nor a string." command-type project-type)))))
 
 (defun projectile-default-configure-command (project-type)
   "Retrieve default configure command for PROJECT-TYPE."
--- a/projectile.el
+++ b/projectile.el
@@ -3233,25 +3233,38 @@ Should be set via .dir-locals.el.")
 It takes precedence over the default command for the project type when set.
 Should be set via .dir-locals.el.")
 
+(defun projectile-default-generic-command (project-type command-type)
+  "Generic retrieval of COMMAND-TYPEs default cmd-value for PROJECT-TYPE.
+
+If found, checks if value is symbol or string. In case of symbol resolves
+to function `funcall's. Return value of function MUST be string to be executed as command."
+  (let ((plist-retrieved-generic-command (plist-get (gethash project-type projectile-project-types) command-type)))
+    (cond
+     ((stringp plist-retrieved-generic-command) plist-retrieved-generic-command)
+     ((symbolp 'plist-retrieved-generic-command)
+      (if (fboundp plist-retrieved-generic-command)
+          (funcall (symbol-function plist-retrieved-generic-command))))
+     )))
+
 (defun projectile-default-configure-command (project-type)
   "Retrieve default configure command for PROJECT-TYPE."
-  (plist-get (gethash project-type projectile-project-types) 'configure-command))
+  (projectile-default-generic-command project-type 'configure-command))
 
 (defun projectile-default-compilation-command (project-type)
   "Retrieve default compilation command for PROJECT-TYPE."
-  (plist-get (gethash project-type projectile-project-types) 'compile-command))
+  (projectile-default-generic-command project-type 'compile-command))
 
 (defun projectile-default-compilation-dir (project-type)
   "Retrieve default compilation directory for PROJECT-TYPE."
-  (plist-get (gethash project-type projectile-project-types) 'compilation-dir))
+  (projectile-default-generic-command project-type 'compilation-dir))
 
 (defun projectile-default-test-command (project-type)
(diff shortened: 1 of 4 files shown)


bbatsov:
Yeah, this seems like a welcome improvement. Let me know when this is ready for review.

andreas-marschke:
@bbatsov Thanks for the quick reply. Does the documentation work for you?

projectile.el
```diff
@@ -3182,25 +3182,38 @@ Should be set via .dir-locals.el.")
...
 
+(defun projectile-default-generic-command (project-type command-type)
+  "Generic retrieval of COMMAND-TYPEs default cmd-value for PROJECT-TYPE.
+
+If found, checks if value is symbol or string. In case of symbol resolves
+to function `funcall's. Return value of function MUST be string to be executed as command."
+  (let ((plist-retrieved-generic-command (plist-get (gethash project-type projectile-project-types) command-type)))
+    (cond
+     ((stringp plist-retrieved-generic-command) plist-retrieved-generic-command)
+     ((symbolp 'plist-retrieved-generic-command)
```
bbatsov: This doesn't look correct to me - here you're passing a quoted symbol to symbolp, which means this will always eval to true.

projectile.el
```diff
@@ -3182,25 +3182,38 @@ Should be set via .dir-locals.el.")
 It takes precedence over the default command for the project type when set.
 Should be set via .dir-locals.el.")
 
+(defun projectile-default-generic-command (project-type command-type)
+  "Generic retrieval of COMMAND-TYPEs default cmd-value for PROJECT-TYPE.
+
+If found, checks if value is symbol or string. In case of symbol resolves
+to function `funcall's. Return value of function MUST be string to be executed as command."
+  (let ((plist-retrieved-generic-command (plist-get (gethash project-type projectile-project-types) command-type)))
```
bbatsov: I'd just call this `command`.

projectile.el
```diff
@@ -3182,25 +3182,38 @@ Should be set via .dir-locals.el.")
...
+    (cond
```
bbatsov: Probably this `cond` should have a default branch raising some error if the command is not a symbol or a string.

bbatsov:
Ping :-)

projectile.el
andreas-marschke: For the error condition `(cond (t ... ))` what would your preferred method of warning be? Simple `(error "")` or would you prefer a full `(throw ... )`?

Also it'd be interesting to know how to test this. I was thinking about this:
```elisp
  (catch 'err
      (projectile-default-compilation-command 'has-command-at-point))
    (error
     (should
      (equal "The value for: default-compilation-command in project-type: has-command-at-point was neither a function nor a string."
             (car (cdr err))))))
```

But that seems to fail in ert. If there is a better way I'm happy to learn. :)

andreas-marschke:
@bbatsov Apologies I was occupied with refactoring my `.emacs.d` from pre-Emacs24 to using  `use-package` and friends and inevitably forgot I needed to resolve this.

bbatsov:
Can you also handle the conflict in the docs and rebase on top of the current master?

bbatsov:
@andreas-marschke ping :-)

bbatsov:
@andreas-marschke ping 2 :-)

andreas-marschke:
@bbatsov Thank you, for your patience and your persistence.

bbatsov:
Thank you! :-)

velimir0xff:
Seems like this broke compile command for all projects w/o `:compilation-dir` defined. At least for me on attempt to compile project, command fails with a message:
```
user-error: The value for: compilation-dir in project-type: rebar was neither a function nor a string.
```

velimir0xff:
https://github.com/bbatsov/projectile/blob/35be8e4ffeabb50bb991a3ee6262f80fcc5622b8/projectile.el#L3347-L3356

`projectile-compilation-dir` function has a fallback to `projectile-project-root` as a default compilation directory, but `projectile-default-generic-command` throws an error on attempt to a get `command-type` from a hash, which is not defined for some project type.

PR #1257, opened 2018-07-24
Avoid "No projects needed to be removed." messages in global mode

remvee:
When enabling projectile-global-mode every time a new file buffer is
opened, a "No projects needed to be removed." message is triggered.
These messages are not helpful to the user, clutter the messages buffer,
may obscure other more informative messages in the mini buffer, and are
rather annoying.


bbatsov:
Nice improvement indeed! Thanks!

PR #1259, opened 2018-07-25
[Fix #1252] Switch to unconnected remote projects

### LINKED ISSUE #1252: Switching to a remote project does not work
## Expected behavior

Switching to a remote project (for example  `/ssh:someserver:/path/to/project`) with `C-c p-p` should bring to that project.

## Actual behavior

An error message "You're not in a project" is displayed.

1 - Add the remote project to `.emacs.d/projectild-bookmarks.eld`, for example `"/ssh:someserver:/path/to/project"`
2 - Switch to that project with `C-c p-p`

A workaround is to open the project directory with `C-x f /ssh:someserver:/path/to/project`) first. Once this is has been done, switching to that project works as expected.

```
Projectile version: 20180621.2239
```

### Emacs version

GNU Emacs 26.1 

### Operating system

MacOS 10.13.5

juergenhoetzel:
Fixes "You're not in a project" error when remote is not connected yet.

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3479,7 +3479,8 @@ Invokes the command referenced by `projectile-switch-project-action' on switch.
 With a prefix ARG invokes `projectile-commander' instead of
 `projectile-switch-project-action.'"
   (interactive "P")
-  (let ((projects (projectile-relevant-known-projects)))
+  (let ((projects (projectile-relevant-known-projects))
+        (projectile-require-project-root nil))
     (if projects
         (projectile-completing-read
          "Switch to project: " projects


bbatsov:
Can you explain to me how this solves the problem as it seems a bit unrelated.

juergenhoetzel:
> Can you explain to me how this solves the problem as it seems a bit unrelated.

This was inspired by the (workaround?) code projectile already uses in:
```elisp
(defun projectile-find-file-in-known-projects ()
  "Jump to a file in any of the known projects."
  (interactive)
  (let ((projectile-require-project-root nil))
    (find-file (projectile-completing-read "Find file in projects: " (projectile-all-project-files)))))
```

to prevent `projectile-project-root` signaling error.

If I interpret the callstack of the error correctly there is no need to use the
`projectile-project-root` functionality anyway when called by
`projectile-switch-project-by-name` because `default-directory` is
already set to the root of the project directory.

bbatsov:
Yeah, you're right. In hindsight raising this error was a big mistake and I'll likely remove that at some point.

PR #1261, opened 2018-07-29
Use regexp-quote in projectile-get-other-files

rsbowman:
to handle files with characters like "+" in them correctly.  Such characters should be matched
literally instead of ending up as part of a regexp.  This happens for example with Objective-C,
where it's common to have two files of the form Class+Protocol.{m,h}.


bbatsov:
Thanks!

PR #1262, opened 2018-07-31
Remember explicitly added projects

ericdanan:
When `projectile-track-known-projects-automatically` is nil, we have to add new projects explicitly with `projectile-add-known-project`. But if I do this and then restart emacs, the newly added project is not remembered. To fix this, an optional argument `MERGE` is added to `projectile-add-known-project`. When it is non-nil (which is the case in interactive calls), `projectile-merge-known-projects` is called at the end of the function.

I didn't add the `projectile-merge-known-projects` call unconditionally because I see that `projectile-add-known-project` is called by `projectile-discover-projects-in-search-path` and it seems to me that we don't want to store these projects persistently. But if that's not the case, then it would perhaps be simpler to remove the optional argument and add the `projectile-merge-known-projects` call unconditionally (and remove it from `projectile-track-known-projects-find-file-hook`).

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -722,12 +722,7 @@ The cache is created both in memory and on the hard drive."
 (defun projectile-track-known-projects-find-file-hook ()
   "Function for caching projects with `find-file-hook'."
   (when (and projectile-track-known-projects-automatically (projectile-project-p))
-    (let ((known-projects (and (sequencep projectile-known-projects)
-                               (copy-sequence projectile-known-projects))))
-      (projectile-add-known-project (projectile-project-root))
-      (unless (equal known-projects projectile-known-projects)
-        (projectile-merge-known-projects)))))
-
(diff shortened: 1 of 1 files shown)


bbatsov:
> I didn't add the projectile-merge-known-projects call unconditionally because I see that projectile-add-known-project is called by projectile-discover-projects-in-search-path and it seems to me that we don't want to store these projects persistently. But if that's not the case, then it would perhaps be simpler to remove the optional argument and add the projectile-merge-known-projects call unconditionally (and remove it from projectile-track-known-projects-find-file-hook).

I guess there's no harm to persist projects that were auto-discovered in the path. Otherwise we're making the code more complex than in needs to be IMO.

ericdanan:
OK, I made the `projectile-merge-known-projects` unconditional and force-pushed.

bbatsov:
Thanks!

PR #1269, opened 2018-08-17
Fixes compilation-dir user-error

andreas-marschke:
This fixes the case where projectile-generic-default-command is used by projectile-compilation-dir() to resolve the compilation-dir. If this is not set by the project type we throw a user-error, this isn't handled by
projectile-compilation-dir() so we need to handle this special case gracefully.

The patch adds the special case where the returned default-generic value returned by the function will purely return nil in that case for compilation-dir.

A test was added to validate the fix.

This issue was initially reported by: @velimir0xff  here: https://github.com/bbatsov/projectile/pull/1255#issuecomment-413854798

--- Checklist:
- [?] You've updated the changelog (if adding/changing user-visible functionality)
- [-] You've updated the readme (if adding/changing user-visible functionality)

Thanks!
---
> - [?] You've updated the changelog (if adding/changing user-visible functionality)

@bbatsov It's not exactly user-visible behavior but it is an edge case. Do you want documentation for it?

@velimir0xff Would this fix your case sufficiently?

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3243,7 +3243,10 @@ to function `funcall's. Return value of function MUST be string to be executed a
      ((stringp command) command)
      ((functionp command)
       (if (fboundp command)
-        (funcall (symbol-function command))))
+          (funcall (symbol-function command))))
+     ((and (not command) (eq command-type 'compilation-dir))
+      ;; `compilation-dir' is special in that it is used as a fallback for the root
+      nil)
      (t
       (user-error "The value for: %s in project-type: %s was neither a function nor a string." command-type project-type)))))
 
--- a/test/projectile-test.el
+++ b/test/projectile-test.el
@@ -706,6 +706,16 @@
 
   (should-error (projectile-default-compilation-command 'has-command-at-point)))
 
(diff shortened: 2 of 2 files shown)


velimir0xff:
@andreas-marschke it would certainly fix the issue I encountered, though I'm not sure that `projectile-default-*` is a good place to throw an error. As to me default function should return `nil` if default is undefined and it's up to a caller to decide what to do with undefined values as I suggested in #1268 Wrapper around `projectile--run-project-cmd` maybe?

Here's, for instance, one more example where caller doesn't expect that function would throw an error: 
https://github.com/bbatsov/projectile/blob/35be8e4ffeabb50bb991a3ee6262f80fcc5622b8/projectile.el#L3270-L3287

As to my mind, user notifications about unavailable features due to under-configured projects seems like a separate feature and should've been considered apart from #1255.

bbatsov:
Maybe we should just prompt the user to supply a command interactively in this case or update their config? That seems like a reasonable and generic behaviour.

velimir0xff:
@bbatsov that would work. How about we fix the broken behaviour to not throw an error and then think of a better option of handling under-configured projects in a separate PR?
As long as this issue is open people can't use all projectile features.

Issue #1283, opened 2018-09-20
get-other-files breaks with multiple candidates


TheBB:
## Expected behavior

`projectile-find-other-file` succeeds and finds any of the matching candidate files.

## Actual behavior

`file-name-nondirectory: Wrong type argument: stringp, nil`

## Steps to reproduce the problem

1. Create an empty project: `mkdir test && cd test`
2. `touch .projectile test.h test.c test.cc`
3. Open `test.h`
4. `M-x projectile-mode`
5. `M-x projectile-find-other-file`

### Projectile version information

```
Projectile 1.1.0snapshot
```

(2e585a4)

### Emacs version

GNU Emacs 27.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.10) of 2018-03-06

### Operating system

Ubuntu Linux 17.10.

PR #1284, opened 2018-09-20
Fix get-other-files: mask potentially nil directory

TheBB:
This is a potential fix for #1283.

* [x]  The commits are consistent with our [contribution guidelines](../CONTRIBUTING.md)
* [x]  You've added tests (if possible) to cover your change(s)
* [x]  All tests are passing (`make test`)
* [x]  The new code is not generating bytecode or `M-x checkdoc` warnings
* [x]  You've updated the changelog (if adding/changing user-visible functionality)
* [x]  You've updated the readme (if adding/changing user-visible functionality)

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1949,7 +1949,8 @@ With FLEX-MATCHING, match any file that contains the base name of current file"
          (candidates
           (cl-sort (copy-sequence candidates)
                    (lambda (file _)
-                     (let ((candidate-dirname (file-name-nondirectory (directory-file-name (file-name-directory file)))))
+                     (let ((candidate-dirname (file-name-nondirectory
+                                               (directory-file-name (or (file-name-directory file) "./")))))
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -1949,7 +1949,8 @@ With FLEX-MATCHING, match any file that contains the base name of current file"
          (candidates
           (cl-sort (copy-sequence candidates)
                    (lambda (file _)
-                     (let ((candidate-dirname (file-name-nondirectory (directory-file-name (file-name-directory file)))))
+                     (let ((candidate-dirname (file-name-nondirectory
```
bbatsov: What file is passed here when this breaks? Just the filename without a dir at all?

TheBB: Yes, just the filename. This seems to happen for candidate files in the project root directory.

PR #1290, opened 2018-09-23
[Fix #1205] Add check that project directory exists when switching

### LINKED ISSUE #1205: Puzzling error message when trying to switch to moved project
## Expected behavior

Run projectile-switch-project.
Select a renamed project.
The error message is `You’re not in a project`

$ mkdir /tmp/toto && cd /tmp/toto && echo foo >>test.txt && git init && git add . && git commit -am test
$ emacs test.txt
M-x projectile-find-file

$ rm -rf /tmp/toto
$ emacs
M-x projectile-switch-project
Select toto.

Projectile 20171126.6

### Emacs version

25.1.1

### Operating system

Windows 10

kevinjfoley:
**This change adds a check that the project directory exists when attempting to switch to it.  If it doesn't exist then the project is removed from known projects and an error is reported.**

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3551,25 +3551,28 @@ With a prefix ARG invokes `projectile-commander' instead of
 Invokes the command referenced by `projectile-switch-project-action' on switch.
 With a prefix ARG invokes `projectile-commander' instead of
 `projectile-switch-project-action.'"
-  (let ((switch-project-action (if arg
-                                   'projectile-commander
-                                 projectile-switch-project-action)))
-    (run-hooks 'projectile-before-switch-project-hook)
-    (let ((default-directory project-to-switch))
-      ;; use a temporary buffer to load PROJECT-TO-SWITCH's dir-locals before calling SWITCH-PROJECT-ACTION
-      (with-temp-buffer
-        (hack-dir-local-variables-non-file-buffer))
-      ;; Normally the project name is determined from the current
-      ;; buffer. However, when we're switching projects, we want to
-      ;; show the name of the project being switched to, rather than
-      ;; the current project, in the minibuffer. This is a simple hack
-      ;; to tell the `projectile-project-name' function to ignore the
-      ;; current buffer and the caching mechanism, and just return the
-      ;; value of the `projectile-project-name' variable.
-      (let ((projectile-project-name (funcall projectile-project-name-function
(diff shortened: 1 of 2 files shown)


bbatsov:
Btw, it seems to me that this can be solved also by just having people add the project-cleanup function to their before-switch hook.

projectile.el
```diff
@@ -3551,25 +3551,28 @@ With a prefix ARG invokes `projectile-commander' instead of
...
-      ;; show the name of the project being switched to, rather than
-      ;; the current project, in the minibuffer. This is a simple hack
-      ;; to tell the `projectile-project-name' function to ignore the
-      ;; current buffer and the caching mechanism, and just return the
-      ;; value of the `projectile-project-name' variable.
-      (let ((projectile-project-name (funcall projectile-project-name-function
-                                              project-to-switch)))
-        (funcall switch-project-action)))
-    (run-hooks 'projectile-after-switch-project-hook)))
+  (if (projectile-project-p project-to-switch)
```
bbatsov: I think the real thing to check for here would be whether the target dir exists at all.

bbatsov:
There's also a deeper problem - obviously most commands would fail if someone deletes the project outside of Emacs. Do we really want to solve this or not? I guess for project switching it makes sense to do some existence check, but everything else would be an overkill.

projectile.el
kevinjfoley: That's what I originally checked, however if the target dir still exists but is no longer a project (`.git` is gone, etc.) then there's still going to be issues.

Checking `projectile-project-p` should cover both cases.

kevinjfoley: Let me know if you think there should be a check for both with more specific errors.  Or I can add more info to the single error like `Check the directory still exists and is a valid project` or something.

kevinjfoley:
I agree, this seems like a case that won't come up very often so anything more than an existence check seems like too much.

projectile.el
bbatsov: I guess the current arrangement will be fine. Just one small suggestion - move the check to its own `when` that simply raises the `error` if needed. That way we won't have to nest all the code under the top `if`.

kevinjfoley:
I've moved the check into it's own `unless` - let me know if you were suggesting something different

PR #1300, opened 2018-09-29
[Fix #1299] Support `nil` for `projectile-git-submodule-command`

### LINKED ISSUE #1299: Setting projectile-git-submodule-command to nil causes error
## Expected behavior
`(setq projectile-git-submodule-command nil)` disables recursing into Git submodules per the docstring

## Actual behavior

```
  call-process("bash" nil t nil "-c" nil)
  apply(call-process "bash" nil t nil ("-c" nil))
  process-file("bash" nil t nil "-c" nil)
  shell-command-to-string(nil)
  projectile-files-via-ext-command(nil)
  projectile-get-immediate-sub-projects("/Users/andrew/.dotfiles/emacs/.emacs.d/")
  projectile-get-all-sub-projects("/Users/andrew/.dotfiles/emacs/.emacs.d/")
  projectile-get-sub-projects-files()
  projectile-get-repo-files()
  projectile-dir-files-external("/Users/andrew/.dotfiles/emacs/.emacs.d/" "/Users/andrew/.dotfiles/emacs/.emacs.d/")
  projectile-dir-files("/Users/andrew/.dotfiles/emacs/.emacs.d/" "/Users/andrew/.dotfiles/emacs/.emacs.d/")
```
`(setq projectile-git-submodule-command nil)`
Call `projectile-find-file` in a repo with submodules.

```
Projectile 1.1.0snapshot
```

### Emacs version

GNU Emacs 26.1 (build 1, x86_64-apple-darwin14.5.0, NS appkit-1348.17 Version 10.10.5 (Build 14F2511)) of 2018-05-30

### Operating system

OS X

andschwa:
- [ ] The new code is not generating bytecode or `M-x checkdoc` warnings **this doesn't pass anyway**
- [ ] You've updated the changelog (if adding/changing user-visible functionality) **this fix corrects behavior to existing documentation**
- [ ] You've updated the readme (if adding/changing user-visible functionality) **see above**

Resolves #1299.

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1270,9 +1270,13 @@ dirty project list."
                                 :action 'projectile-vc)))
 
 (defun projectile-files-via-ext-command (root command)
-  "Get a list of relative file names in the project ROOT by executing COMMAND."
-  (let ((default-directory root))
-   (split-string (shell-command-to-string command) "\0" t)))
+  "Get a list of relative file names in the project ROOT by executing COMMAND.
+
+If `command' is nil or an empty string, return nil.
+This allows commands to be disabled."
+  (when (stringp command)
+    (let ((default-directory root))
+      (split-string (shell-command-to-string command) "\0" t))))
 
 (defun projectile-index-directory (directory patterns progress-reporter)
   "Index DIRECTORY taking into account PATTERNS.
--- a/test/projectile-test.el
+++ b/test/projectile-test.el
@@ -198,6 +198,10 @@
     (let ((projectile-indexing-method 'alien))
       (should (equal '("a/b/c" "a/d/e") (projectile-dir-files "/my/root/" "a/"))))))
 
+(ert-deftest projectile-test-files-via-ext-command ()
+  (should (not (projectile-files-via-ext-command "" "")))
+  (should (not (projectile-files-via-ext-command "" nil))))
+
 (ert-deftest projectile-test-setup-hook-functions-projectile-mode ()
   (noflet ((projectile--cleanup-known-projects () nil)
            (projectile-discover-projects-in-search-path () nil))
--- a/test/projectile-test.el
+++ b/test/projectile-test.el
@@ -198,6 +198,10 @@
     (let ((projectile-indexing-method 'alien))
       (should (equal '("a/b/c" "a/d/e") (projectile-dir-files "/my/root/" "a/"))))))
 
+(ert-deftest projectile-test-git-submodule-command ()
+  (should (string-prefix-p "git" (projectile-get-sub-projects-command 'git)))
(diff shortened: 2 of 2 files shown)


test/projectile-test.el
```diff
@@ -198,6 +198,13 @@
     (let ((projectile-indexing-method 'alien))
       (should (equal '("a/b/c" "a/d/e") (projectile-dir-files "/my/root/" "a/"))))))
 
+(ert-deftest projectile-test-git-submodule-command ()
```
andschwa: Took a good guess as to where in the file to place this new test. Let me know if you want it moved.

projectile.el
```diff
@@ -1095,6 +1095,7 @@ Files are returned as relative paths to the project ROOT."
 (defun projectile-get-sub-projects-command ()
   (let ((vcs (projectile-project-vcs)))
     (cond
+     ((not projectile-git-submodule-command) "")
      ((eq vcs 'git) projectile-git-submodule-command)
```
bbatsov: I think you actually need to add an if on the next line, give that all of this is `git`-specific. Or maybe better - make the receivers of this method handle `nil` instead of an empty string. I guess this change from nil to "" broke the functionality in the first place.

andschwa: > make the receivers of this method handle nil instead of an empty string.

If you're willing to take that, it is definitely better!

andschwa: Fixed up. Much simpler than I expected. And added extra tests cause why not.

projectile.el
```diff
@@ -1260,8 +1260,11 @@ dirty project list."
                                 :action 'projectile-vc)))
 
 (defun projectile-files-via-ext-command (command)
-  "Get a list of relative file names in the project root by executing COMMAND."
-  (split-string (shell-command-to-string command) "\0" t))
+  "Get a list of relative file names in the project root by executing COMMAND.
+
+If `command' is nil or an empty string, return nil.
+This allows commands to be disabled."
+  (when command (split-string (shell-command-to-string command) "\0" t)))
```
andschwa: `s/when/stringp`?

bbatsov: Yeah. I guess it'd be best if it was `stringp` and some non-empty check, as it doesn't make much sense to invoke an empty string command.

bbatsov: Although if `shell-command-to-string` does nothing with an empty string then we are fine.

andschwa: Yeah `(shell-command-to-string "")` returns `""`; it just raises on `nil`, which was my problem.

andschwa:
Rebased and (my) tests fixed.

bbatsov:
Thanks!

andschwa:
@bbatsov And thanks to you!

PR #1308, opened 2018-10-02
Fix get-immediate-sub-projects

DamienCassou:
The call to `projectile-files-ext-command` returns filenames relative
to the project root, not to the current directory.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1206,7 +1206,7 @@ they are excluded from the results of this function."
          ;; search for sub-projects under current project `project'
          (submodules (mapcar
                       (lambda (s)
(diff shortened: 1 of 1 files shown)


DamienCassou:
I haven't added any test because there were none for `projectile-get-immediate-sub-projects` already.

bbatsov:
@DamienCassou Nice catch! I guess before my recent refactorings `default-directory` was just let-bound to be the same as `path` and I totally missed this.

Thanks!

> I haven't added any test because there were none for projectile-get-immediate-sub-projects already.

Yeah, sadly we've got almost no tests for anything right now.

PR #1309, opened 2018-10-02
 Migrate `projectile-test-ignored-directory-p' to buttercup 

kevinjfoley:
-----------------

Thanks!

### DIFF
--- a/test/projectile-test.el
+++ b/test/projectile-test.el
@@ -100,8 +100,11 @@ test temp directory"
       (expect projectile-known-projects :to-equal nil))))
 
 (describe "projectile-test-ignored-directory-p"
-  (it "ignores specified directories"
-    (spy-on 'projectile-ignored-directories :and-return-value '("/path/to/project/tmp" "/path/to/project/t\\.*"))
+  (it "ignores specified literal directory values"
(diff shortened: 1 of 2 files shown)


kevinjfoley:
Only have one test in this PR to make sure I have the hang of things.

Also I think the test should be modified slightly because as is the regular expression covers both truthy test cases so we should change `/path/to/project/tmp` to be something that doesn't match the regex and test that.  I can add that to this PR if you agree or open a separate one if that's more appropriate.

bbatsov:
Feel free to add this to the current PR. The old tests are pretty bad and should not be copied one to one.

test/projectile-test.el
```diff
@@ -98,3 +98,10 @@ test temp directory"
           (projectile-known-projects-file (projectile-test-tmp-file-path)))
       (projectile-clear-known-projects)
       (expect projectile-known-projects :to-equal nil))))
+
+(describe "projectile-test-ignored-directory-p"
+  (it "ignores specified directories"
```
bbatsov: I think you can have a separate `if` for the regexp check to highlight it.

kevinjfoley: Good call, I added a separate spec for it

bbatsov:
Thanks!

PR #1310, opened 2018-10-02
[Fix #1016] Switching projects does not include the current project

### LINKED ISSUE #1016: Switching projects does not include the current project
## Expected behavior

The list of projects to switch to should include the current project. My workflow includes C-c p p to open other files of the current project, but this seemed to have stopped working with an update.
## Actual behavior

When switching projects from a buffer of a file whose project is tracked, the list of projects to switch to does not include the project of the file in the buffer.

C-c p p to switch to a project, and open a file. C-c p p again, and the same project is not listed.

```
Projectile version: MELPA projectile-20160526.832
```
### Emacs version

24.5.1
### Operating system

Linux

s-surineni:
Added a function projectile--move-current-project-to-end which will move current project to the end of the list while switching projects. Earlier it was being removed

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3472,13 +3472,33 @@ An open project is a project with any open buffers."
                              (list (abbreviate-file-name (projectile-project-root))))
     projects))
 
+(defun projectile--move-current-project-to-end (projects)
+  "Move current project (if any) to the end of list in the list of PROJECTS"
+  (if (projectile-project-p)
+      (append
+       (projectile--remove-current-project projects)
+       (list (abbreviate-file-name (projectile-project-root))))
+    projects))
+
 (defun projectile-relevant-known-projects ()
-  "Return a list of known projects except the current one (if present)."
-  (projectile--remove-current-project projectile-known-projects))
+  "Return a list of known projects."
+  (cond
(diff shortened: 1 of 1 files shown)


s-surineni:
I don't know if I have written changelog properly. Please point me toward correct way if I am wrong.

bbatsov:
I think you deleted the entire changelog by accident. :-)

s-surineni:
Should I undo the changes to changelog?

s-surineni:
Just hit me that I am missing `if (projectile-project-p)` check and added it

projectile.el
```diff
@@ -3472,13 +3472,21 @@ An open project is a project with any open buffers."
                              (list (abbreviate-file-name (projectile-project-root))))
     projects))
 
+(defun projectile--move-current-project-to-end (projects)
```
bbatsov: I think you should add some defcustom for selecting this behavior or the old one (as discussed in the ticket). It can be named `projectile-current-project-on-switch` with values like `'remove`, ''move-to-end`, `'keep`. 

I also think this function would benefit from an unit test.

bbatsov:
Thanks!

PR #1312, opened 2018-10-02
Previous / next project buffer commands

### LINKED ISSUE #896: Cycle through current project's buffer's only
Is there a way to cycle through buffers like with next-buffer/previous-buffer but limited only to buffers associated with the current project?  If not, this would be great -- perhaps projectile could add [buffer ring](http://www.emacswiki.org/emacs/BufferRing) integration to achieve this?

mookid:
Add commands to switch to previous / next buffers in the current project.

- [x] update menu
- [x] update manual
- [x] handle non-project case

Fix #896.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4044,6 +4044,8 @@ thing shown in the mode line otherwise."
         ["Kill project buffers" projectile-kill-buffers]
         ["Save project buffers" projectile-save-buffers]
         ["Recent files" projectile-recentf]
+        ["Previous buffer" projectile-previous-project-buffer]
+        ["Next buffer" projectile-next-project-buffer]
         "--"
         ["Toggle project wide read-only" projectile-toggle-project-read-only]
         ["Edit .dir-locals.el" projectile-edit-dir-locals]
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -4018,6 +4019,8 @@ thing shown in the mode line otherwise."
     (define-key map (kbd "x t") #'projectile-run-term)
     (define-key map (kbd "x s") #'projectile-run-shell)
     (define-key map (kbd "z") #'projectile-cache-current-file)
+    (define-key map (kbd "<left>") #'projectile-previous-project-buffer)
```
bbatsov: Those should also be mentioned in the menu and the manual.

projectile.el
```diff
@@ -4137,6 +4140,32 @@ Otherwise behave as if called interactively.
...
+    (when (cdr-safe projectile-project-buffers)
+      (while (and (< counter max-iterations)
+                  (not (gethash (current-buffer) other-project-buffers)))
+        (apply orig-fun args)
+        (incf counter)))))
+
+(defun projectile-next-project-buffer ()
+  "In selected window switch to the next project buffer."
+  (interactive)
+  (projectile--repeat-until-project-buffer #'next-buffer))
```
bbatsov: How does the command behave when invoked outside a project? I guess it should just show some error as suggesting to switch to a project in this case doesn't make much sense.

mookid: alternatively, it may (warn and) forward to previous-buffer / next-buffer

bbatsov: Yeah, that might actually be even better.

bbatsov:
Thanks!

PR #1313, opened 2018-10-03
tests for add-known-project, configure-command, get-all-sub-projects

jmingtan:
The test case `~/project and ~/project/ should not be added separately to the known projects list` (L140) is kinda brittle, I had to change the test folder name from `~/a/` to `~/project/` otherwise the test would fail whenever the prior test on L136 was run as well. Let me know if this sounds like a problem 😃 

-----------------

- [x The commits are consistent with our [contribution guidelines](../CONTRIBUTING.md)

Thanks!


bbatsov:
Thanks!

PR #1316, opened 2018-10-05
Correctly set the dir-local variable indent-tabs-mode to nil

re-mx:
Disable `indent-tabs-mode` correctly.

The previous settings will turn on `indent-tabs-mode`.

-----------------

- [ ] ~~You've updated the changelog (if adding/changing user-visible functionality)~~

Thanks!


bbatsov:
Are you sure about this? I think that binding the var like it was done is exactly the same.

re-mx:
I am not sure if there is an automated way to prove it.

But you can find a file to try. Write a few new lines of code to see if it uses TAB indentation.

bbatsov:
Well, this can't hurt anyways, and explicit is better than implicit. :-)

PR #1317, opened 2018-10-06
Fix project type detection order

### LINKED ISSUE #1315: fix project type detection order
```
;;; Project type registration
;;
;; Project type detection happens in a reverse order with respect to
;; project type registration (invocations of `projectile-register-project-type').
```
project type detection happens in a reverse order in the source code.
But actually, the detection method seems to be positive.
I tried the following code to try to override `make project` registration in the source code.
```
(projectile-register-project-type 'make-override '("Makefile"))
```
Calling `projectile-project-info` in a test project, Still showing `make project`.

yydcnjjw:
fix project type detection order
#1315
-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2457,6 +2457,7 @@ Fallsback to a generic project type when the type can't be determined."
                                    (and (projectile-verify-files marker) project-type)
                                  (and (funcall marker) project-type))))
                            (hash-table-keys projectile-project-types))
+                          :from-end t
                           'generic)))
     (puthash (projectile-project-root) project-type projectile-project-type-cache)
     project-type))


bbatsov:
I'm not sure if that's a good approach, as we're simply getting some hash-table-keys here and I doubt there's a guarantee about the order of those keys. Maybe we shouldn't be using a hash to store the project types to begin with?

bbatsov:
Yeah, looking at the code I think that we should simply stuff the project types in an alist, not a hashtable and this will work out pretty well.

bbatsov:
I've made the necessary change. Try out the latest version.

PR #1325, opened 2018-10-13
fix for projectile-project-files relative path bug

### LINKED ISSUE #1324: Open File creates a new file
## Expected behavior
We have source files into `src/` directory (e.g. `src/main.cpp`).

Launch the command `projectile-find-file` and type `main`.
The file `main.cpp` should be open.

## Actual behavior
A new file (empty buffer) will be created instead. The `pwd` is the "project" directory (`.`).

### Project Tree
~~~
.
├── .projectile
├── .gitignore  
├── CMakeLists.txt
├── include
│   └── Foo.hpp
└── src
    ├── Foo.cpp
    └── main.cpp
~~~

### `.projectile` content
~~~
+/src
~~~

### Emacs Configuration
~~~
;; == Package Managment ==
(require 'package)
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/"))
(package-initialize)

;; == Install UsePackage ==
(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))

;; == Projectile ==
(use-package projectile
  :ensure t
  :config
  (define-key projectile-mode-map (kbd "C-c p") 'projectile-command-map)
  (projectile-mode +1))
~~~

```
Projectile version: 20181009.851
```

### Emacs version

GNU Emacs 25.2.2

### Operating systems

* Ubuntu 18.04
* ArchLinux

aiba:
Fixes #1324

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1756,7 +1756,11 @@ https://github.com/abo-abo/swiper")))
       (when projectile-enable-caching
         (message "Projectile is initializing cache..."))
       (setq files (cl-mapcan
-                   (lambda (dir) (projectile-dir-files dir))
+                   (lambda (dir)
+                     (mapcar (lambda (f)
+                               (file-relative-name (concat dir f)
+                                                   project-root))
+                             (projectile-dir-files dir)))
                    (projectile-get-project-directories project-root)))
 
       ;; Save the cached list.


bbatsov:
That's going to fix the problem, but it's also going to kill the performance as you'll be iterating over each file in Emacs Lisp which is very slow on big projects. (that's also the main reason I did the change that broke this). The problem is that this expansion is not needed at all for most project (unless someone has defined explicitly some dirs in `.projectile`, so I definitely don't think this should be done for every project. Maybe we can check if there are some subdirs defined prior to expanding those?

bbatsov:
Thinking a bit more about this, the problem is mostly that we currently just run the command like `git ls-files` without passing the dir as a param, which results in the paths relative to the current dir, not the project dir. Probably we should be running all such commands from the project root dir and pass the sub-dir explicitly (or something along those lines).

aiba:
Yeah that makes total sense.  This fix is fast enough for my projects, so I thought I'd share it, but I won't be offended if you'd rather fix it in a way that's more performant or architecturally consistent with the rest of the code.

bbatsov:
I don't have much time for this, so I'll just merge your fix and add some note for me to speed things up later. 

I'll also disable completely .projectile config for `'turbo-alien` to offset the effect of the path expansion.

bbatsov:
Thanks for working on this!

PR #1327, opened 2018-10-16
Add a command to re-execute the last external command in a project.

### LINKED ISSUE #1279: `recompile' functionality
I would love to have functionality in Projectile which mimics "recompile". That is, it runs what ever was the last project command -- configure, test, build etc.

This fits with my usage -- I normally have a single build command which I repeat a lot till it's all working, then I move onto the next step (whatever that might be).

mookid:
Here external command = caller of `projectile--run-project-cmd`.

Implemented is a non-prompting version that repeats the last command run, merging test, compile, configure etc commands.

Fix #1279

cc @phillord 

To be discussed: can anything interesting be done with the command history? maybe a prompting version could be interesting.

before merging:
- [x] update the changelog 
- [X] prompting version

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3442,6 +3442,17 @@ project of that type"
       (funcall cmd)
     (compile cmd)))
 
+(defvar projectile-project-command-history (make-hash-table :test 'equal)
+  "The history of last executed project commands, per project.
+
+Projects are indexed by their project-root value.")
+
+(defun projectile--get-command-history (project-root)
+  (or (gethash project-root projectile-project-command-history)
+      (puthash project-root
+               (make-ring 16)
+               projectile-project-command-history)))
+
 (cl-defun projectile--run-project-cmd
     (command command-map &key show-prompt prompt-prefix save-buffers)
   "Run a project COMMAND, typically a test- or compile command.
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -3429,7 +3442,9 @@ running the command."
          (command (projectile-maybe-read-command show-prompt
                                                  command
                                                  prompt-prefix)))
-    (puthash default-directory command command-map)
+    (when command-map
+      (puthash default-directory command command-map)
+      (ring-insert (projectile--get-command-history default-directory) command))
```
phillord: Why is this not `projectile-project-root' rather than `default-directory'

phillord: This is the right idea but needs to catch some errors -- if there is not last, it crashes.

```
  string-match("\\`\\s *cd\\(?:\\s +\\(\\S +?\\|'[^']*'\\|\"\\(?:[^\"`$\\]\\|\\\\.\\)*\"\\)\\)?\\s *[;&\n]" nil)
  compilation-start(nil nil)
  compile(nil)
  (if (functionp cmd) (funcall cmd) (compile cmd))
  projectile-run-compilation(nil)
```

Also, it's seems to work if I am in a *compilation* buffer, but not elsewhere in the project. Would it not make more sense to hash by the project root than default directory?

mookid:
@phillord I am not sure I understand what you mean concerning the compilation buffer, but I think that I addressed the robustness issue.

phillord:
Only last thing I can suggest is a menu item to aid discoverability.

But, as it stands this is exactly what I wanted. Thank you for spending the time on it!

mookid:
> Only last thing I can suggest is a menu item to aid discoverability.

done!

projectile.el
```diff
@@ -3496,6 +3512,31 @@ with a prefix ARG."
...
+(defun projectile-repeat-last-command (show-prompt)
+  "Run last projectile external command.
+
+External commands are: `projectile-configure-project',
+`projectile-compile-project', `projectile-test-project' and
+`projectile-run-project'.
+
+If the prefix argument SHOW_PROMPT is non nil, the command can be edited."
+  (interactive "P")
+  (let* ((project-root (projectile-project-root))
```
bbatsov: You need to wrap the call to `projectile-project-root` in `projectile-ensure-project`, otherwise you can get a nil outside a project.

bbatsov:
Apart from my small remark - great work!

bbatsov:
You'll also have to rebase on top of the current `master`.

bbatsov:
Thanks!

Issue #1323, opened 2018-10-10
Error switching to project containing nested git repos


countvajhula:
## Expected behavior
It should be possible to switch to a project containing nested git repos that are not submodules (e.g. it should treat the inner repos the same as any other subfolder within a project).

_Example_: My `dotfiles` repo has this structure due to installed vim packages being nested git repos.

## Actual behavior
When attempting to switch to such a project, it fails and the following error is seen:
```
apply: Setting current directory: No such file or directory, /Users/siddhartha/work/sandbox/testtest/my-project/fatal: no submodule mapping found in .gitmodules for path 'inner-project'
inner-project/
```

## Steps to reproduce the problem

1. In a folder for testing, perform the following steps (or just paste this code into the shell):
```
mkdir my-project
cd my-project
mkdir inner-project
cd inner-project
git init
touch file1.txt
git add file1.txt
git commit -m "a file"
cd ..
git init
touch project-file.txt
git add project-file.txt
git commit -m "a file"
git add -A
git commit -m "add inner repo"
```
2. Open a file in the project (i.e. `my-project/project-file.txt`) using emacs
3. Switch to another project
4. Attempt to switch back to the `my-project` project. It fails with the above error.

### Projectile version information

```
Projectile 20181009.851
```

**Note: the issue is not present in `Projectile 1.0.0`**

### Emacs version

*GNU Emacs 26.1 (build 1, x86_64-apple-darwin17.6.0, Carbon Version 158 AppKit 1561.4)
 of 2018-09-10*

### Operating system

*MacOS High Sierra 10.13.5*

bbatsov:
Likely I've broken something with my recent refactorings. I'm a bit busy the next couple of weeks, but you can help me out if you'd like to debug where exactly the problem is. See http://www.projectile.mx/en/latest/troubleshooting/ for details.

enordstrm:
I had a similar issue, which might be related, where I have a GIT repository with multiple sub modules and within that GIT repository I've marked one specific sub directory as a project.  Projectile, after the recent ranges, want the sub module paths relative to the project directory but the default sub module command set gets them relative to the GIT top level directory which causes an issue.

Workaround/solution to that was to change the sub module command.

```elisp
(setq projectile-git-submodule-command "git submodule --quiet foreach 'echo $(git rev-parse --show-toplevel)' | xargs realpath --relative-to=$PWD"))
```

troglotit:
[...]

PR #1330, opened 2018-10-17
Disregard STDERR in results from executing shell commands

countvajhula:
# Background

This is a proposed fix for the issue described in https://github.com/bbatsov/projectile/issues/1323

In the case where a project contains nested git repos that are not submodules, projectile commands do not work and give output like:

```
apply: Setting current directory: No such file or directory, /Users/siddhartha/work/sandbox/testtest/my-project/fatal: no submodule mapping found in .gitmodules for path 'inner-project'
inner-project/
```

What is happening here is that at some point, projectile runs the command `git submodule --quiet foreach 'echo $path' | tr '\\n' '\\0'` to get paths to subprojects within the project. When this happens, in the case of nested repos mentioned above:

1. git yields the following output `Result: "fatal: no submodule mapping found in .gitmodules for path 'inner-project'\ninner-project�"` -- _Note that the part after the `\n` is the legitimate output expected here._

2. Projectile concatenates this (whole) string with the base project root to derive the full path to the subproject, resulting in a junk path

3. As it is treating this error output as a path, projectile then calls the same command on that new folder to get subprojects recursively which then yields `File is missing: "Setting current directory", "No such file or directory", "/Users/siddhartha/work/sandbox/my-project/fatal: no submodule mapping found in .gitmodules for path 'inner-project'`

... which causes the operation to fail.

# Solution

In running shell commands, the built-in function `shell-command-to-string` combines STDOUT with STDERR. For the purposes of using this output programmatically we only need STDOUT, since STDERR contains error output rather than the legitimate results of command execution. This PR introduces a modified version of the built-in function as suggested [here](https://emacs.stackexchange.com/questions/21422/how-to-discard-stderr-when-running-a-shell-function). Relevant excerpt from the Emacs docs regarding this change:

```
DESTINATION can also have the form (REAL-BUFFER STDERR-FILE); in that case,
REAL-BUFFER says what to do with standard output, as above,
while STDERR-FILE says what to do with standard error in the child.
STDERR-FILE may be nil (discard standard error output),
t (mix it with ordinary output), or a file name string.
```

**Other changes included:**
- rename the function `projectile-files-via-ext-command` → `projectile-get-results-from-shell-command`
- migrate tests for this function from ERT to buttercup
- add a couple more tests for it

-----------------

~- [ ] You've updated the readme (if adding/changing user-visible functionality) [N/A]~

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 * Make the mode line configurable via `projectile-dynamic-mode-line` and `projectile-mode-line-function`.
 * [#1205](https://github.com/bbatsov/projectile/issues/1205): Check that project directory exists when switching projects.
 * Move Projectile's menu out of the "Tools" menu.
+* Ignore STDERR when running shell commands so it doesn't pollute usable output
 
 ### Bugs fixed
 
--- a/projectile.el
+++ b/projectile.el
@@ -1147,9 +1147,9 @@ function is executing."
   (let ((vcs (projectile-project-vcs directory)))
     (cond
     ((eq vcs 'git)
-     (nconc (projectile-files-via-ext-command directory (projectile-get-ext-command vcs))
+     (nconc (projectile-get-results-from-shell-command directory (projectile-get-ext-command vcs))
             (projectile-get-sub-projects-files directory vcs)))
-    (t (projectile-files-via-ext-command directory (projectile-get-ext-command vcs))))))
+    (t (projectile-get-results-from-shell-command directory (projectile-get-ext-command vcs))))))
 
 (define-obsolete-variable-alias 'projectile-get-repo-files 'projectile-dir-files-external "1.1")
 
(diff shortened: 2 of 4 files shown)


countvajhula:
Hello there, this is ready for review. I renamed `projectile-files-via-ext-command` to `projectile-get-results-from-shell-command` in writing tests for it, since it didn't appear to be specific to files... Let me know if that doesn't work for some reason, and thanks for making a great tool!

bbatsov:
I agree with need to do something about this, but I'm not sure that's the best approach. Shouldn't we simply fail if the underlying shell commands fails? That seems more appropriate to me. I'm afraid that now we'll make harder to debug the cause of certain problems.

countvajhula:
Makes sense, I agree silent failure would be bad. At the same time, having output in STDERR does not necessarily mean that the command failed, and we would like to succeed in such cases where there is usable output. For example, `$ ls <non-existent folder> <actual folder>` in a shell gives output in both STDOUT as well as STDERR. The issue https://github.com/bbatsov/projectile/issues/1323 is similar to this case.

One possibility is to log STDERR to the `*Messages*` buffer instead of just discarding it. The emacs docs have this to say here:
```
You can't directly specify a buffer to put the error output in;
that is too difficult to implement. But you can achieve this
result by sending the error output to a temporary file and then
inserting the file into a buffer when the subprocess finishes.
```
From: https://www.gnu.org/software/emacs/manual/html_node/elisp/Synchronous-Processes.html

Does that seem like a reasonable solution? I'm not too familiar on emacs conventions around temporary files (e.g. where to create them / whether they should be created at all). WDYT?

Issue #1158, opened 2017-08-08
projectile-find-file-in-directory issues


blaenk:
1. If the operation is canceled with C-g, `projectile-completing-read` yields `nil` which is then fed to `expand-file-name` which then emits a wrong type argument error.
2. The illusion of "being in" the target directory is broken by `projectile-cached-project-root`. If I call `projectile-find-file-in-directory` with some other projectile project A as the target directory, _from_ yet another projectile B directory, the illusion breaks down because asking for the projectile root yields B.

I was able to fix these things externally like so:

``` elisp
(defun my-edit-dots ()
  (interactive)
  (ignore-errors
    (let* ((target (getenv "DOTSPATH"))
           (projectile-cached-project-root target))
      (projectile-find-file-in-directory target))))
```

bbatsov:
> The illusion of "being in" the target directory is broken by projectile-cached-project-root. If I call projectile-find-file-in-directory with some other projectile project A as the target directory, from yet another projectile B directory, the illusion breaks down because asking for the projectile root yields B.

This was removed in Projectile 1.1, btw. 

> If the operation is canceled with C-g, projectile-completing-read yields nil which is then fed to expand-file-name which then emits a wrong type argument error.

Yep, I guess we need to add some nil check there.

PR #1332, opened 2018-10-20
Check directory arguments passed interactively.

mookid:
projectile-completing-read and projectile-dir-files: check that argument is a directory, otherwise early exit. 

projectile--directory-p: new function (a robust version of file-directory-p).

### DIFF
--- a/test/projectile-test.el
+++ b/test/projectile-test.el
@@ -246,7 +246,13 @@ test temp directory"
     (expect (projectile-get-project-directories "/my/root/") :to-equal '("/my/root/foo" "/my/root/bar/baz"))))
 
 (describe "projectile-dir-files"
+  (it "fails unless directory exists"
+    (spy-on 'file-directory-p :and-call-fake
+            (lambda (filename) (equal filename "/my/root/")))
+    (expect (projectile-dir-files "asdf") :to-throw))
   (it "lists the files in directory and sub-directories"
+    (spy-on 'file-directory-p :and-call-fake
+            (lambda (filename) (equal filename "/my/root/")))
     (spy-on 'projectile-patterns-to-ignore)
     (spy-on 'projectile-index-directory :and-call-fake (lambda (dir patterns progress-reporter)
                                                          (expect dir :to-equal "/my/root/")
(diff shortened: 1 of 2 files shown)


bbatsov:
So, the purpose of the helper is mostly to avoid a NPE you can get if you the raw Emacs function?

mookid:
Exactly! it's only to fail gracefully in #1158.

bbatsov:
Got it. Well, it'd be nice if you added a couple of tests for the two commands this is affecting that verify errors are raised accordingly. One of the big outstanding tasks for Projectile is to backfill the test suite, as much of the functionality we've got has no tests at all.

test/projectile-test.el
```diff
@@ -861,3 +861,30 @@ test temp directory"
...
+    (spy-on 'file-directory-p :and-call-fake
+            (lambda (filename) (equal filename "/my/root/")))
+    (spy-on 'projectile-patterns-to-ignore :and-return-value nil)
+    (spy-on 'projectile-index-directory :and-call-fake
+            (lambda (dir patterns progress-reporter)
+              (setq indexed-dir dir)
+              '("/my/root/a/b/c" "/my/root/a/d/e")))
+    (spy-on 'projectile-dir-files-alien :and-return-value '("a/b/c" "a/d/e"))
+    (spy-on 'cd :and-return-value nil)
+    (expect (projectile-dir-files "/my/root/") :to-equal '("a/b/c" "a/d/e"))
```
mookid: TODO: call with `projectile-indexing-method` equal to multiple values like in the original test

test/projectile-test.el
```diff
@@ -123,12 +123,15 @@ test temp directory"
...
   (it "lists the files in directory and sub-directories"
+    (spy-on 'file-directory-p :and-call-fake
+            (lambda (filename) (equal filename "/my/root/")))
     (spy-on 'projectile-patterns-to-ignore)
     (spy-on 'projectile-index-directory :and-call-fake (lambda (dir patterns progress-reporter)
                                                          (expect dir :to-equal "/my/root/")
                                                          '("/my/root/a/b/c" "/my/root/a/d/e")))
     (spy-on 'projectile-dir-files-alien :and-return-value '("a/b/c" "a/d/e"))
     (spy-on 'cd)
+    (expect (projectile-dir-files "asdf") :to-throw)
```
bbatsov: This should be a separate example (`it`).

bbatsov:
Thanks!

PR #1333, opened 2018-10-20
Add a new defcustom to configure projectile-kill-buffers.

### LINKED ISSUE #1233: Killing all project buffers also kills non-file buffers
## Expected behavior

Using projectile-kill-buffers should not also take out non-file buffers, such as a REPL that is meant to exist for all projects.

## Actual behavior

Using projectile-kill-buffers deletes the SLIME or Sly REPL buffers with Common Lisp projects.

Open a project file, then open SLIME or Sly. Issue projectile-kill-buffers and both are deleted.
This does not occur if SLIME or Sly is started before visiting a project file.

```
Projectile version: MELPA projectile 20180324.2226
```

### Emacs version

25.3

### Operating system

Arch Linux

mookid:
~~3~~ 2 options are implemented:
- all-buffers (corresponds to the existing behavior)
~~keep-processes~~
- kill-only-files: kill only the buffer that have a buffer-file-name.

changelog:
* projectile-kill-buffers-filter: new defcustom.

------------- 
Note:
- option 2 corresponds to the use-case described in #1233  
- maybe option kill-only-files is not useful; and maybe other choices are. Suggestions are welcome!

edit: eventually kept only `kill-only-files`, while allowing any predicate to be used.

cc @mfiano 
fix #1233

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -118,6 +118,22 @@ using the native indexing method."
   :group 'projectile
   :type 'boolean)
 
+(defcustom projectile-kill-buffers-filter 'kill-all
+  "Determine which buffers are killed by `projectile-kill-buffers'.
+
+When the kill-all option is selected, kills each buffer.
+
+When the keep-processes option is selected, keep buffers associated to a
+process.
+
+When the kill-only-files option is selected, kill only the buffer
+associated to a file."
+  :group 'projectile
+  :type '(radio
+          (const :tag "All project files" kill-all)
+          (const :tag "Preserve process buffer" keep-processes)
+          (const :tag "Project file buffers" kill-only-files)))
+
 (defcustom projectile-file-exists-local-cache-expire nil
   "Number of seconds before the local file existence cache expires.
 Local refers to a file on a local file system.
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -118,6 +118,22 @@ using the native indexing method."
...
+When the kill-all option is selected, kills each buffer.
+
+When the keep-processes option is selected, keep buffers associated to a
+process.
+
+When the kill-only-files option is selected, kill only the buffer
+associated to a file."
+  :group 'projectile
+  :type '(radio
+          (const :tag "All project files" kill-all)
```
bbatsov: files -> buffers

projectile.el
```diff
@@ -118,6 +118,22 @@ using the native indexing method."
...
+          (const :tag "Preserve process buffer" keep-processes)
```
bbatsov: I don't get what's the difference between the final two options - keep-processes and and kill-only-files are more or less the same as something's either a file buffer or some special buffer (although those won't necessary have process attached to them).

projectile.el
```diff
@@ -3130,17 +3146,28 @@ to run the replacement."
...
-        (mapc #'kill-buffer (cl-remove-if 'buffer-base-buffer buffers)))))
+    (when (yes-or-no-p
+           (format "Are you sure you want to kill buffers for '%s'? "
+                   project-name))
+      (dolist (buffer buffers)
+        (when (and
+               ;; we take care not to kill indirect buffers directly
+               ;; as we might encounter them after their base buffers are killed
+               (not (buffer-base-buffer buffer))
+               (case projectile-kill-buffers-filter
```
bbatsov: You probably meant to use `pcase` here.

projectile.el
mookid: You are probably right that the difference may not make much sense in terms of usage. I made the customization point extensible to allow for more fine-grained control, while removing the keep-processes choice.

projectile.el
mookid: done!

mfiano:
Thanks so much for this! Can't wait for it to be merged

bbatsov:
Thanks!

PR #1346, opened 2018-11-06
Bring back buffer kill count message

nickdrozd:
This was removed in 90997d6, but I find it handy as a warning that I
might be killing the wrong project. If I mean to kill one with only a
few buffers and I see the message "kill 20 buffers", I'll be saved
from making a very annoying mistake.

Plural/singular handling probably doesn't matter much. Personally, I
would never use this command for a project with a single buffer -- I
would just use `kill-this-buffer`.

-----------------

Thanks!


bbatsov:
👍

PR #1348, opened 2018-11-09
WIP Add projectile-tags-exclude-supports-globs

cybergrind:
Hey. I'm using https://github.com/universal-ctags/ctags as my default ctags implementation and it seems that --excludes options doesn't work as intended. It performs 2 kinds of matches

```
For each file name considered by ctags, each pattern specified using this option will
be compared against both the complete path (e.g.   some/path/base.ext)  and  the  base  name  (e.g.
base.ext)  of  the  file,  thus allowing patterns which match a given file name irrespective of its
path, or match only a specific path. If appropriate support is available from the  runtime  library
of  your  C  compiler, then the pattern may contain the usual shell wildcards (not regular expressions)
common on Unix (be sure to quote the option parameter to protect the wildcards from being  expanded
by the shell before being passed to ctags; also be aware that wildcards can match the slash charac‐
ter, '/').
```

What does it mean practically for non-file match: it compares absolute path with relative path passed from `projectile-ignored-directories-rel` and doesn't perform partial match in the middle of string but works well for top-level directories (because of the second, basename, check)

`/home/username/project_root/src/ignored_dir` doesn't match to `src/ignored_dir`

But we can pass globs in such case, so if we check for `*/src/ignored_dir` it will match.

I'm not sure what is the best possible implementation for this functionality. I've started with an additional option that additionally passes all non-matched to directory/file strings to exclude function.

If this way isn't good, I'm thinking about extending `projectile-parse-dirconfig-file` to support additional modifier that will fill some additional variable (e.g. `*` for ctags globs).

Or maybe you suggest a better way to implement this functionality.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -234,6 +234,11 @@ set to 'ggtags', then ggtags will be used for
           (const :tag "standard" find-tag))
   :package-version '(projectile . "0.14.0"))
 
+(defcustom projectile-tags-exclude-supports-globs nil
+  "If t - treat all non-matched ignores from .projectile as globs"
+  :group 'projectile
+  :type 'boolean)
+
 (defcustom projectile-sort-order 'default
   "The sort order used for a project's files."
   :group 'projectile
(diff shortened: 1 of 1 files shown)


cybergrind:
Hacked it locally with code below, don't think that it worth to add this feature.

```elisp
(use-package projectile
  :config
  (defun projectile-tags-exclude-patterns ()
    "Return a string with exclude patterns for ctags. Add support for wildcards passthrough"
    (mapconcat (lambda (pattern) (format "--exclude=\"%s\""
               (append (projectile-paths-to-ignore) projectile-globally-ignored-directories) " ")
  )
)
```

PR #1350, opened 2018-11-16
Don't cache files with ignored suffixes

ericdanan:
In `projectile-ignored-file-p`, check that the file doesn't match against `projectile-ignored-file-suffixes`.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1551,10 +1551,14 @@ Regular expressions can be used."
   "Check if FILE should be ignored.
 
 Regular expressions can be used."
-  (cl-some
-   (lambda (name)
-     (string-match-p name file))
-   (projectile-ignored-files)))
+   (or (cl-some
+        (lambda (suf)
+          (string-suffix-p suf file t))
+        projectile-globally-ignored-file-suffixes)
+       (cl-some
+        (lambda (name)
+          (string-match-p name file))
+        (projectile-ignored-files))
 
 (defun projectile-check-pattern-p (file pattern)
   "Check if FILE meets PATTERN."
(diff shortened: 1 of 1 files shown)


ericdanan:
Sorry, I had to fix the commit by restoring some parentheses that mysteriously disappeared.

projectile.el
```diff
@@ -1551,10 +1551,14 @@ Regular expressions can be used."
...
 
 Regular expressions can be used."
-  (cl-some
-   (lambda (name)
-     (string-match-p name file))
-   (projectile-ignored-files)))
+   (or (cl-some
+        (lambda (suf)
+          (string-suffix-p suf file t))
+        projectile-globally-ignored-file-suffixes)
```
bbatsov: Wasn't this supposed to be handled by `projectile-ignored-files`?

ericdanan: It isn't, `projectile-ignored-files` only handles `projectile-globally-ignored-files`.

ericdanan:
I am still interested in this PR. Please let me know if you would like any clarification or modification.

projectile.el
bbatsov: I think we should just put the handling of the suffixes there as well. Otherwise it becomes a big weird to have external handling for this.

ericdanan: I am not sure how to do this efficiently, because `projectile-ignored-files` merely lists entries the `.projectile` file and `projectile-globally-ignored-files` variable. If we want it to list all project files with extensions in `projectile-globally-ignored-file-suffixes` as well, we would first need to recompute the list of project files. But this seems overkill to me. Do you have any suggestion on how to proceed?

bbatsov: Well, we can always expand the suffixes as `**/*.suffix` patterns in the context of the broader function. If this wildcard expansion doesn't work out well I guess we should go with your suggestion. 

Still, the changes will have to be covered by some tests.

PR #1354, opened 2018-11-22
Alien indexing submodules should return relative paths

### LINKED ISSUE #1353: File paths to submodules are not relative to project root
## Expected behavior
projectile-find-files should be able to find files in submodules relative to project root

## Actual behavior
all files in submodules are returned with absolute path, while files in project root (non submodule files) are still returned as relative paths
```
mkdir test && cd test
 git init
 git submodule add  https://github.com/bbatsov/projectile.git
 touch test.txt

```
at this point, navigate to the `test` project and try to find files with projectile, you'll see test.txt in the search candidates as relative path, but all files in the submodule are in absolute paths.

This also breaks hybrid method because ignore patterns in .projectile won't match these absolute paths any more.

Projectile version: 20181106.1631

### Emacs version
27.0.50

### Operating system
MacOS 10.14

fantasticsid:
- When using alien indexing method projectile currently returns the
  full paths to files in submodules, this makes filtering in hybrid
  mode to break because the ignore patterns would be matched against
  the absolute path to submodules files, instead of relative to
  project root

In the projectile 1.0.0 release, files are searched and then processed to make them relative to project root:
https://github.com/bbatsov/projectile/blob/09d1ef17a20c42dc6a2b1622df8faa8fb1c6ad9f/projectile.el#L990

But this has since been refactored and was broken.

This PR fixes #1353
-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1261,9 +1261,11 @@ they are excluded from the results of this function."
   "Get files from sub-projects for PROJECT-ROOT recursively."
   (projectile-flatten
    (mapcar (lambda (sub-project)
-             (let ((project-relative-path (file-relative-name sub-project project-root)))
+             (let ((project-relative-path
+                    (file-name-as-directory (file-relative-name
+                                             sub-project project-root))))
                (mapcar (lambda (file)
-                       (concat (file-name-as-directory project-relative-path) file))
+                       (concat project-relative-path file))
                      ;; TODO: Seems we forgot git hardcoded here
                      (projectile-files-via-ext-command sub-project projectile-git-command))))
            (projectile-get-all-sub-projects project-root))))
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -1261,10 +1261,11 @@ they are excluded from the results of this function."
   "Get files from sub-projects for PROJECT-ROOT recursively."
   (projectile-flatten
    (mapcar (lambda (sub-project)
-             (mapcar (lambda (file)
-                       (concat sub-project file))
+             (let ((project-relative-path (file-relative-name sub-project project-root)))
+               (mapcar (lambda (file)
+                       (concat (file-name-as-directory project-relative-path) file))
```
bbatsov: I think you should move the `file-name-as-directory` to the `let`. Seems wasteful to call this within the `mapcar`.

fantasticsid: fixed in 748ed2b

fantasticsid:
Hey @bbatsov , I've pushed some updates to this PR. Could you take another look? Thanks!

bbatsov:
Thanks for working on this! I really appreciate it! 🙇

PR #1357, opened 2018-12-03
Port advices to new mechanism

Silex:
**don't merge yet**

-----

Projectile uses this defadvice for `delete-file`:

``` elisp
(defadvice delete-file (before purge-from-projectile-cache (filename &optional trash))
  (if (and projectile-enable-caching (projectile-project-p))
      (let* ((project-root (projectile-project-root))
             (true-filename (file-truename filename))
             (relative-filename (file-relative-name true-filename project-root)))
        (if (projectile-file-cached-p relative-filename project-root)
            (projectile-purge-file-from-cache relative-filename)))))
```

It does the right thing, except it's very annoying when you delete a huge directory through `dired` for example, because it slows down the deletion by an order of magnitude.

There are several options around this problem:

- Add ability to disable the above advice, and let the user deal with cache invalidation.
- Advice `dired-do-delete` so it disables the `delete-file` advice, perform its task then enable the advice and finally invalidate the cache.

Additionnaly, should the old `defadvice` be ported to `advice-add` & friends?

Anyway, tell me what you reckon is the way forward, I might make a PR.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -960,7 +960,7 @@ Invoked automatically when `projectile-mode' is enabled."
   (mapcar #'projectile-discover-projects-in-directory projectile-project-search-path))
 
 
-(defadvice delete-file (before purge-from-projectile-cache (filename &optional trash))
+(defun delete-file-projectile-remove-from-cache (filename &optional trash)
   (if (and projectile-enable-caching (projectile-project-p))
       (let* ((project-root (projectile-project-root))
              (true-filename (file-truename filename))
@@ -3577,12 +3577,11 @@ If the prefix argument SHOW_PROMPT is non nil, the command can be edited."
     (unless (string= command executed-command)
       (ring-insert command-history executed-command))))
 
-(defadvice compilation-find-file (around projectile-compilation-find-file)
+(defun compilation-find-file-projectile-find-compilation-buffer (orig-fun marker filename directory &rest formats)
   "Try to find a buffer for FILENAME, if we cannot find it,
 fallback to the original function."
-  (let ((filename (ad-get-arg 1))
-        full-filename)
-    (ad-set-arg 1
+  (let (full-filename)
+    (setq filename
                 (or
                  (if (file-exists-p (expand-file-name filename))
                      filename)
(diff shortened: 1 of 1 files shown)


Silex:
Okay I started to move toward the right direction, can someone help me rewrite https://github.com/Silex/projectile/blob/improve-advices/projectile.el#L3583-L3601 in a better way? There has to be something more simple/readable.

projectile.el
```diff
@@ -3577,32 +3577,25 @@ If the prefix argument SHOW_PROMPT is non nil, the command can be edited."
     (unless (string= command executed-command)
       (ring-insert command-history executed-command))))
 
-(defadvice compilation-find-file (around projectile-compilation-find-file)
+(defun compilation-find-file-projectile-find-compilation-buffer (marker filename directory &rest formats)
```
mookid: the newly defined function should take the original definition as first argument; you should funcall this argument instead of compilation-find-file inside the body.

also, please keep the refactoring minimal (or split into commits) to ease review.

thanks!

Silex: @mookid: right, will do. Thanks

Silex:
Alright, please comment on the `[MAYBE?]` commit, the rest should be ok.

bbatsov:
LGTM.

Silex:
Alright I renamed the `[MAYBE]` refactor commit. I didn't really test those changes.

bbatsov:
I guess someone will test them pretty soon. 😉 Thanks for working on this!

Silex:
:wink: :+1:

projectile.el
```diff
@@ -3577,32 +3582,25 @@ If the prefix argument SHOW_PROMPT is non nil, the command can be edited."
...
+                               (mapcar
+                                (lambda (f)
+                                  (expand-file-name
+                                   filename
+                                   (expand-file-name f root)))
+                                dirs)))))
+      (when new-filename
+        (setq filename new-filename))))
+
+  (funcall #'orig-fun marker filename directory formats))
```
mookid: this is wrong: you are trying to call a function *named* "orig-fun":
```
ELISP> (defun my-advice (orig-fun &rest args) (funcall #'orig-fun args))
my-advice
ELISP> (defun foo () 1)
foo
ELISP> (advice-add 'foo :around #'my-advice)
nil
ELISP> (foo)
*** Eval error ***  Symbol’s function definition is void: orig-fun
```

PR #1360, opened 2018-12-06
Show path to cache initialization

gizmomogwai:


projectile.el
```diff
@@ -1786,7 +1786,7 @@ https://github.com/abo-abo/swiper")))
     ;; Calculate the list of files.
     (when (null files)
       (when projectile-enable-caching
-        (message "Projectile is initializing cache..."))
+        (message (format "Projectile is initializing cache in %s ..." (projectile-project-root))))
```
bbatsov: `format` is not needed here, as `message` also supports formatting options. `in` should probably be `for` and `(projectile-project-root)` can simply be replaced with a reference to the `project-root` param.

gizmomogwai:
thanks for the improvements.

gizmomogwai:
could you merge this, or should i rebase or abandon the commit?

bbatsov:
Merge it is. Thanks!

PR #1363, opened 2018-12-17
Fix bug in compilation-find-file-projectile-find-compilation-buffer

Silex:
@bbatsov oops :sweat_smile: 

Thanks to @mookid for noticing!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3603,7 +3603,7 @@ fallback to the original function."
       (when new-filename
         (setq filename new-filename))))
 
-  (funcall #'orig-fun marker filename directory formats))
+  (apply orig-fun `(,marker ,filename ,directory ,@formats)))
 
 (defun projectile-open-projects ()
   "Return a list of all open projects.


Silex:
@mookid: I had to move from e924ecf to ea3eb81 because of the `formats` variable, because otherwise it'd do `(compilation-find-file marker filename directory '(nil))`.

Do you see a better way to write this?

mookid:
+ note that instead of (car (remove-if-not ...   you can use find-if.
+ otherwise, nothing to say about your code. it's a bit ugly because the function has a bunch of arguments, but as long as you tested it that's ok.  Maybe you can try to catch the current bug in a unit test?

kaushalmodi:
@bbatsov Please merge this ASAP.

Silex:
@bbatsov: just a thought, maybe consider giving rights to merge to the repository for trusted people of your choice for situations like this? Then people would have a list of contributors to summon in case of urgencies (I can volunteer).

Silex:
:+1:

PR #1367, opened 2019-01-05
Fix the Makefile so that we can compile projectile

arichiardi:
The functionality has been basically copied from cider and now we can call make
in order to produce projectile.elc.

-----------------


bbatsov:
Thanks!

PR #1370, opened 2019-01-11
Fix get-immediate-sub-projects when argument is not the root

### LINKED ISSUE #1369: projectile-get-immediate-sub-projects won't work with non-root argument
## Expected behavior

```emacs
(projectile-get-immediate-sub-projects "/project/a-sub-dir/")
```
should return a list of `/project`'s sub-modules that are also within `a-sub-dir/`, `nil` if none.

## Actual behavior

Returns non-existing paths, where each sub-module path is appended to `/a-sub-dir`.

1. Create a new git project in `/project`
2. Add a git-submodule "foo" to the project under `/project/lib/foo`
3. Add a folder to the project under `/project/a-sub-dir`
4. Execute the code above

You get `("/project/a-sub-dir/lib/foo") which is a non-existing path. Instead, you wanted `nil` because there is no submodule below `/project/a-sub-dir`.

```
Projectile 2.0.0
```

(I'm following master)

### Emacs version

GNU Emacs 26.1.90 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.1)

### Operating system

Fedora 28

DamienCassou:
This fixes #1369.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+* [#1369](https://github.com/bbatsov/projectile/issues/1369): Fix projectile-get-immediate-sub-projects when passed a non-root argument
+
 ## 2.0.0 (2019-01-01)
 
 ### New features
--- a/projectile.el
+++ b/projectile.el
@@ -1241,13 +1241,14 @@ PATH is the vcs root or project root from which to start
 searching, and should end with an appropriate path delimiter, such as
 '/' or a '\\'.
 
-If the vcs get-sub-projects query returns results outside of path,
+If the vcs get-sub-projects query returns results outside of PATH,
 they are excluded from the results of this function."
-  (let* ((vcs (projectile-project-vcs path))
+  (let* ((root (projectile-project-root path))
+         (vcs (projectile-project-vcs path))
          ;; search for sub-projects under current project `project'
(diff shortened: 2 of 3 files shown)


DamienCassou:
I'm having a look at the build failure and I will try to add a test case.

DamienCassou:
I'm good. What do you think?

bbatsov:
But this command is generally supposed to be invoked in the root of a project, so I'm not certain if we'd be fixing a bug here.

DamienCassou:
Bozhidar Batsov <notifications@github.com> writes:
> But this command is generally supposed to be invoked in the root of a
> project, so I'm not certain if we'd be fixing a bug here.

I would like to call projectile-project-files in a subfolder of a git
project.

-- 
Damien Cassou
http://damiencassou.seasidehosting.st

"Success is the ability to go from one failure to another without
losing enthusiasm." --Winston Churchill

dustinpaluch:
I have this issue too. With a setup like this:
Screen Shot 2020-01-16 at 4 40 13 PM

When I do `M-x projectile-find-file` from `desired-project-root`, I expect to get the find-file UI with only the contents of `desired-project-root` as my choices. Instead, I get:

`apply: Setting current directory: No such file or directory, /test/desired-project-root/submodule-foobar/`

Because projectile-get-immediate-sub-projects returns a non-existent path.

Issue #97, opened 2013-04-26
ack and grep should include .projectile ignores


arianitu:
It would be really nice if there was an option to include your .projectile ignores when using ack and grep.

nordlow:
I'm using `counsel-projectile-rg` currently defined as

```
(defun counsel-projectile-rg (&optional options)
  "Search the current project with rg.

OPTIONS, if non-nil, is a string containing additional options to
be passed to rg. It is read from the minibuffer if the function
is called with a prefix argument."
  (interactive)
  (if (and (eq projectile-require-project-root 'prompt)
           (not (projectile-project-p)))
      (counsel-projectile-rg-action-switch-project)
    (let* ((ivy--actions-list (copy-sequence ivy--actions-list))
           (path
            (mapconcat 'shell-quote-argument
                       (or (projectile-normalise-paths
                            (car (projectile-parse-dirconfig-file)))
                           '("."))
                       " "))
           (ignored
            (mapconcat (lambda (i)
                         (concat "--glob !" (shell-quote-argument i)))
                       (append
                        (projectile--globally-ignored-file-suffixes-glob)
                        (projectile-ignored-files-rel)
                        (projectile-ignored-directories-rel))
                       " "))
           (counsel-rg-base-command
            (format (counsel-projectile--string-trim-right counsel-rg-base-command " \\.")
                    (concat ignored " %s " path))))
      (ivy-add-actions
       'counsel-rg
       counsel-projectile-rg-extra-actions)
      (counsel-rg (eval counsel-projectile-rg-initial-input)
                  (projectile-project-root)
                  options
                  (projectile-prepend-project-name
                   (concat (car (split-string counsel-rg-base-command)) ": "))))))
```

This change affect `counsel-projectile-rg` aswell?

PR #1371, opened 2019-01-12
Fix grep

jacksonrayhamilton:
This fixes https://github.com/bbatsov/projectile/issues/97 as it related to grep.

find’s “-name” option never matches a path, so binding `grep-find-ignored-files` would never allow us to ignore such.  Therefore, .`projectile` paths like the following would not ignore anything when grepping (except for directories unusually bearing extensions):

```
-/foo/bar.txt
```

It was only possible to ignore _all_ files in a project with a given name, and only by specifying that name with a leading slash:

```
-/foo.txt
```

Specifying a file or directory name without a leading slash would not ignore anything:

```
-foo.txt
-foo/
```

Also, if an ignore was a path to a directory that existed from the root, then all such paths to directories in a project would be ignored.  However, if the path was to a direct descendant of the root which was _not_ a directory (or did not even exist), then all _non-directory files_ in the project with that direct descendant’s name would be ignored, and if the path was to a 2nd-generation or later non-directory or non-existent descendant, then nothing would be ignored.

These strange cases stemmed from probable misuse of `grep-find-ignored-directories`, which seems to only be intended for ignoring directory names appearing anywhere, and misuse of `grep-find-ignored-files`, which seems to only be intended for ignoring file names appearing anywhere and without respect to parent directories.  Therefore, binding ignores via `grep-find-ignored-directories` and `grep-find-ignored-files` greatly limited what could be ignored.

To ignore paths to files and to get more consistent behavior, I thought it was warranted to write a special exclusion expression supporting paths and which only ignored files and directories relative to the root.  Also, to fulfill the promise of [the documentation](https://docs.projectile.mx/en/latest/projects/#ignoring-files) and provide support for relative paths (or patterns as I interpreted it), it was warranted to write another special exclusion expression supporting patterns occurring anywhere.

- [ ] The new code is not generating bytecode or `M-x checkdoc` warnings _I documented `projectile-rgrep-default-command` to the same extent as its source function, `rgrep-default-command`.  Let me know if you’d prefer more documentation._
- ~You've updated the readme (if adding/changing user-visible functionality)~

Thanks for considering these changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+* [#97](https://github.com/bbatsov/projectile/issues/97): Respect `.projectile`
+  ignores which are paths to files and patterns when using `projectile-grep`.
+
 ## 2.0.0 (2019-01-01)
 
 ### New features
--- a/projectile.el
+++ b/projectile.el
@@ -2815,7 +2815,8 @@ This is a subset of `grep-read-files', where either a matching entry from
 (defun projectile-rgrep-default-command (regexp files dir)
   "Compute the command for \\[rgrep] to use by default.
 
-Extension of the Emacs 25.1 implementation of `rgrep-default-command'."
+Extension of the Emacs 25.1 implementation of `rgrep-default-command', with
+which it shares its arglist."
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks for working on this! I really appreciate it! 🙇

PR #1374, opened 2019-01-23
Add documentation for projectile-sort-order

danielcompton:
Adds documentation on the different sorting methods and how they interact with the alien indexing method. I found the different methods a little bit confusing, and especially that they didn't work when alien indexing was turned on.

-----------------


bbatsov:
Thanks for tackling this! The documentation definitely has a lot of room of improvement. :-)

PR #1392, opened 2019-03-23
Fix cabal project detection

### LINKED ISSUE #1391: A `.cabal` directory is incorrectly considered a project indicator
When using cabal, a `.cabal` directory will exist inside the user's
home directory. Currently, projectile will interpret that as
a Cabal project existing directly in the home directory, and thus
every file opened that does not belong to a more specific project will
be categorized as a Haskell Cabal project.

Create a `.cabal` directory in your `$HOME`, the open a file in any directory that does not belong to a projectile project. Your buffer will be incorrectly be classified as belonging to a cabal project.

```
Projectile 20190126.1117 (melpa)
```

### Emacs version

GNU Emacs 26.1 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.4)

### Operating system

Debian Buster (amd64)

rotty:
Adjust the wildcard used to detect cabal projects to exclude a
`.cabal` dotfile (or dot-directory).

When using cabal, a `.cabal` directory will exist inside the user's
home directory. Without this change, projectile will interpret that as
a Cabal project existing directly in the home directory, and thus
every file opened that does not belong to a more specific project will
be categorized as a Haskell Cabal project.

Fixes #1391.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 * [#97](https://github.com/bbatsov/projectile/issues/97): Respect `.projectile`
   ignores which are paths to files and patterns when using `projectile-grep`.
+- [#1391](https://github.com/bbatsov/projectile/issues/1391) A
+  `.cabal` sub-directory is no longer considered project indicator.
 
 ## 2.0.0 (2019-01-01)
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #1394, opened 2019-03-30
Add new 'related-files-fn' option to use custom function to find test/impl/other files

pinetr2e:
I am working on the projects where multiple programming languages are used and the source file of each language is using different test prefix.  Currently projectile only supports a simple form of test prefix/suffix, which ignores extension.

Besides, for a given source file, there can be several related files(definition file, etc) I would like to visit them as 'other' files, which depend on the projects.

This PR enables users to use a custom function to specify the location of test, impl and other files in a consistent way. Instead of trying to provide 'policies' for all different sort of projects,  It will provide a 'mechanism' for users to express any rules.
 
Please check the updated projectile.md for the usage details.


doc/projects.md
```diff
@@ -90,6 +90,7 @@ Option           | Documentation
 :test-dir        | A path, relative to the project root, where the test code lives.
 :test-prefix     | A prefix to generate test files names.
 :test-suffix     | A suffix to generate test files names.
+:related-file    | A function to specify test/impl/other files in a more flexible way.
```
bbatsov: I'd name this `related-files` or `related-files-fn`. In hinsight I regret not adding a `-fn` suffix to the keys that are functions.

bbatsov:
That's a very cool idea! I'm just wondering about the exact format of it all. 

> Besides, for a given source file, there can be several related files(definition file, etc) I would like to visit them as 'other' files, which depend on the projects.

I think it might be nice if there was a declarative way to express this as well (for simple cases at least), so people won't always have to define a custom function. 

Might also be a good idea to expose some use-facing `find-related-file` command that gives you to the option to jump to all files related to the current one. That might be useful if we decide to grow this to include other types of files (e.g. docs, etc).

doc/projects.md
pinetr2e: Sounds good. I will change as suggested.

pinetr2e:
I like the suggestion. Now, I added `projectile-find-related-file` as user commands.  It will list 'kinds' if there is more than one kind before listing the related files.   

For the descriptive way of specifying the related files for the simpler cases, we can add 'helper function' which generates necessary 'related-files-fn' for users. In this way, I think, we can express many 'policies' without changing the mechanism. I would like to work on those changes in different PR on top of this PR.
(We can even use that policy to replace test-prefix/test-suffix as well)

bbatsov:
> I would like to work on those changes in different PR on top of this PR.

Fine by me!

> (We can even use that policy to replace test-prefix/test-suffix as well)

I'd be open to this. I feel that that approach was just a simple hack. 

Thanks for working on this! Great work!

PR #1401, opened 2019-04-08
Add :related-files-fn custom function helpers

pinetr2e:
This PR enables users to pass a list of custom functions to `:related-files-fn option` so that multiple custom functions can be merged effectively.

This PR also adds several helper functions to generate common custom functions easily. These helper functions can be used with other user-defined custom functions.

Example of helper usages:
```el
(setq my/related-files
      (list
       (projectile-related-files-fn-extensions :other '("cpp" "h" "hpp"))
       (projectile-related-files-fn-test-with-prefix "cpp" "Test")
       (projectile-related-files-fn-test-with-suffix "el" "_test")
       (projectile-related-files-fn-groups
        :doc
        '(("doc/common.txt"
           "src/foo.h"
           "src/bar.h")))))

(projectile-register-project-type
   ;; ...
   :related-files-fn #'my/related-files)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Add `related-files-fn` option to use custom function to find test/impl/other files
 * [#1019](https://github.com/bbatsov/projectile/issues/1019): Jump to a test named the same way but in a different directory.
 * [#982](https://github.com/bbatsov/projectile/issues/982) Add heuristic for projectile-find-matching-test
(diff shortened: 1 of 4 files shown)


bbatsov:
Great work! 

I think the next step would be to update the bundled project types in terms of the new config. Gradually we should deprecate the special config for `:test-prefix` and `:test-suffix`, as discussed previously.

PR #1402, opened 2019-04-09
Make projectile-ripgrep work like projectile-ag.

mattiasb:
Make `projectile-ripgrep` work like `projectile-ag`.

Specifically
- Expose the prefix arg as an actual argument
- Show that we're doing a regexp search
- Update docstring


bbatsov:
Thanks!

PR #1412, opened 2019-05-12
Improve error handling in default-generic-command

rgrinberg:
Improve the error message in two cases:

* The project type is not set
* The project type is not recognized

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3677,17 +3677,22 @@ Should be set via .dir-locals.el.")
 If found, checks if value is symbol or string.  In case of symbol
 resolves to function `funcall's.  Return value of function MUST
 be string to be executed as command."
-  (let ((command (plist-get (alist-get project-type projectile-project-types) command-type)))
-    (cond
-     ((stringp command) command)
-     ((functionp command)
-      (if (fboundp command)
-          (funcall (symbol-function command))))
-     ((and (not command) (eq command-type 'compilation-dir))
-      ;; `compilation-dir' is special in that it is used as a fallback for the root
-      nil)
-     (t
-      (user-error "The value for: %s in project-type: %s was neither a function nor a string." command-type project-type)))))
+  (when (not project-type)
+    (user-error "project-type is not set"))
+  (let ((project-type-props (alist-get project-type projectile-project-types)))
+    (when (not project-type-props)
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -3677,17 +3677,22 @@ Should be set via .dir-locals.el.")
...
-     ((stringp command) command)
-     ((functionp command)
-      (if (fboundp command)
-          (funcall (symbol-function command))))
-     ((and (not command) (eq command-type 'compilation-dir))
-      ;; `compilation-dir' is special in that it is used as a fallback for the root
-      nil)
-     (t
-      (user-error "The value for: %s in project-type: %s was neither a function nor a string." command-type project-type)))))
+  (when (not project-type)
```
bbatsov: Is this nil check really useful? I don't see what exactly we gain from it.

rgrinberg: It seems useful. It's quite common that one opens a buffer without a project attached. This is the most accurate error message to reflect that.

In fact, it's the other error message that is less useful. It doesn't seem likely to set a project-type outside of the list of possible types, but it can still happen so I thought it was useful.

bbatsov: Yeah, but as this is basically parameter validation probably the caller should actually raise some error if there's not some project. I just find it weird to raise errors when a mandatory parameter is missing.

rgrinberg: > Yeah, but as this is basically parameter validation probably the caller should actually raise some error if there's not some project

That wouldn't be a bad idea either. Though it will be a bit more work as there are many callers.

> I just find it weird to raise errors when a mandatory parameter is missing.

Why is that? I find that pre-condition checks for the assumptions this function makes are useful. I agree that this isn't really a user, but a programming error however.

bbatsov: Because if we go down this route we should be validation parameters in every function and we generally don't want this. :-) Anyways, maybe this is ok in this particular case, but it seems to me we can do better.

rgrinberg: Well, it's a bit of a shame that it's tedious in emacs lisp, but the idea is quite sound. See any type system or clojure's spec for example.

bbatsov: I get your point, but every language comes with its own set if peculiarities. I've programmed in many dynamic languages and basic argument validation is uncommon in all of them.

PR #1414, opened 2019-05-13
Fix root caching when root doesn't exist

rgrinberg:
Previously, the cache for all the root finding functions would only work in the
case when the root was found. However, caching is just useful for buffers where
there's no project defined. Therefore, we enable caching in such situations.

A concrete case where this drastically improves the performance is when used with `ivy-rich`. `ivy-rich` fetches the project root for every buffer and it's important for this to be as fast possible.

-----------------

Thanks!


rgrinberg:
I still need to fix the tests here, but I just want to make sure the idea is welcomed by the maintainers before doing more work.

PR #1417, opened 2019-05-16
Fix remote projects removed if can't connect

### LINKED ISSUE #1340: Project is removed from the list if the project root is not found
## Expected behavior

Projects should be kept in the project list, even if the project root is not found. For example, a project with a remote root directory may not be reachable if the host or the network connection is down.

This is related to #1328.

## Actual behavior

Projects with no project root are removed from the project list.

- Create a `~/testproject` directory
- Add the `~/testproject` to the project list with <kbd>M-x projectile-add-known-project</kbd>
- Delete the `~/testproject` directory
- Switch to the project.

```
Projectile 20181024.1707
```

### Emacs version

*26.1*

### Operating system

*macOS 10.14*

m3rcuriel:
Fixes #1340 for TRAMP remote projects.

-----------------

- [?] You've added tests (if possible) to cover your change(s)

-------------

(I don't this adds user-visible functionality)

Not sure how I could add tests to cover this.


projectile.el
```diff
@@ -4030,7 +4030,7 @@ With a prefix ARG invokes `projectile-commander' instead of
 Invokes the command referenced by `projectile-switch-project-action' on switch.
 With a prefix ARG invokes `projectile-commander' instead of
 `projectile-switch-project-action.'"
-  (unless (projectile-project-p project-to-switch)
+  (unless (or (projectile-project-p project-to-switch) (file-remote-p project-to-switch))
```
bbatsov: Please, add some comment here explaining why that's needed.

bbatsov:
> (I don't this adds user-visible functionality)

If it fixes a bug it's a user-visible change. It should get an entry in the changelog.

projectile.el
bbatsov: ping :-)

rsaite:
Hi @m3rcuriel  @bbatsov , I ran into the same issue (remote projects being removed from projectile projects) and tried to apply the fix with the following result on Spacemacs 0.300.0:
* Remote projects are no longer removed from projectile projects. :heavy_check_mark: 
* But there are issues with loading them:
** If I use `<SPC>-p-p-<Project name>-<RET>` (`helm-projectile-switch-project`), nothing happens in this buffer, i.e., I do not get a view of the remote project directory. :x: 
** If I click the link to the project on the home buffer, a buffer with a view of the remote project directory opens :heavy_check_mark:  and immediately with it a buffer with a file in the project (the first file?) :x: :question: If I then try `<SPC>-p-p-<Project name>-<RET>`  again, it does work, because a TRAMP-connection has been established.

So, the issue seems to be that somehow clicking on the project link the home buffer opens a TRAMP-connection (but I don't know how this works/what code is executed when clicking the link), while `helm-projectile-switch-project` (which calls/is remapped to `projectile-switch-project`) does not establish this connection. This should probably also be fixed for this PR to make sense. (I would fix it if I could, but this is my first time digging through elisp code and I am not sure how this should be implemented.)

sshantt:
I'm also having issues with `projectile-switch-project` not initializing ssh connections. Once I somehow connect to remote, `projectile-switch-project` works fine.

PR #1418, opened 2019-05-21
Enhance go project detection by referring to go.mod file

rudle:
This PR fixes an issue encountered when working in a go project that doesn't have any `.go` files at the root. The fix is to check for a `go.mod` file at the root of the project, avoiding a recursive descent of the project. The `go.mod` file is new-ish and will be present in projects that are using Go modules. The file will be present in all (or almost all) Go projects when Go 1.13 is released later this year.

I didn't add any tests, but this is the fix I applied to my local install of projectile and it worked great.

Feel free to remove or tweak my Changelog entry, this change may not be big enough to warrant documentation.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2508,8 +2508,8 @@ test/impl/other files as below:
 
 (defun projectile-go-project-p ()
   "Check if a project contains Go source files."
-  (or (projectile-verify-file-wildcard "*.go")
-      (projectile-verify-file "go.mod")))
+  (or (projectile-verify-file "go.mod")
+      (projectile-verify-file-wildcard "*.go")))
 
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -2505,7 +2505,8 @@ test/impl/other files as below:
 
 (defun projectile-go-project-p ()
   "Check if a project contains Go source files."
-  (projectile-verify-file-wildcard "*.go"))
+  (or (projectile-verify-file-wildcard "*.go")
```
bbatsov: I think you should change the order here for efficiency's sake.

rudle:
Thanks for the review. Feedback and merge conflicts have been addressed.

bbatsov:
Thanks!

PR #1430, opened 2019-06-17
Fix go build: go build ./... doesn't create a binay.

Osram-Gru:
Fix projectile-compile-project in go projects: Calling `go build ./...` doesn't build a binary.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2534,7 +2534,7 @@ test/impl/other files as below:
                                   :test "cabal test"
                                   :test-suffix "Spec")
 (projectile-register-project-type 'go projectile-go-project-test-function
-                                  :compile "go build ./..."
+                                  :compile "go build"
                                   :test "go test ./..."
                                   :test-suffix "_test")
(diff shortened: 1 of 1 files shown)


bbatsov:
Should we change `go test` as well? (I've never used `go` so I've got no idea what makes sense)

Osram-Gru:
I'm not that firm in go, neither, but I think in `go test` it could make sense.

bbatsov:
OK, let's update it as well.

Osram-Gru:
Ah, sorry for being ambigous: `go test ./...` can make sense, so I think we should leave it as it is 0:)

Osram-Gru:
Hi bbatsov, may I ping you on this PR? 
`go build ./...` doesn't make sense, but `go test ./...` does. So you could accept this PR just like it is. If there's something else missing, please let me know!

bbatsov:
Thanks for the reminder (and the PR, of course)!

PR #1437, opened 2019-07-25
Fixed .projectile not working with hybrid indexing

tomc1998:
The git indexing was returning a list of files starting with './'. This
means the `projectile-remove-ignored` function didn't work, since it
searches for string prefixes - ./ is now stripped from the start of
files.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1323,27 +1323,28 @@ otherwise operates relative to project root."
         (ignored-dirs (projectile-ignored-directories-rel)))
     (cl-remove-if
      (lambda (file)
-       (or (cl-some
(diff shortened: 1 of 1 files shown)


m-ar-c:
I confirm that with the hybrid indexing method, patterns in my .projectile are ignored (projectile 20200710.947). When I manually apply tomc1998's code, .projectile file is working again.

m-ar-c:
I confirm that the patterns in .projectile does nothing (files still shows up) whit hybrid indexing method with projectile 20200710.947. I tried to manually apply tomc1998 ' code but it didn't work (could not load projectile anymore).

PR #1438, opened 2019-08-10
Make sure projectile-files-via-ext-command returns files not errors

maurelio1234:
If a command fails, it may return an error message like:
`fatal: No url found for submodule path 'XXX' in .gitmodules`.

It just filters out error messages to avoid the calling code to try to
use it as a file name.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1356,10 +1356,15 @@ VCS is the VCS of the project."
   "Get a list of relative file names in the project ROOT by executing COMMAND.
 
 If `command' is nil or an empty string, return nil.
-This allows commands to be disabled."
+This allows commands to be disabled.
+
+Only text sent to standard output is taken into account."
   (when (stringp command)
     (let ((default-directory root))
-      (split-string (shell-command-to-string command) "\0" t))))
+      (with-temp-buffer
+        (call-process-shell-command command nil '(t nil))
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -1306,8 +1306,9 @@ VCS is the VCS of the project."
 If `command' is nil or an empty string, return nil.
 This allows commands to be disabled."
   (when (stringp command)
-    (let ((default-directory root))
-      (split-string (shell-command-to-string command) "\0" t))))
+    (let* ((default-directory root)
+           (files (split-string (shell-command-to-string command) "\0" t)))
+      (cl-remove-if-not #'file-exists-p files))))
```
bbatsov: The problem with this is that if you get a long list of files the processing is going to be very slow. I've deliberately avoided iterating over the project files in many places for that particular reason.

maurelio1234: I see. I will propose something else. Thanks for your feedback ;)

maurelio1234: Just pushed a commit using a different approach: just read what is written to stdout.

immerrr:
Got hit by this, and was going to propose a PR, but found that there is one already.

@bbatsov any chance on getting this merged?

bbatsov:
Seems I forgot about it, but it looks in good shape. Thanks, @maurelio1234!

PR #1448, opened 2019-09-29
Update dotnet project detection

Gastove:
Previously, this function looked only at `.[cf]sproj` files in the _root_ of the
project directory. _However_, this is not how dotnet projects are typically laid
out. Instead, the `.[cf]sproj` file is often located in a directory like
`src/ProjectName/Project.fsproj`. This is because a given repo might contain
_many_ projects, which are then tied together into a root "solution" file.

This commit checks for the solution (`.sln`) file /first/, then does the
existing "same dir" check plus a `src/*/` check for the project files.

I looked through the current test suite, and I didn't _see_ any tests for this kind of predicate function. I am, however, very happy to add one if one is preferred. 

(Changelog and a docstring for the function I edited coming in the next commit.)

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 * [#1419](https://github.com/bbatsov/projectile/pull/1419): When possible, use [fd](https://github.com/sharkdp/fd) instead
   of `find` to list the files of a non-VCS project. This should be much faster.
 
+### Changes
+
+* A project now registers as `dotnet` if it has a solution (`*.sln*`) file at the project root or project files (`*.[fc]sproj`) in `src/*/`, as well as the old "project files in project root" heuristic.
+
 ### Bugs fixed
 
 * [#97](https://github.com/bbatsov/projectile/issues/97): Respect `.projectile` ignores which are paths to files and patterns when using `projectile-grep`.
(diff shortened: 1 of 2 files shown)


Gastove:
Going by what I see in the codebase, I'm not going to write a unit test for `projectile-dotnet-project-p` or update the `README`. I'm leaving those boxes un-ticked in my PR description until I hear back from a maintainer. Happy to do either if either is desired.

bbatsov:
The proposed solution would pull the subprojects into the "main" project which may or may not be what people would expect. I don't use .net, so I can't really decide what's the best course of action here.

Gastove:
Oh, that's a great point. 

Can you help me understand something here? Right now, what I observe is:

1. I open a `.NET` project; there's a `.git` directory in the root, so that becomes the project. 
2. Now, no matter where I am in the file heirarchy, it appears projectile is looking up into the project root, where it doesn't find anything particularly helpful; the project is marked `generic`. 

I don't know what the "correct" pattern for handling sub-projects is, but I'm delighted to implement it if you can point me in the right direction. Where _this_ PR is concerned, I expect the most correct thing to do is to remove the new `.[fc]sproj` checks, and instead look at the project root only for the `.sln` file -- `.sln` (solution) files tie together a complete project, so if one is present, I feel like it's very safe to consider the directory with the solution file in it the root of a `.NET` project.

PR #1450, opened 2019-10-05
Call `switch-project-action` within project's temp buffer

kevinjfoley:
This makes sure that any buffer local variables that were loaded by
`hack-dir-local-variables-non-file-buffer` are loaded with the
appropriate variables when the action is called.

Also removes the code to set project name to the new project as it should be properly set given we're now running from the new project's buffer.

-----------------

  - ~I'm not sure of how/if this should be tested, open to suggestions~

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4053,17 +4053,17 @@ With a prefix ARG invokes `projectile-commander' instead of
     (let ((default-directory project-to-switch))
       ;; use a temporary buffer to load PROJECT-TO-SWITCH's dir-locals before calling SWITCH-PROJECT-ACTION
       (with-temp-buffer
-        (hack-dir-local-variables-non-file-buffer))
-      ;; Normally the project name is determined from the current
-      ;; buffer. However, when we're switching projects, we want to
-      ;; show the name of the project being switched to, rather than
-      ;; the current project, in the minibuffer. This is a simple hack
-      ;; to tell the `projectile-project-name' function to ignore the
-      ;; current buffer and the caching mechanism, and just return the
-      ;; value of the `projectile-project-name' variable.
-      (let ((projectile-project-name (funcall projectile-project-name-function
-                                              project-to-switch)))
-        (funcall switch-project-action)))
+        (hack-dir-local-variables-non-file-buffer)
+        ;; Normally the project name is determined from the current
+        ;; buffer. However, when we're switching projects, we want to
+        ;; show the name of the project being switched to, rather than
+        ;; the current project, in the minibuffer. This is a simple hack
+        ;; to tell the `projectile-project-name' function to ignore the
+        ;; current buffer and the caching mechanism, and just return the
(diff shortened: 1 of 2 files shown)


kevinjfoley:
I added a test to demonstrate but there is probably a more straight forward/clearer way to test, would love some feedback on that.

bbatsov:
I have a feeling that the code used to look like this and then someone reported the issues with the project name. Seems like some enchanted circle. :-)

kevinjfoley:
You're right!  I took a look at the log (should have done that in the first place) and found the related conversation at - https://github.com/bbatsov/projectile/issues/1067

It looks like this change would bring back some of the issues there.  I think moving the hook run into the the `with-temp-buffer` body would resolve everything.  Can you think of any issues with that?

I'll try to go through the thread and add a test for the issues described there to try to make sure this doesn't reintroduce old issues.

Deewiant:
The key part of this is reverting the change from 72a8be50b9e6d4c23a178f777043e67794fc9728 that moves the `switch-project-action` call outside the `with-temp-buffer`. It seems like cdf9c228ccdcb57b73184f10ea3f1e2e4e03d320 ended up fixing the #1067 issue but the move of `switch-project-action` was left, defeating the reason for adding the dir-local variable handling in the first place.

This PR does look problematic as-is. In particular, "it should be properly set given we're now running from the new project's buffer" doesn't make sense to me, since we're running in a temp buffer that isn't associated to the new project as such, so I think the `projectile-project-name` handling still has its place.

I've rebased this at https://github.com/Deewiant/projectile/commits/switch-project-dir-local so that the `projectile-project-name` handling is left as-is, just moving it inside the same `with-temp-buffer` that receives the dir-local variables. It looks like it'd satisfy everyone, and the test from here passes. (I had actually made the analogous change myself at https://github.com/Deewiant/projectile/commit/f8d2b0613c0df8b50890211e00b26708e052ce3a and was preparing to PR that and only noticed this old PR afterwards, d'oh.)

@kevinjfoley Would you be happy to update this PR by reverting the `projectile-project-name`-related part? (You could grab my rebase or then re-do it, it's fairly trivial either way.)
@bbatsov Would that be sufficient to get this in, or do you still need tests added for #1067 to determine whether this is acceptable?

I'm keen on having this change in to be able to override `projectile-git-submodule-command` on a per-project basis: I have one particular project with some huge git submodules that cause `projectile-find-file` to really bog down and as a result `projectile-switch-project` is basically unusable for that project. This change resolves that issue and I'd rather have it upstreamed than maintain it on my end.

kevinjfoley:
@Deewiant - I'm happy to take a look and adjust based on your changes. However I'm currently in vacation so I won't be able to take a look until Sunday at the earliest.

bbatsov:
No rush.

kevinjfoley:
Rebased to keep the removed code.  Sorry about the delay!

PR #1451, opened 2019-10-05
Return nil when a command doesn't have a default value

MetaDark:
Fixes: https://github.com/bbatsov/projectile/issues/1270
Equivalent to: https://github.com/bbatsov/projectile/pull/1268 (but error is still signalled if command is invalid)
Generalizes: https://github.com/bbatsov/projectile/pull/1269

This fixes the regression introduced in https://github.com/bbatsov/projectile/pull/1255 that prevents a user from running a command when it doesn't have a default value.

@bbatsov originally proposed in https://github.com/bbatsov/projectile/pull/1269 that the solution could be generalized by prompting the user; but if we did this, we would prompt the user twice when using projectile-*-project commands. By returning nil, we leave it up to the caller to handle errors. This is equivalent to the behaviour before https://github.com/bbatsov/projectile/pull/1255.

---
- [x] No previously passing tests are failing
  - `projectile-find-file-in-directory` fails when called in a non-existing directory, but it was failing before this change


bbatsov:
Thanks!

PR #1454, opened 2019-10-13
Autoload for projectile-locate-dominating-file

rgrinberg:
Trivial change to allow using `projectile-locate-dominating-file` with autoload


bbatsov:
Why would you want to autoload functions which are not commands? That's extremely uncommon.

rgrinberg:
In doom-emacs these functions are used as library functions. On second
thought, I’m not sure why we don’t just require projectile

hlissner:
I agree that library functions shouldn't be autoloaded by packages, so I've fixed this on Doom's end. This can be closed.

bbatsov:
Thanks!

PR #1457, opened 2019-10-19
Autoload projectile-add-known-project

hlissner:
As a follow up to #1454, this only autoloads `projectile-add-known-project`, as it is an interactive command.


bbatsov:
Thanks!

PR #1460, opened 2019-10-24
Fix incorrect order of if statement forms in projectile-replace

### LINKED ISSUE #1459: projectile-replace: Symbol’s function definition is void: fileloop-initialize-replace
## Expected behavior
`projectile-replace` should work correctly. 

## Actual behavior
Instead, when running emacs 26.2, an attempt is made to call a void function, `fileloop-initialize-replace`

The following error is shown: `if: Symbol’s function definition is void: fileloop-initialize-replace`

Try to call `projectile-replace`. It will fail due to incorrect ordering of then-vs-else forms in the `(if (fboundp #'fileloop-continue)` condition.

I have already fixed this locally. I'll submit a PR with a reordering of the resulting forms of the `(if (fboundp #'fileloop-continue)` condition.

```
Projectile version: 20191021.1201
```

### Emacs version
26.2

### Operating system
MacOS

roryhow:
This PR resolves #1459 

The resulting forms of the `(fboundp #'fileloop-continue)` condition are in the incorrect order. So, this swaps them so that the code works as intended. 

No "new" code is written here, just a reordering of older things, so many of the below checkboxes aren't valid for this case.

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3481,23 +3481,22 @@ to run the replacement."
                      (format "Replace %s with: " old-text))))
          (files (projectile-files-with-string old-text directory)))
     (if (fboundp #'fileloop-continue)
-        ;; Emacs 25 and 26
-        ;;
-        ;; Adapted from `tags-query-replace' for literal strings (not regexp)
-        (progn
-          (setq tags-loop-scan `(let ,(unless (equal old-text (downcase old-text))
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks for spotting and fixing my silly mistake!

PR #1461, opened 2019-10-24
Changed index to ignore directory by names only

### LINKED ISSUE #1250: projectile-globally-ignored-directories not working with native indexing.
## Expected behavior
Ignore non-top-level dirs listed in projectile-globally-ignored-directories with both methods of indexing (native and alien). 

## Actual behavior
Fix introduced in #1119 fixes ignoring of non-top-level dirs for alien indexing mode but not for native.

1. Put this in init.el:
> projectile-globally-ignored-directories (append '("*\_\_pycache\_\_/")
> 							            projectile-globally-ignored-directories)

2. Run C-u C-c p-f in a python project with generated \_\_pycache\_\_ dirs.
3. The \_\_pycache\_\_ dirs are there even though they have to be ignored.

OSX 10.13 and Windows 10.
Emacs 26.1 and older.
projectile 20180616.252 from Melpa

ax487:
This should fix #1250. Essentially, there is now a list of ignored directory names which are checked during the native indexing.

Apologies for any stupid mistakes I have never hacked on emacs before...

Regarding the checkboxes:

- projectile already generates checkdoc warnings?!
- should the fucntions be visible?

-----------------

Thanks!


bbatsov:
> projectile already generates checkdoc warnings?!

Don't worry about those.

> should the fucntions be visible?

Probably not, although in Elisp everything is public. :-)

Some unit tests would be great as well.

opsnull:
any updating ？

SreenivasVRao:
Hi folks - would appreciate if this change were merged in. :)

bbatsov:
@ax487 ping :-) Mind updating this on top of `master`?

projectile.el
```diff
@@ -1222,27 +1222,22 @@ Files are returned as relative paths to DIRECTORY."
 
 The function calls itself recursively until all sub-directories
 have been indexed.  The PROGRESS-REPORTER is updated while the
-function is executing.  The list of IGNORED-FILES and
```
bbatsov, line 1225: I'm guessing something got merged incorrectly by looking at the diff.

ax487: Should be fixed now, thanks for the spot

ax487:
@bbatsov: Could you take another look at the merge request? I added a working unit test and an entry in the changelog

projectile.el
```diff
@@ -1224,25 +1224,21 @@ The function calls itself recursively until all sub-directories
 have been indexed.  The PROGRESS-REPORTER is updated while the
 function is executing.  The list of IGNORED-FILES and
 IGNORED-DIRECTORIES may optionally be provided."
-  ;; we compute the ignored files and directories only once and then we reuse the
```
bbatsov, line 1227: Why did you remove this comment? It's kind of important. 😄

projectile.el
```diff
@@ -1224,25 +1224,21 @@ The function calls itself recursively until all sub-directories
...
-  ;; pre-computed values in the subsequent recursive invocations of the function
   (let ((ignored-files (or ignored-files (projectile-ignored-files)))
-        (ignored-directories (or ignored-directories (projectile-ignored-directories))))
-    (apply #'append
+	      (ignored-directories (or ignored-directories (projectile-ignored-directories))))
```
bbatsov, line 1228: This seems misaligned.

projectile.el
```diff
@@ -1224,25 +1224,21 @@ The function calls itself recursively until all sub-directories
...
+    (apply 'append
```
bbatsov, line 1229: The change to append seems accidental.

projectile.el
```diff
@@ -1682,6 +1678,21 @@ according to PATTERNS: (ignored . unignored)"
      (projectile-project-ignored-files)))
    (projectile-unignored-files)))
 
+(defun projectile-ignored-directory-names ()
```
bbatsov, line 1681: I think this should have "globally" somewhere in its name.

projectile.el
```diff
@@ -1682,6 +1678,21 @@ according to PATTERNS: (ignored . unignored)"
...
+  "Return list of ignored directory names."
+  (projectile-difference
+   projectile-globally-ignored-directories
+   projectile-globally-unignored-directories))
+
+(defun projectile-ignored-directory-name-p (directory-name)
```
bbatsov, line 1687: Same here.

bbatsov:
Btw, from what I gather the gist of the change is changing the original predicate that checks for ignored directories with a different one. What exactly is the problem with the original predicate function?

projectile.el
ax487: My mistake, fixed

ax487:
@bbatsov: Should all be resolved now

bbatsov:
> Btw, from what I gather the gist of the change is changing the original predicate that checks for ignored directories with a different one. What exactly is the problem with the original predicate function?

Don't forget about this. In particular I'm most worried about you removing:

```
(projectile-ignored-directory-p (file-name-as-directory f) ignored-directories)
```

The code was written in this way to reduce the number of calculations of ignored directories during the directory indexing, and we have to be careful to make the indexing slow.

ax487:
@bbatsov: The problem with the original predicate is the following:

The `projectile-ignored-directories` function is used to compute the list of directory names based in part on the list of `projectile-globally-ignored-directories` which is a customizable variable. In the process, the directories are expanded with respect to the root, so `.svn` becomes `/project/.svn`, which does not match subdirectories like `/project/subdirectory/.svn`. This is the root cause of the problem and the reason why the predicate `projectile-ignored-directory-p` is not working.

Ironically, we need to match the *local* directory name to the `projectile-globally-ignored-directories`...

It occurs to me that my approach is not correct either, since it does not respect the contents of the ".projectile" file, that is, the set of project-local ignored directories.

I think I will instead fix the `projectile-ignored-directory-p` prefix directly. I will include some optional parameters to increase the speed.

ax487:
@bbatsov: I finished the rewrite, I hope this makes everything a little clearer... 

Performance-wise I used the same caching as before and added a test to ensure that the globally-ignored directory names are computed only once.

bbatsov:
The patch seems simpler indeed and I'll merge it, but it seems to me we can simplify things further if we merged the files coming from the project config and globally ignored folders and just treated them differently based on the number of segments in them - e.g. dirs like "dir" get ignored at every level, something like "path/to/dir" gets ignored only if it's relative to the root. 

Then we won't have to juggle several different ignored lists to check against and the results would be the same IMO.

SreenivasVRao:
Hi - I'm not sure why, but I'm still seeing the same issue. I do have the latest commit at `3ad2008` 

I have $PROJECT_DIR/src/subproject1 and $PROJECT_DIR/src/subproject2

and with projectile-find-file, I still see `$PROJECT_DIR/src/subproject*/.git` in the completions.

My ignored-dirs are
```
("*.git*" "env" ".log" "build" "flow-typed" "node_modules" "~/doom-emacs/.local/" ".idea" ".vscode" ".ensime_cache" ".eunit" ".git" ".hg" ".fslckout" "_FOSSIL_" ".bzr" "_darcs" ".tox" ".svn" ".stack-work" ".ccls-cache" ".cache" ".clangd")
```
and I use `hybrid` indexing. Any advice?

ax487:
@SreenivasVRao: I think you should remove the "*", there is no globbing for the ignored directories.

bbatsov:
Indeed. I guess we should cover this explicitly in the docs.

SreenivasVRao:
Hi - sorry my example is a little confusing - you can see both "*.git*" and ".git" in that list above. I tried without the `*.git*` and I still got the same issue.

SreenivasVRao:
(bump) sorry to bother you folks. Some advice on debugging would be helpful.

bbatsov:
See https://docs.projectile.mx/projectile/troubleshooting.html#debugging-projectile-commands

You want to debug `projectile-index-directory`.

PR #1462, opened 2019-10-29
Add minimal support for shadow-cljs

Macroz:
Add minimal support for [http://shadow-cljs.org/](shadow-cljs) projects.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [#1418](https://github.com/bbatsov/projectile/pull/1418): The presence of a `go.mod` file implies a go project.
 * [#1419](https://github.com/bbatsov/projectile/pull/1419): When possible, use [fd](https://github.com/sharkdp/fd) instead
   of `find` to list the files of a non-VCS project. This should be much faster.
+* Add minimal support for [http://shadow-cljs.org/](shadow-cljs) projects.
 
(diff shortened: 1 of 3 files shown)


Macroz:
@bbatsov Would you be willing to consider this addition? I don't use `shadow-cljs` much yet but I noticed this was missing.

I think most `shadow-cljs` projects have Git or some other type of projectile supported project type already. However it's nice to get the plain project working as well.

bbatsov:
Don't shadow projects typically have build tool marker in them? (e.g. `deps.edn`, `project.clj`, etc) You can't do much with shadow without adding some deps.

Macroz:
As for dependencies, you can do fine without as described in the [manual](https://shadow-cljs.github.io/docs/UsersGuide.html#_build_tool_integration). But it's true, like I note in the original question, that most projects will have e.g. Git. But isn't that true for the other types as well? As it is now, projectile + CIDER does not work in a plain shadow-cljs project unless you add the `.projectile` file or Git. This would fix that.

bbatsov:
Usually the other project types bring to the table some build commands, which is why we have them. They are not added just for their value as project markers. Anyways, there's no harm in adding this, but it won't add much value either.

Macroz:
I guess this is not worthwhile so I'll close the issue.

PR #1473, opened 2019-12-01
fix Travis

peterbecich:
The third and fourth jobs were broken due to what was fixed by this, I think: https://github.com/rejeep/evm/pull/130

All jobs now succeed: https://travis-ci.org/peterbecich/projectile/builds/619137833

The Emacs 26.1 job is updated to Emacs 26.3.

-----------------

- ~You've added tests (if possible) to cover your change(s)~

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 sudo: false
 language: emacs-lisp
+dist: xenial
 env:
   - EMACS_BINARY=emacs-25.1-travis MAKE_TEST=test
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks! Long-term we should switch the build to Circle and the newer Emacs docker builds like here https://github.com/clojure-emacs/clojure-mode/blob/master/.circleci/config.yml, but this will do for now.

Issue #1275, opened 2018-08-24
projectile-find-file-in-directory does not work in TRAMP projects


codemac:
https://github.com/bbatsov/projectile/blob/a4b447d980b10cbb8d175d64e4305b4504c03d83/projectile.el#L3569

The resolution probably relates to expand-file-name being passed a tramp directory. I will post here any updates I find to fixing it.

codemac:
Currently it attempts to open the file on the local FS, so /ssh:machine:/blah/blah ==> /blah/blah locally.

codemac:
projectile-indexing-method 'native solves this, but takes a super long time to index, so this may be a deeper issue with how the alien method returns remote projects' name.

codemac:
Yeah, so it seems the cached filenames need to be normalized when indexing between alien & native methods for tramp projects. The absolute paths the alien method produces shortcut the TRAMP default-directory. The `expand-file-name` method returns the name without the tramp default directory when handed an absolute filename.

bbatsov:
@codemac Mind putting some PR together to tackle this. 

> Yeah, so it seems the cached filenames need to be normalized when indexing between alien & native methods for tramp projects.

Which cached filenames? Those in the project cache, right?

codemac:
Sorry, @bbatsov yes these are the cached project filenames.

codemac:
I stopped using TRAMP, and only use projectile locally due to other extensions issues with assuming local buffers.

I don't have time to implement this, but I do think this is still an issue for projectile's file cache.

TrunovS:
> Currently it attempts to open the file on the local FS, so /ssh:machine:/blah/blah ==> /blah/blah locally.

Yes. Alien indexing method returned absolute file paths  which in turn failed `projectile-find-file` in Tramp session for me. Got that working with fix in the way `projectile-dir-files-alien` create list of files with relative paths instead of absolute paths. And it seams that fixed `find-file-in-directory` too.

thornjad:
I use TRAMP every day and am entirely unable to use Projectile on any remote project. I'd really like to and don't want this issue to be closed as stale.

TrunovS:
> I use TRAMP every day and am entirely unable to use Projectile on any remote project. I'd really like to and don't want this issue to be closed as stale.

Got that working with some combination of:
  - ```(add-to-list 'tramp-remote-path 'tramp-own-remote-path)```
 - ```(setq-default projectile-indexing-method 'alien)```
[...]

PR #1477, opened 2019-12-07
Fix tramp alien index find file with relative paths.

TrunovS:
Alien indexing method returned absolute file paths which in turn failed `projectile-find-file` in Tramp session for me. Got that working with fix in the way `projectile-dir-files-alien` create list of files with relative paths instead of absolute paths. And I believe its solves the way `projectile-find-file-in-dir` works with alien indexing with tramp.

Closes: #1275

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1194,12 +1194,18 @@ function is executing."
 ;; any post-processing of the files obtained via the external command.
 (defun projectile-dir-files-alien (directory)
   "Get the files for DIRECTORY using external tools."
-  (let ((vcs (projectile-project-vcs directory)))
-    (cond
-    ((eq vcs 'git)
-     (nconc (projectile-files-via-ext-command directory (projectile-get-ext-command vcs))
-            (projectile-get-sub-projects-files directory vcs)))
-    (t (projectile-files-via-ext-command directory (projectile-get-ext-command vcs))))))
+  (let ((vcs (projectile-project-vcs directory))
+        ;; getting directory name without ":" for tramp connections
+        (c_directory (car (last (split-string directory ":")))))
+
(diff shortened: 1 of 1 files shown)


TrunovS:
@bbatsov It seams something wrong with test concerning projectile-project-root and its cache? But I cant understand why.

projectile.el
```diff
@@ -1194,12 +1194,18 @@ function is executing."
...
-    (cond
-    ((eq vcs 'git)
-     (nconc (projectile-files-via-ext-command directory (projectile-get-ext-command vcs))
-            (projectile-get-sub-projects-files directory vcs)))
-    (t (projectile-files-via-ext-command directory (projectile-get-ext-command vcs))))))
+  (let ((vcs (projectile-project-vcs directory))
+        ;; getting directory name without ":" for tramp connections
+        (c_directory (car (last (split-string directory ":")))))
+
+    (mapcar (lambda (f)
```
bbatsov: I haven't looked at the test failure, but I can tell you that walking over all the files and applying an extra transformation is going to have a terrible performance impact on any non-trivial project.

TrunovS: Thanks!) Ok, that's was kind of hack to make file list more compact without long prefix `/home/user/<proj_root>/` for hg repos. It's seams better place for that was `projectile-hg-command` (change hg locate to hg files). By chance `hg files` also returns project file list in more compact way without long prefix.

TrunovS:
@bbatsov Last commit likely fixed some test failures for Travis emacs 26 and snaptshot, but not so fo emacs 25.

TrunovS:
This feature is somewhat buggy and still need to be refined.

PR #1478, opened 2019-12-08
Fix Emacs 27 warnings

Silex:
Pretty much self explanatory :-)

The only thing I'm not sure is about the "emacs27" strings, what should I use?

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -63,6 +63,8 @@
 (declare-function vc-dir "vc-dir")
 (declare-function vc-dir-busy "vc-dir")
 (declare-function string-trim "subr-x")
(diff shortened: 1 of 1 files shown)


bbatsov:
> The only thing I'm not sure is about the "emacs27" strings, what should I use?

No idea. :-)

Silex:
Found this:

```
(declare-function FN FILE &optional ARGLIST FILEONLY)

Tell the byte-compiler that function FN is defined, in FILE.
The FILE argument is not used by the byte-compiler, but by the
‘check-declare’ package, which checks that FILE contains a
definition for FN.
```

I thus changed it to `fileloop`. You can merge :wink:

PR #1483, opened 2019-12-18
Allow multiple root-dirs to be search by projectile-grep

### LINKED ISSUE #1482: projectile-grep stalls when project has multiple roots
# Expected behavior

A project containing multiple directories specified in the `.projectile` file as in

```
+/foo
+/bar
+/baz
```

Hangs `projectile-grep` asking for user input to either, kill the current grep and replace the `*grep*` buffer or continue to hang. Either only the first directory is searched or only the last results are kept, not all results are visible at the same time. This behaviour is specific to `grep` as it uses the same `grep` buffer.

## Actual behavior

Search all directories included in the project for results and make the results visible at the same time.

1) Create a project with multiple subdirectories 
2) Add the subdirectories to the `.projectile` file as described above
3) Run `projectile-grep` in the project in a way that produces results

Projectile 2.1.0snapshot

### Emacs version

GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.20 Version 10.14.3 (Build 18D109)) of 2019-09-02

### Operating system

MacOS 10.15.2

sideshowcoder:
When a project is configured via `.projectile` to include multiple directories
explicitly this results in `projectile-grep` to run over multiple roots.

```
+/foo/bar
+/baz/
```

With a single global *grep* buffer this results in either

a) losing all but the search result of the last root
b) Stalling as the user is asked per root to kill the grep buffer

To avoid this the name of the `*grep*` buffer should to include the current
`root` being search associated with the buffer like `*grep <~/source/foo/bar>*`.

Fixes #1482

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [#1385](https://github.com/bbatsov/projectile/issues/1385): Update `projectile-replace` for Emacs 27.
 * [#1432](https://github.com/bbatsov/projectile/issues/1432): Support .NET project.
 * [#1270](https://github.com/bbatsov/projectile/issues/1270): Fix running commands that don't have a default value.
+* [#1482](https://github.com/bbatsov/projectile/issues/1482): Run a seperate grep buffer per project root
 
 ## 2.0.0 (2019-01-01)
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Nice improvement! Thanks!

PR #1485, opened 2020-01-01
[Fix #1475] Fix unignoring directories in hybrid mode

### LINKED ISSUE #1475: projectile-find-file not working for .gitignored directories despite them being unignored
projectile-globally-unignored-directories is a variable defined in ‘projectile.el’.
Its value is ("/vendor/")

(tried vendor/ and vendor too)

vendor is gitignored and I have tried to enable it as a search candidate by using the setting above. Indexing is off and indexing method is, if relevant, set to alien in case its used elsewhere for search algorithms.

So: how do I enable projectile to search gitgnored subdirectories. In this case I want it to search the symfony vendor subdirectory.

spaceotter:
When using hybrid mode and unignoring files that are ignored by git, the
settings in .projectile would not work. The arguments to internal
function projectile-get-repo-ignored-directory were in the wrong
order. So it would return nil instead of the desired list of git ignored
files.

~~The issue is caught by the projectile-add-unignored test.~~
Looks like I was mistaken and this needs a test.

-----------------


bbatsov:
Thanks!

PR #1486, opened 2020-01-10
Allow Projectile to spawn extra process buffers (shell/eshell/term/vterm/ielm) with the prefix ARG

freesteph:
Hey! Thanks for projectile. As a disclaimer I don't write very much Lisp (yet).

# Rationale
I need to fire multiple terminals in the projects I work on. One to run the server and follow the logs, maybe an interactive console and some other bits & bobs sometimes so add a mechanism to override the default behaviour and spawn new vterm buffers as needed.

# Summary 
Going to copy and paste my commit if that's OK:

>If `projectile-run-vterm` is called with a prefix ARG it will generate
> a new buffer in the style of `generate-new-buffer-name` (e.g '*vterm
> my-proj<2>*') and switch to it.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -775,6 +775,13 @@ just return nil."
   "Get the symbol at point and strip its properties."
   (substring-no-properties (or (thing-at-point 'symbol) "")))
 
+(defun projectile-generate-process-name (process make-new)
+  "Infer the buffer name for PROCESS or generate a new one if MAKE-NEW is true."
+  (let* ((project (projectile-ensure-project (projectile-project-root)))
+         (base-name (format "*%s %s*" process (projectile-project-name project))))
+    (if make-new
+        (generate-new-buffer-name base-name)
+      base-name)))
 
 
 ;;; Serialization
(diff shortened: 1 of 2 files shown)


bbatsov:
I'm fine with the proposed changed. It needs a changelog entry, through. 

One small problem is that this introduces a small inconsistency with other similar commands that don't support the prefix arg. Might be a good idea to unify the eshell, term and vterm commands.

projectile.el
```diff
@@ -3394,36 +3406,38 @@ Switch to the project specific ielm buffer if it already exists."
...
-      (let ((program (or program
-                         (read-from-minibuffer "Run program: "
-                                               (or explicit-shell-file-name
-                                                   (getenv "ESHELL")
-                                                   (getenv "SHELL")
-                                                   "/bin/sh")))))
+      (let ((program (read-from-minibuffer "Run program: " default-program)))
         (projectile-with-default-dir project
-          (set-buffer (make-term term program))
+          (set-buffer (term-ansi-make-term buffer-name program))
```
freesteph: https://github.com/emacs-mirror/emacs/blob/1ebe1605775c0f273957c1541d5b526bb70f7cfb/lisp/term.el#L4181

I also a needed something identical to `make-term` without tampering with my `buffer-name`.

projectile.el
```diff
@@ -775,6 +775,12 @@ just return nil."
   "Get the symbol at point and strip its properties."
   (substring-no-properties (or (thing-at-point 'symbol) "")))
 
+(defun projectile-generate-prompt-name (cmd make-new)
+  "If MAKE-NEW is true generate a new name for starting CMD.  Otherwise find the current one."
```
bbatsov: The second sentence is misleading/redundant.

projectile.el
```diff
@@ -775,6 +775,12 @@ just return nil."
...
+  (let* ((project (projectile-ensure-project (projectile-project-root)))
+         (base-name (format "*%s %s*" cmd (projectile-project-name project))))
+    (if make-new (generate-new-buffer-name base-name)
```
bbatsov: I'd put the `if` branch on a line of its own.

projectile.el
```diff
@@ -3360,32 +3366,38 @@ regular expression."
...
 ;;;###autoload
-(defun projectile-run-shell ()
+(defun projectile-run-shell (arg)
   "Invoke `shell' in the project's root.
 
-Switch to the project specific shell buffer if it already exists."
-  (interactive)
+Switch to the project specific shell buffer if it already exists.
+
+Accepts a prefix argument ARG to indicate creation of a new process instead."
```
bbatsov: Accepts -> Use

projectile.el
```diff
@@ -775,6 +775,12 @@ just return nil."
   "Get the symbol at point and strip its properties."
   (substring-no-properties (or (thing-at-point 'symbol) "")))
 
+(defun projectile-generate-prompt-name (cmd make-new)
```
bbatsov: I think `cmd` is a bit misleading here - it's actually just a basename. I'd also use `process/comint-buffer` instead of `prompt`.

projectile.el
```diff
@@ -3394,36 +3406,38 @@ Switch to the project specific ielm buffer if it already exists."
...
-         (buffer (concat "*" term "*")))
-    (unless (get-buffer buffer)
+Switch to the project specific term buffer if it already exists.
+
+Accepts a prefix argument ARG to indicate creation of a new process instead."
+  (interactive "P")
+  (let ((project (projectile-ensure-project (projectile-project-root)))
+        (buffer-name (projectile-generate-prompt-name "term" arg))
+        (default-program (or explicit-shell-file-name
+                                               (getenv "ESHELL")
```
bbatsov: The indentation here is broken.

bbatsov:
Overall your changes look good to me. I've added a bit of small feedback here and there. Once you address it I'll merge the PR.

projectile.el
freesteph: I went with `(defun projectile-generate-process-name (process make-new)` because I tried a bunch of different wordings and couldn't decide. Feel free to push a commit atop or let me know what you think reads best and I can also do the change.

bbatsov:
Thanks! 

P.S. Naming is hard! 😆

Issue #675, opened 2015-03-04
Indexing files (native) with a .projectile + a few patterns is extremely slow


sandos:
I tried with this .projectile file:

> -vx55-num01
> -vx67-cpu02
> -vx63-cpu01
> -670Debug
> -650Debug
> -bcout.cg
> -$tf
> -Debug

It even has no wildcards, but it is still so slow, with native indexing, in windows that I never got to see it finish. I am not able to debug this further, sadly. It looks, in procmon, like its reindexing the same folder over and over, likely trying to expand file patterns in the same folder which is crazy. It should not need to expand the pattern for every filename.

I tried slightly smaller example with the same configuration as above. The folder structure was:

> .projectile
> build.proj
> downloadAllTFS.bat
> tfstatus
> typetest/.tfignore
> typetest/DMR typetest SW.doc
> typetest/wintms/ABOUT.DFM

Logs from process monitor for this sample directory: https://gist.github.com/sandos/f19bca2d1d8ff0180665

tuhdo:
You should use alien. It will use external program to provide the file list. Much much faster.

sandos:
I actually tried that, with no luck. Did not investigate much though.

PR #1495, opened 2020-02-04
Optimize native indexing

shitikanth:
This is intended as a fix for #675. 

On my system, native indexing was taking a few seconds with projects with ~100 files and was completely unusable (>15 seconds) for larger projects with ~1000 files. Decided to run the indexing function under the profiler, and the function was spending most of the time simply inside calls to  `(projectile-ignored-files)` and `(projectile-ignored-directories)`. 

As far as I can tell, this is completely unnecessary as the output of these functions depends only on  `default-directory` and will not change during the entire indexing process. This patch creates a version of `projectile-indexing-method` where I store and reuse the results of these two functions.

**Results**
```
;; Before patch
(benchmark-run-compiled 5 (projectile-dir-files-native "~/src/projectile"))
(18.034280782 12 1.424345668998285)

;; After patch
(benchmark-run-compiled 5 (projectile-dir-files-native "~/src/projectile"))
(0.349568535 0 0.0)

;; Larger project with ~2000 files, unusably slow before the patch
(benchmark-run-compiled 5 (projectile-dir-files-native "~/src/folly"))
(0.9296409800000001 0 0.0)

```

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1181,6 +1181,9 @@ Files are returned as relative paths to DIRECTORY."
 The function calls itself recursively until all sub-directories
 have been indexed.  The PROGRESS-REPORTER is updated while the
 function is executing."
+  (projectile-index-directory--internal directory patterns progress-reporter (projectile-ignored-files) (projectile-ignored-directories)))
+
+(defun projectile-index-directory--internal (directory patterns progress-reporter ignored-files ignored-directories)
   (apply #'append
          (mapcar
           (lambda (f)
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -1181,6 +1181,9 @@ Files are returned as relative paths to DIRECTORY."
 The function calls itself recursively until all sub-directories
 have been indexed.  The PROGRESS-REPORTER is updated while the
 function is executing."
+  (projectile-index-directory--internal directory patterns progress-reporter (projectile-ignored-files) (projectile-ignored-directories)))
+
+(defun projectile-index-directory--internal (directory patterns progress-reporter ignored-files ignored-directories)
```
bbatsov: I think it'd be simpler to just let-bind the ignored files/folders and not introduce new helper functions.

bbatsov:
Great work! I've added a small remark. Probably we should also mention this change in the changelog.

bbatsov:
@shitikanth ping :-)

bbatsov:
@shitikanth ping 2 :-)

PR #1503, opened 2020-02-24
Leave archive before searching for project root

Stebalien:
When TRAMP archives are enabled, back out of the archive before trying to find
the project root. Starting inside the archive makes opening archives quite slow.

-----------------

  - Would require gvfs, as far as I know.
  - Don't pass locally, with or without this change.


bbatsov:
Thanks!

PR #1504, opened 2020-02-27
Use projectile-with-default-dir in projectile-run-vterm

dkellner:
This avoids "cd \<project root\>" as first line of the newly created terminal and is consistent with other functions, e.g. `projectile-run-term`.

I did not update `CHANGELOG.md` because this is a really small enhancement, even if it is user-visible.

-----------------

Thanks!


bbatsov:
Thanks!

PR #1505, opened 2020-02-28
[Fix #705] let tag generation be aware of ignore pattern.

### LINKED ISSUE #705: projectile-regenerate-tags does not ignore files in projectile-globally-ignored-file-suffixes
`projectile-regenerate-tags` appears to only exclude ignored directories. It does not appear to ignore ignored files.

My project has a lot of `.min.js` files that are being added to `TAGS`; creating a very bloated and slow `TAGS` file. I have added `.min.js` to `projectile-globally-ignored-file-suffixes`, but no luck excluding them.

Looking at `projectile-tags-exclude-patterns`, it looks like only directories are excluded.

zhongweiy:
let tag generation be aware of ignore pattern.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3276,7 +3276,9 @@ regular expression."
   "Return a string with exclude patterns for ctags."
   (mapconcat (lambda (pattern) (format "--exclude=\"%s\""
                                        (directory-file-name pattern)))
-             (projectile-ignored-directories-rel) " "))
+             (append
+              (projectile-ignored-directories-rel)
+              (projectile-patterns-to-ignore)) " "))
 
 ;;;###autoload
 (defun projectile-regenerate-tags ()


projectile.el
```diff
@@ -3276,7 +3276,9 @@ regular expression."
   "Return a string with exclude patterns for ctags."
   (mapconcat (lambda (pattern) (format "--exclude=\"%s\""
                                        (directory-file-name pattern)))
-             (projectile-ignored-directories-rel) " "))
+             (append
+              (projectile-ignored-directories-rel)
```
kini: Maybe you can also change this line as follows:

```suggestion
              (projectile-ignored-directories)
```

That will allow ignore paths containing multiple components to work, fixing another issue. I was going to make a PR for that but since it touches the same lines as this, might as well do it in this PR :)

The problem: Suppose we have a project at location `/project-root/` and it contains a directory `foo/bar`. Then writing `-/foo/bar` in the `.projectile` file should mean that that directory is skipped when generating tags. But actually it doesn't match because `ctags` follows the following steps:

1. Encounter directory `/project-root/foo` and check the pattern `foo/bar` against the full path `/project-root/foo` and the basename `foo`. It doesn't match either one, so `foo` is entered.
2. Encounter directory `/project-root/foo/bar` and check the pattern `foo/bar` against the full path `/project-root/foo/bar` and the basename `bar`. It doesn't match either one, so `bar` is entered.

kini: By the way, I think making the change I suggest will expose another issue with `projectile-ignored-directories`.

I guess that the `projectile-globally-ignored-directories` variable is supposed to include arbitrarily deep subdirectories with names like `.idea`, `.ensime_cache`, etc., but the return value of `projectile-ignored-directories` just contains the absolute paths `/project-root/.idea`,  `/project-root/.ensime_cache` etc., and doesn't include `/project-root/foo/.idea` for example. But before my suggested change of removing `-rel`, the tag generation would still skip `/project-root/foo/.idea` because the command line option `--exclude=".idea"` would be passed to `ctags` and this would match against `/project-root/foo/.idea`.

zhongweiy: kini, thanks for your review! Do you think combining "projectile-ignored-directories-rel" and "projectile-ignored-directories" will solve all issues?

kini: I would hesitate to make such a change since it would affect various other parts of the code. Hopefully someone who is more familiar with the whole codebase can chime in.

kini:
Bump.

bbatsov:
The API for ignored stuff should definitely be simplified at some point, as even I don't remember how things were supposed to work at this point. Sorry for taking so long to merge this - at first I wanted to take a closer look at the ignored stuff code, but I never found the time to do so.

PR #1507, opened 2020-03-07
[Fix #1456] Add check for file existence in replace-regexp

### LINKED ISSUE #1456: projectile-replace-regexp stops after encountering a missing file.
When using `projectile-replace-regexp`, it seems that if a given file being scanned is missing, the command just stops. Ideally, it just just skip the missing file and continue with the replace operation.

```
next-file: Opening input file: No such file or directory, /workplace/src/Package/src/com/company/product/Command.java
```

A-Wiedemann:
projectile-replace-regexp just stopped if a given file being scanned
was missing. The solution is to filter out non-existing files before
invoking tags-query-replace.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master (unreleased)
 
+* [#1456](https://github.com/bbatsov/projectile/issues/1456) Fix `projectile-replace-regexp` immediate stop after encountering a missing file.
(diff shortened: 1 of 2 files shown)


A-Wiedemann:
This is my very first contribution. Please let me know if I missed something out (:

bbatsov:
Your fix is correct, but I'm a bit worried about its performance impact in some huge project, as that's an expensive check. I'll have to think about this.

PR #1509, opened 2020-03-16
Add a compilation-dir for cmake projects.

Liorst4:
Added a compilation directory for cmake projects.
Now it builds the project under 'build' instead of in source.


bbatsov:
Thanks!

PR #1519, opened 2020-04-22
accommodate golang modules

dickmao:
The change in 65ba429, while good, don't necessarily allow projectile to recognize go modules (usually living in ~/go/pkg/mod) as bonafide projects.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -317,6 +317,7 @@ If variable `projectile-project-name' is non-nil, this function will not be used
     "configure.in"       ; autoconf old style
     "configure.ac"       ; autoconf new style
     "cscope.out"         ; cscope
+    "go.mod"             ; golang default package root as of 1.13
     )
   "A list of files considered to mark the root of a project.
 The topmost match has precedence."


bbatsov:
> don't necessarily allow projectile to recognize go modules (usually living in ~/go/pkg/mod) as bonafide projects.

Why so? I see we already check about this file in the function-based project detection. As far as projectile is concerned the outcome should be the same.

dickmao:
For `projectile-root-local` to pick up a golang module directory,  `projectile-project-root-files` needs "go.mod".  It'd be nice on users to have it there beforehand lest the user has to customize it himself.

```
emacs -Q --batch -f package-initialize --eval "(require 'projectile)" \
    --eval "(defun test (bogey) (let ((module \"github.com/sirupsen/logrus@v1.4.2\")
(projectile-project-root-files (if bogey (list bogey) projectile-project-root-files))
(projectile-known-projects-file (make-temp-file \"projectile-bookmarks\" nil \".eld\" \"()\"))) \
  (shell-command (format \"GOPATH=/var/tmp GO111MODULE=on go get %s\" module)) \
  (projectile-mode)
  (find-file-existing (expand-file-name \"go.mod\" (format \"/var/tmp/pkg/mod/%s\" module)))
  (kill-buffer)
  (if (member (file-name-as-directory (format \"/var/tmp/pkg/mod/%s\" module)) projectile-known-projects) (print 'good) (print 'bad))))" \
    --eval "(test nil)" \
    --eval "(test \"go.mod\")"
```

should print "bad" when we use the defaults, and "good" when we explicitly add "go.mod"

bbatsov:
Got it. Thanks!

PR #1521, opened 2020-04-26
Fix typos

tarsius:
Fix two very simple typos that were found using the `codespell` tool.

- [x] The default checkboxes are quite irrelevant in this case.


bbatsov:
Thanks!

Issue #1522, opened 2020-04-27
consider allowing comments in .projectile file


pipehat:
I'd like to be able to add comments in my .projectile file to clarify what some of the ignore patterns are for. Similar to .gitignore. Perhaps, e.g.:

```
-- .projectile ----------------------
# ignore emacs auto save stuff
-*~
-.#
```

Looks like this could be done with another pcase clause in projectile-parse-dirconfig-file.

bbatsov:
Yeah, I think that's a nice idea and something that'd be pretty easy to implement.

PR #1526, opened 2020-05-03
Support for optional comments in .projectile files by customizing projectile-dirconfig-comment-prefix.

pipehat:
Resolves bbatsov/projectile#1522

Setting the customizable `projectile-dirconfig-comment-prefix` variable to a character, e.g. `?#`, will cause lines in the .projectile which start with that character to be skipped over instead interpreting them as paths.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -358,6 +358,13 @@ containing a root file."
   :group 'projectile
   :type '(repeat function))
 
+(defcustom projectile-dirconfig-comment-prefix
+  nil
+  "If specified, starting a line in a project's .projectile file with this character marks that line as a comment instead of a pattern. Similar to '#' in .gitignore files."
+  :group 'projectile
+  :type 'character
+  :package-version '(projectile . "2.2.0"))
+
 (defcustom projectile-globally-ignored-files
   (list projectile-tags-file-name)
   "A list of files globally ignored by projectile."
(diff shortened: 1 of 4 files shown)


projectile.el
```diff
@@ -358,6 +358,12 @@ containing a root file."
   :group 'projectile
   :type '(repeat function))
 
+(defcustom projectile-dirconfig-comment-prefix
+  nil
```
bbatsov, line 362: I guess we should set this to something by default.

projectile.el
```diff
@@ -358,6 +358,12 @@ containing a root file."
...
+  "If specified, starting a line in a project's .projectile file with this character marks that line as a comment instead of a pattern. Similar to '#' in .gitignore files."
+  :group 'projectile
+  :type 'character)
```
bbatsov, line 365: Don't forget to add the package-version property here.

projectile.el
pipehat: Hmm I thought it should default to nil (that is, disabled) unless a user sets it themselves. To avoid surprising someone with existing ignore directives. Eg if we made this var ‘#’ by default, someone with an existing line in their projectile file like:

    #*

... would suddenly lose that ignore functionality.

projectile.el
pipehat: done!

projectile.el
bbatsov: Yeah, that makes sense.

bbatsov:
Thanks!

PR #1528, opened 2020-05-06
Improve performance of native indexing.

### LINKED ISSUE #675: Indexing files (native) with a .projectile + a few patterns is extremely slow

pipehat:
Originally reported under bug #675, native indexing for non-trivial
projects is extremely slow due to repeated invocations of
(projectile-ignored-[files|directories]) during recursion.

@shitikanth proposed a solution under #1495 in Feb 2020.

This PR incorporates @bbatsov's feedback on #1495 and may
resolve #1495 and close #675.

Thanks to @shitikanth for initial research and solution idea.

Benchmarking suggests at least a 10x improvement.
e.g.: indexing the projectile source project itself under linux:

(benchmark-run 3 (projectile-dir-files-native default-directory))

OLD: (2.109717349 25 0.5421294649999999)
NEW: (0.098484531 1 0.022120505000000013)

The effect was even more pronounced for me on Windows.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 
 ### Bugs fixed
 
+* [#675](https://github.com/bbatsov/projectile/issues/675): Performance improvement for native project indexing strategy.
 * [#97](https://github.com/bbatsov/projectile/issues/97): Respect `.projectile` ignores which are paths to files and patterns when using `projectile-grep`.
 * [#1391](https://github.com/bbatsov/projectile/issues/1391): A `.cabal` sub-directory is no longer considered project indicator.
 * [#1385](https://github.com/bbatsov/projectile/issues/1385): Update `projectile-replace` for Emacs 27.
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks for driving this to the finish line!

PR #1530, opened 2020-05-13
Fix whitespace

tarsius:
Fix a small whitespace error, which causes `codespell` to suggest that `files'` should be replaced `file's`.


bbatsov:
Thanks!

Issue #1523, opened 2020-04-27
Is there any way to limit the number of open files per project?


jkaye2012:
This is a feature that I would personally find very useful. I wanted to ask about it here because if this does not exist (I have not been able to find any documentation to that effect), this is useful enough to me that I would look into submitting a PR to add the feature if the maintainers think it would be useful as well.

Basically, I just want to be able to set a limit per project on how many files (buffers) Emacs will keep open for that project. So if I have 10 files open with a limit of 10, and I open an 11th, Emacs will automatically close the least-recently-used file in the project. Projectile seems like the right place to me because it already knows about projects and relates buffers to their corresponding projects.

Please let me know your thoughts on this, if it exists, and if not if it is worth submitting a PR for.

Thanks!

jkaye2012:
I guess it is also possible for me to create and distribute an extension package, so whether that is better than including the feature in core Projectile is probably a discussion worth having as well.

bbatsov:
Building something like this would be pretty simple, as it can either be a check triggered when using `projectile-find-file` or a hook into the default `find-file`.

pipehat:
I wonder if auto-closing old buffers could be frustrating or confusing in some cases.

A variation to consider: perhaps 'pin' buffers open instead of an open-buffer limit.

It's already possible to close all project buffers (projectile-kill-buffers). What if we just make it easy to toggle whether a given buffer should survive that process.?

e.g., with a project buffer open (and perhaps 20 more in the background):
M-x projectile-pin-buffer            toggles buffer's pin state
M-x projectile-kill-buffers           kills unpinned project buffers
C-u M-x projectile-kill-buffers    kills ALL project buffers, even if pinned

jkaye2012:
It's not about keeping them open, it's specifically about auto-closing
them. When I have time, I'll add this as a strictly opt-in feature. I don't
think killing buffers is something that should be done by default, to your
point.

bbatsov:
Fixed in #1532.

PR #1532, opened 2020-05-15
Adding support for limiting project buffer count

palikar:
I've added two new customs that can be used to control the feature suggested in #1523. The check if the current buffer count is cover the cap is performed every time the `find-file` hook is executed *and* projectile mode is active. 

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -737,13 +737,10 @@ position."
           (const :tag "Move to end" move-to-end)
           (const :tag "Keep" keep)))
 
-(defcustom projectile-limit-project-buffer-count nil
-  "If non-nil, the buffer count for a particular project will be capped."
-  :group 'projectile
-  :type 'boolean)
+(defcustom projectile-max-buffer-count nil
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -729,6 +729,16 @@ position."
           (const :tag "Move to end" move-to-end)
           (const :tag "Keep" keep)))
 
+(defcustom projectile-limit-project-buffer-count nil
```
bbatsov, line 732: I think we don't really need this option. If the other one is `nil` we can consider this as no limit.

palikar: Sure, I just pushed a version without this custom.

bbatsov:
You might also mention this somewhere in the docs and you'll also have to rebase on top of `master`.

palikar:
@bbatsov  I was just adding documentation but you merged the PR before I can push the change :smiley:

bbatsov:
My bad. You can always add it in follow-up PR.

PR #1533, opened 2020-05-17
[Fix #1465] Add Bazel WORKSPACE detection to Projectile

### LINKED ISSUE #1465: Please add "WORKSPACE" as a project marker
Could you please add "WORKSPACE" files as a project marker?
The Bazel build system uses this file to indicate the root directory of a
project.

Thank you for an awesome tool! :)

[Bazel Usage](https://docs.bazel.build/versions/master/build-ref.html)

jin:
This PR adds a new projectile project type for the Bazel build system. It's multi-language, multi-platform, so I've placed it under the "Universal" category. Its marker file is `WORKSPACE`, and commands for building, testing and running are `bazel build`, `bazel test` and `bazel run` respectively.

I've tried to run `make test`, and there's one test that's failing on both master and my branch. I'm not sure if it's related to my change.

Fixes https://github.com/bbatsov/projectile/issues/1465

```
projectile-project-root ensures that we update the cache if it's expired

Traceback (most recent call last):
  (let ((default-directory sandbox)) (progn (make-directory "project/" t) (l...
  (progn (make-directory "project/" t) (let ((temp-file "project/.projectile...
  (let ((projectile-projects-cache (make-hash-table :test (function equal)))...
  (let ((--dolist-tail-- (quote ("./file1.el" "./file2.el"))) f) (while --do...
  (while --dolist-tail-- (setq f (car --dolist-tail--)) (buttercup-expect (f...
  (buttercup-expect (lambda nil (quote (member f (gethash (projectile-projec...
  (buttercup-fail "%s" "Expected `(member f (gethash (projectile-project-roo...
  (signal buttercup-failed "Expected `(member f (gethash (projectile-project...
FAILED: Expected `(member f (gethash (projectile-project-root) projectile-projects-cache))' to be non-nil, but instead it was nil.
```

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Optional support for comments in .projectile dirconfig files using `projectile-dirconfig-comment-prefix`.
 * [#1497](https://github.com/bbatsov/projectile/pull/1497): New command `projectile-run-gdb` (<kbd>x g</kbd> in `projectile-command-map`).
+* Add [Bazel](https://bazel.build) project type.
 
 ### Bugs fixed
 
--- a/doc/projects.md
+++ b/doc/projects.md
@@ -50,6 +50,7 @@ configure.in      | autoconf old style
 configure.ac      | autoconf new style
 cscope.out        | cscope
 Makefile          | Make
+WORKSPACE         | Bazel workspace file
 
 There's also Projectile's own `.projectile` which serves both as a project marker
 and a configuration file. We'll talk more about later in this section.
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

jinnovation:
As it turns out, I just so happened to have a `workspace/` directory in my home directory that, by virtue of the macOS file system being case-insensitive and the fact that `projectile-verify-file` doesn't check that the file *isn't* a directory, caused my home directory to be recognized as a Bazel project, which it clearly is not.

Maybe it's worth adding such a check to `projectile-verify-file` et al to catch such cases? I think the fact that the marker file name in this case is `WORKSPACE` makes such collisions much more likely; I imagine it's far less likely to have a directory called, say, `setup.py`. 🤔

Issue #323, opened 2014-04-20
Projectile should use vc-root-dir in Emacs >= 24.5


dcolascione:
I've added a new vc-root-dir vc function to Emacs trunk, which will become 24.5. This function is a new generic interface for finding the root directory of a VC tree, and it'd be nice if Projectile could use it instead of (or at least in addition to) its hardcoded list of VCS root files and directories.

bbatsov:
I'll close this ticket, as now that's a trivial matter of tweaking `projectile-project-root-functions` - https://docs.projectile.mx/projectile/projects.html#customizing-project-detection

PR #1536, opened 2020-05-19
Usage of vc root dir

palikar:
I saw that  #323 was marked high-priority but the support for `vc-root-dir` was still not implemented.

I've added an extra projectile-root-files-function that works only if the directory is the current default directory  (i.e if the requested project-root is of the current project). It uses the suggested in #323 `vc-root-dir` and should cover the majority of the uses cases of `projectile-project-root`

I've also added documentation for the feature introduced in #1532.

-------------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -351,6 +351,7 @@ containing a root file."
 
 (defcustom projectile-project-root-files-functions
   '(projectile-root-local
+    projectile-root-vc-root
     projectile-root-bottom-up
     projectile-root-top-down
     projectile-root-top-down-recurring)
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks for tackling this! You'll have to rebase on top of the current `master`.

projectile.el
```diff
@@ -1069,6 +1070,12 @@ topmost sequence of matched directories.  Nil otherwise."
                  (not (projectile-file-exists-p (expand-file-name f (projectile-parent dir)))))))))
    (or list projectile-project-root-files-top-down-recurring)))
 
+(defun projectile-root-vc-root (dir)
+  "Retrieve the root directory of the project at DIR using 'vc-root-dir'.
+Works only if DIR is the current default directory"
+  (when (file-equal-p dir default-directory)
```
bbatsov, line 1076: I don't get why this check is needed.

bbatsov:
I'll close this ticket for now, as I'm doing some other related work in this area. In the mean time - I've added this to the docs  https://docs.projectile.mx/projectile/projects.html#customizing-project-detection

Issue #1227, opened 2018-03-21
`projectile-replace` does not respect `.projectile` file


matthew-piziak:
## Expected behavior

I have `- *~` in my `.projectile` file. `projectile-find-file` correctly ignores my emacs backup files ending with `~`. I expect that `projectile-replace` follows the same behavior, and ignores backup files.

## Actual behavior

Projectile hops into my backup files, I don't notice, and then when I hit `y` it tells me that the file is readonly (true) and kills my replace session (frustrating).

## Steps to reproduce the problem

Copy a file in your project and append a `~`. Add `- *~` to your `.projectile` file. Notice that `projectile-find-file` (With prefix argument, to clear cache) ignores these files. However, `projectile-replace` will still enter them.

### Projectile version information

```
Projectile version: 20180227.1743
```

### Emacs version

GNU Emacs 27.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.26) of 2018-02-12

### Operating system

Arch Linux

matthew-piziak:
bump

stoeckley:
I've noticed this as well and came here to open an issue about this, but see I don't have to. I use `projectile-grep` and `projectile-replace` all the time. The former respects the .projectile file, the latter does not. Why is that?

stoeckley:
That's a shame, and I'm sorry to hear that Mr. Bot.

matthew-piziak:
In our age robots perform all actions, but we humans are hired to do nothing.

cammil:
Is this getting looked into? I have the same issue

bbatsov:
> The former respects the .projectile file, the latter does not. Why is that?

Because `projectile-replace` doesn't use any file filtering for performance reasons - it just finds all files that contain the replacement string in the project directory. That's fairly easy to change, but will make the replacement slow on very big projects. See `projectile-files-from-cmd` and you'll understand what I'm talking about.

matthew-piziak:
That's illuminating, thank you. But I'm having trouble thinking of the fundamental limitation. Are grep/rg/etc. slower when you ask them to filter files?

wkenyon:
Hey, just bumping this, the behaviour seems strange as it tries to replace instances in large binary files that I don't want it to look into, this either takes ages or crashes emacs.

PR #1537, opened 2020-05-19
Fixing projectile-replace to respect .projectile

palikar:
#1227 is a long standing issue and I've decided to give it a go. Previously the commands that are finding files with a certain pattern in them were ran on the entire project. Now I've modified the commands slightly so that they only check the files that are not ignored because of the .projectile file.

I am not sure if this is the best way to implement this. Feedback and other opinions will be welcomed.

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3487,21 +3487,23 @@ are impossible (for instance on Windows), returns a list of all
 files in the project."
   (if (projectile-unixy-system-p)
       (let* ((search-term (shell-quote-argument string))
+             (files (projectile-project-files directory))
+             (files-string (mapconcat 'identity files " "))
              (cmd (cond ((executable-find "ag")
-                         (concat "ag --literal --nocolor --noheading -l -- "
-                                 search-term))
+                         (format "ag --literal --nocolor --noheading -l -- %s %s"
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -3487,21 +3487,23 @@ are impossible (for instance on Windows), returns a list of all
 files in the project."
   (if (projectile-unixy-system-p)
       (let* ((search-term (shell-quote-argument string))
+             (files (projectile-project-files directory))
```
bbatsov, line 3490: I'm reasonably sure that's going to end very badly in a huge project. :-)

bbatsov:
I think it'd better to try to pass some ignores/includes to ag, then all the project files explicitly, as fetching the list, stitching it together and breaking it apart will probably rather slow.

orlandini:
Any updates/possible workarounds? It would be really useful to have it fixed but my lisp knowledge is quite limited.

bbatsov:
I think a better solution would be pass the ignored stuff to the various tools as parameters, but this will be more involved in terms of code. The proposed solution will work fairly well on small projects, but will be extremely slow on something big.

PR #1539, opened 2020-05-23
Add a custom setting to control automatic project discovery

fishyfriend:
This PR adds a custom setting `projectile-auto-discover` that controls whether to automatically run `projectile-discover-projects-in-search-path` when Projectile mode is activated.

The main purpose is to enable faster Emacs startup when starting `projectile-mode` automatically from the init file. On my system (~2015 i5, Emacs 27, Linux, with 140 project directories) turning off auto-discovery improves startup time by around 1.7 seconds.

The default setting for `projectile-auto-discover` leaves current behavior unchanged. 

One already has to manually run one of the `projectile-discover-*` commands to update project bookmarks during an Emacs session, so it seems reasonable to allow the user to choose to always use manual updating rather than forcing an update every time they start Emacs or turn on the mode. Since project bookmarks are cached and loaded from disk, it's usually not necessary to update them anyway.

The following tests are failing on master on my machine so I don't think the failures are related to my changes. All other tests are passing.
* `"projectile-grep multi-root grep grep multi-root projects"`
* `"projectile-compilation-dir should not fail on bad compilation dir config"`

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Optional support for comments in .projectile dirconfig files using `projectile-dirconfig-comment-prefix`.
 * [#1497](https://github.com/bbatsov/projectile/pull/1497): New command `projectile-run-gdb` (<kbd>x g</kbd> in `projectile-command-map`).
 * Add [Bazel](https://bazel.build) project type.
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #1545, opened 2020-06-15
Fix projectile-detect-project-type for the case when type's marker is a function

doublep:
It is possible to pass a function instead of a list of marker files to `projectile-register-project-type`. However, if said function is a lambda, `projectile-detect-project-type` will later die with an exception, because it tries to handle it as a list of strings: lambdas (or closures) are also `listp`.

This simple change fixes this. I didn't add any tests, looks too minor to me.

-----------------

Thanks!


bbatsov:
Nice catch!

PR #1547, opened 2020-06-15
Add emacs-eldev project type

doublep:
[Eldev](https://github.com/doublep/eldev) is a build tool for Elisp projects, comparable to Cask. Normally I would create a separate package (like I did for [Flycheck](https://github.com/flycheck/flycheck-eldev)) for integrating it, but since this is only a few lines, I feel it would be an overkill.

Eldev is virtually unused currently, but this is chicken-and-egg problem: if I integrate it in more places, *maybe* it will get more users as a side-effect.

-----------------

Thanks!


doublep:
> I'd suggest using for the config a name like eldev.el or whatever

It's a bit too late to change. There are a few users and I personally would also hate renaming it in ~10 my own projects and especially rewriting all the documentation.

Issue #1554, opened 2020-06-27
grep and ack search options return error for 'more complex' search-term (GNU/linux)


dalanicolai:
On GNU/linux (Fedora 32, ack version 3.3.1), when trying to replace the string `xpath=//ul[@id="input-internetpinnen_prefix_subscriptions"]//div[contains(text(), "KLM")][last()]` using `projectile-replace`, the grep and ack options in the `projectile-files-with-string` function return the following errors.

(The 'ag' option works fine)

## Expected behavior
Succesful replacement of text (like when using the `ag` option in the `projectile-files-with-string` function)

## Actual behavior
The grep function returns the following error:
```
("/home/dalanicolai/.emacs.d/elpa/26.3/develop/grep: Invalid range end")
```
The ack function returns the error:
```
("/home/dalanicolai/.emacs.d/elpa/26.3/develop/ack: Options '--heading' and '-l' can't be used together.")
```

## Steps to reproduce the problem
- make sure "the silver searcher" is not installed (i.e. there is no executable "ag" in your path)
- evaluate `(projectile-files-with-string "xpath=//ul[@id=\"input-internetpinnen_prefix_subscriptions\"]//div[contains(text(), \"KLM\")][last()]" directory)`, replacing directory with some valid directory path on your system
OR
run `projectile-replace` trying to replace `xpath=//ul[@id=\"input-internetpinnen_prefix_subscriptions\"]//div[contains(text(), \"KLM\")][last()]`

## Environment & Version information
ack version 3.3.1

### Projectile version information
```
Projectile 20200616.1659
```
### Emacs version
26.3

### Operating system
GNU/linux (Fedora 32)

dalanicolai:
Ack version 2.2.1 seems to work fine at the moment (current version of ack on Ubuntu 18.04). For ack version 3.3.1 removing the `--noheading` flag fixes the issue as the `-l` flag automatically adds/implies the `--noheading` flag. Because a new LTS release of Ubuntu (20.04) is already available and ships ack version 3.3.1, I guess the `--noheading` flag can be removed immediately or otherwise very soon

bbatsov:
> make sure "the silver searcher" is not installed (i.e. there is no executable "ag" in your path)

I've been using only `ag` for years now. :-) 

I'll have to check what's up with `grep` - last time I checked this was working fine, but I was on macOS back then. Now I'm also on Ubuntu 20.04. 

>  guess the --noheading flag can be removed immediately or otherwise very soon

[...]

PR #1555, opened 2020-06-27
Add ripgrep search option

dalanicolai:
Added as first option, like in [spacemacs](https://develop.spacemacs.org/doc/DOCUMENTATION.html#searching) and [visual studio code](https://code.visualstudio.com/updates/v1_11#_text-search-improvements).

Add ripgrep search option to the `projectile-files-with-string` function

I will check all boxes below (although I did not really check all these things as I consider it not very relevant here), as this is a minor addition that is succesfully tested on Fedora 32.
Also it is probably useful to mention here that I actually implemented this option as a fix to issue  #1554 (which describes that the grep and ack search options return errors here) 

- [ x] The commits are consistent with our [contribution guidelines](../blob/master/CONTRIBUTING.md)
- [ x] You've added tests (if possible) to cover your change(s)
- [ x] All tests are passing (`make test`)
- [ x] The new code is not generating bytecode or `M-x checkdoc` warnings
- [ x] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)
- [ x] You've updated the readme (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [#1539](https://github.com/bbatsov/projectile/pull/1539): New defcustom `projectile-auto-discover` controlling whether to automatically discover projects in the search path when `projectile-mode` activates.
 * Add [emacs-eldev](https://github.com/doublep/eldev) project type.
 * Add Dart project type.
+* [#1555](https://github.com/bbatsov/projectile/pull/1555) Add search with ripgrep. 
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Looks good. Don't forget to update the changelog as well.

dalanicolai:
Was not sure if this counted as user-visible functionality. Anyway, I've updated the changelog as well now.

isker:
The docstring of the function should be updated too.

bbatsov:
Good point!

bbatsov:
Thanks!

PR #1563, opened 2020-08-06
Add projectile-acquire-root to replace common (projectile-ensure-project (projectile-project-root)) idiom; add function to do action on every buffer in project with that buffer current

nbfalcon:
The idiom (projectile-ensure-project (projectile-project-root)) is very common, and as such I thought it prudent to add a function to replace it, as that could lead to better code. The function projectile-acquire-root was added to do so.

There is a function to run some action on every project buffer (projectile-process-current-project-buffers). However, some functions (e.g. hack-dir-local-variables-non-file-buffer) that cannot take a buffer as an argument. To support executing such functions on all project buffers, I added projectile-process-current-project-buffers-current. The difference between it and its sibling projectile-process-current-project-buffers is that my function passes no arguments, but instead makes the project buffer current as it iterates over all project buffers. A test case was added to ensure that its behaviour is similar to its sibling, yielding the same set of buffers in the same order.
-----------------

- [x ] The new code is not generating bytecode or `M-x checkdoc` warnings

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1125,9 +1125,9 @@ Controlled by `projectile-require-project-root'."
      (t default-directory))))
 
 (defun projectile-acquire-root (&optional dir)
-  "Finds the current project's root, and if that fails, prompts
-the user to supply it. Provides the common
-idiom (projectile-ensure-root (projectile-project-root))"
+  "Find the current project root, and prompts the user for it if that fails.
+Provides the common idiom (projectile-ensure-root (projectile-project-root)).
+Starts the search for the project with DIR."
   (projectile-ensure-project (projectile-project-root dir)))
 
 (defun projectile-project-p (&optional dir)
@@ -1445,8 +1445,8 @@ If PROJECT is not specified the command acts on the current project."
       (funcall action buffer))))
 
 (defun projectile-process-current-project-buffers-current (action)
-  "Invoke ACTION without arguments on every project buffer with
-that buffer current."
+  "Invoke ACTION on every project buffer with that buffer current.
+ACTION is called without arguments."
   (let ((project-buffers (projectile-project-buffers)))
     (dolist (buffer project-buffers)
       (with-current-buffer buffer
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -1438,6 +1444,14 @@ If PROJECT is not specified the command acts on the current project."
     (dolist (buffer project-buffers)
       (funcall action buffer))))
 
+(defun projectile-process-current-project-buffers-current (action)
```
bbatsov, line 1447: I think you probably didn't want to add the second "current". :-)

projectile.el
```diff
@@ -1124,6 +1124,12 @@ Controlled by `projectile-require-project-root'."
      (projectile-require-project-root (error "Projectile can't find a project definition in %s" dir))
      (t default-directory))))
 
+(defun projectile-acquire-root (&optional dir)
```
bbatsov, line 1127: I've meant to add a similar function when I added `projectile-ensure-project`, but I'm not sure about the name, as I don't think `acquire` is the best word we can use here.

projectile.el
bbatsov: Ah, I read first the code and then your explanation - I get the point of the function, but I still think this name is quite confusing.

projectile.el
nbfalcon: Again, you can propose a new name if you want. However, I would argue that "acquire" is fitting here, as it in my opinion has that "ensuring" character. From Merriam Webster (https://www.merriam-webster.com/dictionary/acquire), 1b: "to come into posession by unspecified means". I think this is fitting as you really always get the root (the user is prompted otherwise), and the unspecified means because there are two ways with which the root could be acquired, and due to the unspecified nature new ones could be added later perhaps?

projectile.el
nbfalcon: Hi! I agree that the name is a bit off. However, projectile-process-current-project-buffers is already taken, so I had to add the second current. I wanted to base the name on projectile-process-current-project-buffers, because the behaviour of my function and that one should be similar. You can of course propose a better name, and I would be happy to use that instead.

PR #1567, opened 2020-08-18
Update eldev-emacs project type with recently added features

### LINKED ISSUE #1560: Support for packaging and install project commands
CMake workflow can include install and package commands (with CPack), in addition to build, test, etc. 

Would it be possible to add support for these commands to projectile?

I have tried to add them [here](https://github.com/jehelset/projectile/tree/projectile-install-and-packages-commands) similar to the other commands.

doublep:
Update eldev-emacs project type for:
7ed2ce3 * [Fix #1560] Add install and package project commands
95bcc7d * Add a `:project-file` param to `projectile-register-project-type`

- [ ] You've added tests (if possible) to cover your change(s)  (**not needed**)
- [x] All tests are passing (`make test`)  (**2 tests fail, but the same as without the changes**)
- [ ] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)  (**not needed**)
- [ ] You've updated the readme (if adding/changing user-visible functionality)  (**not needed**)

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2808,9 +2808,11 @@ test/impl/other files as below:
                                   :test-suffix "-test")
 (projectile-register-project-type 'emacs-eldev (lambda () (or (projectile-verify-file "Eldev")
                                                               (projectile-verify-file "Eldev-local")))
+                                  :project-file "Eldev"
                                   :compile "eldev package"
                                   :test "eldev test"
-                                  :run "eldev emacs")
+                                  :run "eldev emacs"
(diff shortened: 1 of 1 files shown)


doublep:
Actually, no tests are failing, it's only because I have "Eldev-local" in my working directory.

bbatsov:
Thanks! At some point we should update Projectile itself to use Eldev as well. Feel free to tackle this if you wish to.

PR #1568, opened 2020-08-19
Use Eldev for development

doublep:
As requested in #1567. File `Eldev` is trivial in this case. Other changes:

* delete `Cask` and `Makefile`;
* adjust `.gitignore`, `.travis.yml`, `doc/modules/ROOT/pages/contributing.adoc` accordingly;
* adjust the two tests that check Projectile's own project type;
* fix a couple of byte-compilation warnings;
* delete redundant test helper files.

By the way, as far as I see CI for the project is disabled currently. Any reason for that?


bbatsov:
Thanks!

I see the builds are happening, but I'm not sure why there's no status reported on the PR - https://travis-ci.org/github/bbatsov/projectile/jobs/719427117

At some point I'll have to replace Travis with Circle or GHA.

PR #1569, opened 2020-08-20
Fix missing quote

dakra:
[This commit](https://github.com/bbatsov/projectile/commit/41461d9d3a41953d55dc03262cfe6eb0e71ea645) has unbalances quotes and breaks projectile.


bbatsov:
Thanks!

PR #1572, opened 2020-08-20
Switch over continuous integration from Travis to GitHub workflows

doublep:
As mentioned in #1568, there is a desire to ditch Travis.

If you rename the workflow, don't forget to adjust the URLs in `README.md`.

Changes in `projectile.el`:
- silence a byte-compilation warning issued on a function call that never happens on Emacs 27+ anyway;
- silence another warning; but this one I guess should rather be fixed using `fboundp`, like in `projectile-replace`; this I will leave to you.


bbatsov:
Great work! :bowing_man:

Issue #1577, opened 2020-08-28
projectile should not add my home directory to known project list


dackon:
Projectile added my home directory to the project list, I checked the [file markers](https://docs.projectile.mx/projectile/projects.html), not one in my home directory. But I have a folder 'Workspace' in my home directory, M-x projectile-project-info shows:

```terminal
Project dir: /Users/foobar/ ## Project VCS: none ## Project type: bazel
```

So I customized the variable and delete the WORKSPACE from the variable:
```lisp
(custom-set-variables
  '(projectile-project-root-files
    '("pubspec.yaml" "info.rkt" "Cargo.toml" "stack.yaml" "DESCRIPTION" "Eldev" "Cask" "shard.yml" "Gemfile" ".bloop" "deps.edn" "build.boot" "project.clj" "build.sbt" "application.properties" "gradlew" "build.gradle" "pom.xml" "poetry.lock" "Pipfile" "tox.ini" "setup.py" "requirements.txt" "manage.py" "angular.json" "package.json" "gulpfile.js" "Gruntfile.js" "mix.exs" "rebar.config" "composer.json" "CMakeLists.txt" "Makefile" "default.nix" "meson.build" "SConstruct" "GTAGS" "TAGS" "configure.ac" "configure.in" "cscope.out")))
```
 
## Expected behavior
Now, projectile should not recognize my home folder as a bazel project

## Actual behavior
projectile still recognize my home folder as a bazel project

### Projectile version information
```
Projectile 20200822.822
```

### Emacs version
GNU Emacs 27.1 (build 1, x86_64-apple-darwin18.7.0, NS appkit-1671.60 Version 10.14.6 (Build 18G95)) of 2020-08-12

### Operating system

macOS 10.15.6

dackon:
I deleted these lines from projectile.el, and projectile works as expected.

```lisp
(projectile-register-project-type 'bazel '("WORKSPACE")
                                  :project-file "WORKSPACE"
                                  :compile "bazel build"
                                  :test "bazel test"
                                  :run "bazel run")
```

bbatsov:
I'll close this ticket, as there's the workaround to you mentioned, plus one can just adjust `projectile-project-root-files` and remove `WORKSPACE` from it. As no one else reported such issue, I'll assume it's rather uncommon. If only they chose a better name for those `bazel` project descriptors...

Be-ngt-oH:
[...]

PR #1580, opened 2020-09-08
honor projectile-project-root-files

dackon:
Do not add project which not defined by projectile-project-root-files. If you want to add new projects, append project-file to  projectile-project-root-files firstly.

Fix this [issue](https://github.com/bbatsov/projectile/issues/1577)

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2550,8 +2550,6 @@ test/impl/other files as below:
     ;; explicit argument of nil and an omitted argument. However, the
     ;; body of the function is free to consider nil an abbreviation
     ;; for some other meaningful value
-    (when (and project-file (not (member project-file projectile-project-root-files)))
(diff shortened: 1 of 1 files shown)


bbatsov:
I don't really think you're fixing anything like this. As a matter of fact - it should be the other way around. Namely we should be adding project root markers only via project type definitions instead of dumping them directly in the list. If some project marker conflicts with something in your file structure obviously it's fine to adjust the list of markers manually, but I don't think your suggestion fixes anything.

dackon:
Sorry, I really cannot find a good way to do this. If I change my workspace name, it will break tons of things I compiled manually. Or I just have to delete the bazel from projectile.el every time when a new version released. By the way, I don't know lisp too much.

PR #1584, opened 2020-09-13
[Fix #1377] Modify tags generation to use directory file name

### LINKED ISSUE #1377: Trailing slash in project dir (not exclude) causes ctags to break
# Summary

I think the trailing slash problem mentioned in #207 for excludes is also causing a problem for me with the project directory. I think ctags version Exuberant Ctags 5.8 on cygwin does stupid things if the target directory has a trailing slash.

I fixed this by changing `default-directory` in the `projectile-regenerate-tags` function in `projectile.el` to be `(string-remove-suffix "/" default-directory)`.

If this makes sense to you, I can submit a pull-request.

## Expected behavior

projectile-regenerate-tags produces a tags file

## Actual behavior

TAGS file is empty

Run projectile-regenerate-tags on emacs installed on windows system with cygwin and ctags 5.8 installed.

Projectile 20190126.1117

### Emacs version

GNU Emacs 25.3.1 (x86_64-w64-mingw32) of 2017-09-12 with ctags 5.8

### Operating system

Windows 10

claritystorm:
[#1377](https://github.com/bbatsov/projectile/issues/1377): Fix `projectile-regenerate-tags` directory.

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3427,7 +3427,11 @@ regular expression."
            (tags-exclude (projectile-tags-exclude-patterns))
            (default-directory project-root)
            (tags-file (expand-file-name projectile-tags-file-name))
-           (command (format projectile-tags-command tags-file tags-exclude default-directory))
+           (command (format projectile-tags-command
+                            tags-file
+                            tags-exclude
+                            ;; Use directory file name for MSYS2 compatibility.
+                            (directory-file-name default-directory)))
            shell-output exit-code)
       (with-temp-buffer
         (setq exit-code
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -3427,7 +3427,10 @@ regular expression."
            (tags-exclude (projectile-tags-exclude-patterns))
            (default-directory project-root)
            (tags-file (expand-file-name projectile-tags-file-name))
-           (command (format projectile-tags-command tags-file tags-exclude default-directory))
+           (command (format projectile-tags-command
+                            tags-file
+                            tags-exclude
+                            (directory-file-name default-directory)))
```
bbatsov, line 3433: You should add some comment why this is needed here, otherwise someone might optimize it away. :-)

claritystorm:
A comment has been added.

bbatsov:
Thanks!

PR #1585, opened 2020-09-14
Add more OCaml extensions to projectile-other-file-alist

leungbk:
From
https://github.com/ocaml/tuareg/blob/ccde45bbc292123ec20617f1af7f7e19f7481545/tuareg.el#L240

----

#


bbatsov:
Thanks!

PR #1588, opened 2020-10-07
Optimize ibuffer support to avoid calling projectile-project-root

jgkamat:
Previously, `projectile-ibuffer` was very slow for me - this is due to calling `projectile-project-root` (which is rather slow) on all buffers. However, there exists already a function `projectile-project-buffer-p` which is much faster - which can be used to isolate buffers which have no chance of being in the directory. 

The only question is whether this is correct - as far as I know if a buffer's default directory is not under a project root, it cannot be within a project. I'm also assuming this function works correctly over tramp (as I haven't tested it on tramp), although it seems to test that properly (?). 

I did find one discrepancy, currently `projectile-ibuffer` seems to include modes that are ignored by other projectile functions (at least `projectile-globally-ignored-modes` - now they will (correctly?) be dropped from ibuffer. If this behavior isn't desired I can try to modify the function to optionally let ignored buffers through or something.

Regardless, this improves ibuffer performance for me quite a lot - before this patch `projectile-ibuffer` took several seconds to run for me, now it completes instantly. Sorry for not having a real benchmark here, but it seems like these filter functions aren't run when using `benchmark`. I understand that this could be accomplished with the projectile cache, but there's no reason (in my mind) for this function to be so slow even without a cache.

Let me know if tests are needed here - since this is a small change I didn't want to touch too many things.

--

- [N/A] You've updated the readme (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 * [#1540](https://github.com/bbatsov/projectile/pull/1540): Add default `test-suffix` to Angular projects.
 * Add a `:project-file` param to `projectile-register-project-type`.
+* [#1588](https://github.com/bbatsov/projectile/pull/1588) Improve performance of `projectile-ibuffer` with many buffers not in project.
 
 ### Bugs fixed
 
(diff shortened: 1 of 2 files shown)


bbatsov:
The suggested change seems fine (and reasonably safe) to me. Thanks!

PR #1592, opened 2020-10-28
[Docs] mode binds for use-package

hrls:
Simple doc fix


bbatsov:
I've always preferred using `:config` for its simplicity, but I guess it makes sense to break this down as it was intended by `use-package`.

PR #1598, opened 2020-11-15
Ignore clangd cache directory

condy0919:
In recent version of clangd, the clangd cache directory moved to `$PROJECT/.cache/clangd`. See https://github.com/llvm/llvm-project/blob/master/clang-tools-extra/clangd/index/Background.h#L61


bbatsov:
Thanks!

PR #1601, opened 2020-11-19
Separate compile-history

TxGVNN:
Hi @bbatsov, The first thank you for the great package.

I would like to send a PR, that is useful for me.

`projectile-compile` will have separated compile-history by using `projectile--get-command-history`. If something needs to be improved please tell me, I will try to resolve it.

Hope it is useful!
-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4069,10 +4069,12 @@ project of that type"
 
 (defun projectile-read-command (prompt command)
   "Adapted from `compilation-read-command'."
-  (read-shell-command prompt command
-                      (if (equal (car compile-history) command)
-                          '(compile-history . 1)
-                        'compile-history)))
+  (let ((compile-history
+         (ring-elements (projectile--get-command-history (projectile-ensure-project (projectile-project-root))))))
+    (read-shell-command prompt command
+                        (if (equal (car compile-history) command)
+                            '(compile-history . 1)
+                          'compile-history))))
 
 (defun projectile-compilation-dir ()
   "Retrieve the compilation directory for this project."


projectile.el
```diff
@@ -4069,10 +4069,12 @@ project of that type"
 
 (defun projectile-read-command (prompt command)
   "Adapted from `compilation-read-command'."
-  (read-shell-command prompt command
-                      (if (equal (car compile-history) command)
-                          '(compile-history . 1)
-                        'compile-history)))
+  (let ((compile-history
+         (ring-elements (projectile--get-command-history (projectile-ensure-project (projectile-project-root))))))
```
bbatsov, line 4073: I don't quite get how this is supposed to work, so can you explain it in more details?

bbatsov:
This change will need a changelog entry. I'll also have to see what's happened with the CI.

projectile.el
TxGVNN: Yup, I see the package has `projectile--get-command-history` function. It provides command history per project.

I would like to separate the command history per project. So I tried to set `compile-history` value from `projectile--get-command-history` (I think it as  reference/a pointer). And I see it works. I ever did `xref--marker-ring` with the same idea. 

If you ask me in detail how it works, I don't think I know the deep

bbatsov: I was mostly curious why do you need `ring-elements` here, but I assume you copied it from somewhere else in the code.

TxGVNN: Yes, you're right. I copied from here https://github.com/bbatsov/projectile//blob/7b6ac286120421216faae43f9e234d801a30ee3d/projectile.el#L4231

Zalewa:
Unfortunately I have to complain that this change (I bisected and arrived at the commit 0d52268) has broken the compile history completely for me. The commands executed from `projectile-compile-project` are no longer saved anywhere into `~/.emacs.d` and neither they are loaded. Each time I start up Emacs, the `projectile-compile-project` is empty. The history in `compile` works good. Reverting the commit 0d52268 fixes the problem.

MetaDark:
I think it's a great idea to have a separate compile history for each project, but I have the same problem as @Zalewa and reverting commit 0d52268 also fixes the problem for me.

PR #1602, opened 2020-11-20
completion system auto-detection

minad:
@bbatsov Regarding the discussion in https://github.com/raxod502/selectrum/issues/233 - would auto-detection of the completion system like this work too for you? I haven't checked the necessary boxes here, but if there is interest, I will proceed. Maybe there is a reason why you didn't implement auto-detection right away, while you have it for the tags backend.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1823,9 +1823,9 @@ project-root for every file."
     (setq res
           (pcase (if (eq projectile-completion-system 'auto)
                      (cond
-                      ((and (fboundp 'ido-mode) ido-mode) 'ido)
-                      ((and (fboundp 'helm-mode) helm-mode) 'helm)
-                      ((and (fboundp 'ivy-mode) ivy-mode) 'ivy)
+                      ((and (boundp 'ido-mode)  ido-mode)  'ido)
+                      ((and (boundp 'helm-mode) helm-mode) 'helm)
+                      ((and (boundp 'ivy-mode)  ivy-mode)  'ivy)
                       (t 'default))
                    projectile-completion-system)
             ('default (completing-read prompt choices nil nil initial-input))
--- a/projectile.el
+++ b/projectile.el
@@ -47,6 +47,9 @@
   (require 'subr-x))
 
 (eval-when-compile
+  (defvar ido-mode)
+  (defvar ivy-mode)
+  (defvar helm-mode)
   (defvar ag-ignore-list)
   (defvar ggtags-completion-table)
   (defvar tags-completion-table)
@@ -199,10 +202,11 @@ When nil Projectile will consider the current directory the project root."
                  (const :tag "Yes" t)
                  (const :tag "Prompt for project" prompt)))
 
-(defcustom projectile-completion-system 'ido
+(defcustom projectile-completion-system 'auto
   "The completion system to be used by Projectile."
   :group 'projectile
   :type '(radio
+          (const :tag "Auto-detect" auto)
           (const :tag "Ido" ido)
           (const :tag "Helm" helm)
           (const :tag "Ivy" ivy)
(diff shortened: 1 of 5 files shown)


minad:
Tests are failing due to some configuration changes on the side of github?

projectile.el
```diff
@@ -1817,36 +1821,38 @@ project-root for every file."
...
-                          :initial-input initial-input
-                          :action (prog1 action
-                                    (setq action nil))
-                          :caller 'projectile-completing-read)
-              (user-error "Please install ivy from \
-https://github.com/abo-abo/swiper")))
-           (t (funcall projectile-completion-system prompt choices))))
+          (pcase (if (eq projectile-completion-system 'auto)
+                     (cond
+                      ((and (fboundp 'ido-mode) ido-mode) 'ido)
```
bbatsov, line 1826: That's not a good way to depect it, as the function would always be bound and you're never going to match anything else. I'm not sure if there's a simple way to auto-detect reliably, if someone has istalled several of those packages.

minad: Well, it also checks the variable, so this will only trigger when the mode is enabled.

minad: @bbatsov I changed fboundp to boundp for more clarity.

bbatsov: > Well, it also checks the variable, so this will only trigger when the mode is enabled.

Yeah, this should work, but I think it will also result in compilation warnings as we're now referring to a potentially unbound variable.

bbatsov:
> Tests are failing due to some configuration changes on the side of github?

Yep, it seems they removed something.

projectile.el
```diff
@@ -47,6 +47,9 @@
   (require 'subr-x))
 
 (eval-when-compile
+  (defvar ido-mode)
```
bbatsov, line 50: Ah, I missed this part. Sorry.

projectile.el
bbatsov: Forget this. I missed the `defvar`s you've added.

bbatsov:
Weird, the error seems to be coming from the https://github.com/purcell/setup-emacs action, but @purcell seems to have fixed this a while ago https://github.com/purcell/setup-emacs/commit/bea3d9e95785b8412caf802b320cd6de386be7a3

I'm wondering if the old action got cached or something.

minad:
Independent of the test failures, please let me know if you generally agree with this or do you need something else?

bbatsov:
Yep, the approach seems fine overall. Thanks for tackling this!

minad:
Thank you!

projectile.el
```diff
@@ -1817,36 +1821,38 @@ project-root for every file."
...
-                          :initial-input initial-input
-                          :action (prog1 action
-                                    (setq action nil))
-                          :caller 'projectile-completing-read)
-              (user-error "Please install ivy from \
-https://github.com/abo-abo/swiper")))
-           (t (funcall projectile-completion-system prompt choices))))
+          (pcase (if (eq projectile-completion-system 'auto)
+                     (cond
+                      ((and (boundp 'ido-mode)  ido-mode)  'ido)
```
purcell, line 1826: You can use `bound-and-true-p` for these btw.

minad: Thank you! Good to know! Maybe @bbatsov can fix this quickly or should I do another PR?

purcell: It's not a big deal, just wanted to mention it.

minad: Yes, it's good to know. These things are easy to miss if you are unexperienced with elisp. The only thing which helps is going over the list of all functions or maybe there is some kind of linter which also checks things like this (similar to hlint).
But I must also say that the emacs code organization seems a bit chaotic. bound-and-true-p is defined in bindings.el, which does not make much sense, except that the function is just used there immediately after the definition? I went over the subr and subr-x libraries for example and would have noticed such a function...

bbatsov: I'll update that myself. I keep forgetting about that function.

minad: @bbatsov Thank you!

Issue #1603, opened 2020-11-22
Many commands fail with stringp type error if called on project-less files


minad:
* projectile-regenerate-tags
* projectile-ibuffer
* ...

I believe this to be a known issue, see for example #1413.

@bbatsov I can go over all interactive commands and check which fail if called without project. Maybe this could also be added to the testsuite and then we can step by step add a check for existing projects. But I am not entirely sure how you would want this to be fixed. Some commands query for a project and some fail. Maybe change the behavior such that all commands simply fail with a clear message and the user has to chose the project first? Would that make sense?

bbatsov:
> Maybe this could also be added to the testsuite and then we can step by step add a check for existing projects.

Yep, that'd be best.

Maybe change the behavior such that all commands simply fail with a clear message and the user has to chose the project first? Would that make sense?

Yep, I agree. That used to be the case originally, before the changes to `projectile-project-root`, that made it return `nil` instead of throwing an error. I guess in most commands that fail all that will be needed is `projectile-ensure-project`.

minad:
For now I found the following failing commands:
~~~
projectile-display-buffer
projectile-find-implementation-or-test-other-frame
projectile-find-implementation-or-test-other-window
projectile-find-other-file
projectile-find-other-file-other-frame
projectile-find-other-file-other-window
projectile-find-test-file
projectile-grep
projectile-ibuffer
projectile-next-project-buffer
projectile-previous-project-buffer
projectile-project-buffers-other-buffer
projectile-project-info
projectile-regenerate-tags
projectile-remove-current-project-from-known-projects
projectile-switch-to-buffer
projectile-switch-to-buffer-other-frame
projectile-switch-to-buffer-other-window
projectile-toggle-between-implementation-and-test
~~~

PR #1605, opened 2020-11-23
Use projectile-acquire-root and fix some nil errors

minad:
@bbatsov See #1603. I started fixing a few things here and there but I am not sure if this is the right thing to fix the root cause once and for all. Is there a reason not to use `projectile-acquire-root` everywhere? Simply never use `projectile-project-root` and `projectile-ensure-root`. This would at least eliminate this problem completely. I am not sure if this would break some things since you might want to use the functions in hooks and there one might want to prefer an error?

An even cleaner design would probably be to never ask the user and add a project argument to every interactive function, where the interactive argument is supplied by asking. If a function is not used interactively then the argument must be supplied. But I am not sure if such a change is still possible with the current design and it also seems a bit too much given how much work it would be and how much it would break?

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3437,11 +3437,11 @@ regular expression."
   (if (and (boundp 'ggtags-mode)
            (memq projectile-tags-backend '(auto ggtags)))
       (progn
-        (let* ((ggtags-project-root (projectile-project-root))
+        (let* ((ggtags-project-root (projectile-acquire-root))
                (default-directory ggtags-project-root))
           (ggtags-ensure-project)
           (ggtags-update-tags t)))
-    (let* ((project-root (projectile-project-root))
+    (let* ((project-root (projectile-acquire-root))
            (tags-exclude (projectile-tags-exclude-patterns))
            (default-directory project-root)
            (tags-file (expand-file-name projectile-tags-file-name))
--- a/projectile.el
+++ b/projectile.el
@@ -800,7 +800,7 @@ just return nil."
 
 (defun projectile-generate-process-name (process make-new)
   "Infer the buffer name for PROCESS or generate a new one if MAKE-NEW is true."
-  (let* ((project (projectile-ensure-project (projectile-project-root)))
+  (let* ((project (projectile-acquire-root))
          (base-name (format "*%s %s*" process (projectile-project-name project))))
     (if make-new
         (generate-new-buffer-name base-name)
@@ -892,7 +892,7 @@ to invalidate."
          (if prompt
              (completing-read "Remove cache for: "
                               (hash-table-keys projectile-projects-cache))
-           (projectile-ensure-project (projectile-project-root)))))
+           (projectile-acquire-root))))
     (setq projectile-project-root-cache (make-hash-table :test 'equal))
     (remhash project-root projectile-project-type-cache)
     (remhash project-root projectile-projects-cache)
@@ -1588,7 +1588,7 @@ Only buffers not visible in windows are returned."
   "Do a `multi-occur' in the project's buffers.
 With a prefix argument, show NLINES of context."
   (interactive "P")
-  (let ((project (projectile-ensure-project (projectile-project-root))))
+  (let ((project (projectile-acquire-root)))
     (multi-occur (projectile-project-buffers project)
                  (car (occur-read-primary-args))
                  nlines)))
@@ -1933,7 +1933,7 @@ project-root for every file."
 
 (defun projectile-current-project-dirs ()
   "Return a list of dirs for the current project."
-  (projectile-project-dirs (projectile-ensure-project (projectile-project-root))))
+  (projectile-project-dirs (projectile-acquire-root)))
 
 (defun projectile-get-other-files (file-name &optional flex-matching)
   "Return a list of other files for FILE-NAME.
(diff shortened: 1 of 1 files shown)


bbatsov:
> Is there a reason not to use projectile-acquire-root everywhere? Simply never use projectile-project-root and projectile-ensure-root. This would at least eliminate this problem completely.

It was added later, that's the only reason. The other functions are still useful, though, at least `project-project-root` is very useful if you don't want to force the error handling. And we can't really rename `projectile-ensure-root`, as that'd be a breaking change (make packages depend on Projectile's API). 

> An even cleaner design would probably be to never ask the user and add a project argument to every interactive function, where the interactive argument is supplied by asking. If a function is not used interactively then the argument must be supplied. But I am not sure if such a change is still possible with the current design and it also seems a bit too much given how much work it would be and how much it would break?

You mean that every function should ask about a project if called outside a project? I think there are some interactive commands that fallback to the current directory in the absence of a project because some users wanted them to behave int this manner, although I generally don't mind a more uniform behaviour. As to how many things would break in this manner - I can only guess. I doubt many interactive commands are used programatically, but who knows.

bbatsov:
I see most command currently don't accept params or have an optional project param anyways, so probably your suggestion is feasible without causing a lot of havoc.

minad:
> It was added later, that's the only reason. The other functions are still useful, though, at least project-project-root is very useful if you don't want to force the error handling. And we can't really rename projectile-ensure-root, as that'd be a breaking change (make packages depend on Projectile's API).

Ok. Maybe functions could still be deprecated though?

> I see most command currently don't accept params or have an optional project param anyways, so probably your suggestion is feasible without causing a lot of havoc.

Well, the clean solution would only work nicely if we can make the project parameter non-optional. Otherwise we wouldn't win much. But this means too much breakage. So if we use an optional parameter, we could still always pass the parameter within projectile.el without breaking the external api. This would make things within projectile.el a bit more robust, but is also harder to enforce. I have to try it. Then I can tell, if it makes sense.

> You mean that every function should ask about a project if called outside a project? I think there are some interactive commands that fallback to the current directory in the absence of a project because some users wanted them to behave int this manner, although I generally don't mind a more uniform behaviour. As to how many things would break in this manner - I can only guess. I doubt many interactive commands are used programatically, but who knows.

Yes, this is what I meant. Generally I prefer a uniform solution and ideally one which is robust against changes. So it shouldn't be easy to make sure that the correct function is used in the future etc. If we always use the projectile-acquire-root function we would never get the nil errors, but maybe other things break. So maybe this is not such a good idea...

bbatsov:
> Ok. Maybe functions could still be deprecated though?

`projectile-project-root` can't be deprecated, as it has valid use-cases without the need for error handling. `projectile-ensure-project` can potentially be deprecated.

> So it shouldn't be easy to make sure that the correct function is used in the future etc. If we always use the projectile-acquire-root function we would never get the nil errors, but maybe other things break. So maybe this is not such a good idea...

I'm not particularly concerned about breaking some interactive command, but we should still make it possible to use without a project certain commands for which that makes sense, as this is what created the current situation. Unfortunately, all of this happened a while ago and I no longer remember what exactly where the complaints people had about Projectile's command requiring a project (which was the initial setup).

minad:
@bbatsov Thank you for your answer!

> projectile-project-root can't be deprecated, as it has valid use-cases without the need for error handling. projectile-ensure-project can potentially be deprecated.

I agree. From my pov, only ensure-project is a candidate for deprecation, projectile-project-root is not.

You can pull this PR if you wish. This is a very minimal first step and I will need more time to think about how to proceed. Just two few things we should keep in mind:

1. Every function should call projectile-acquire-root at most once. Otherwise we could potentially get two prompts if I understood this correctly. I am not sure if this already holds, since projectile-acquire-root might also be called deeper down in the call chain. Maybe check some call tree visualisation?
2. Every function which calls projectile-project-root or takes a project argument must check that the project is valid or throw an error otherwise. This is not the case right now, resulting in errors. (Maybe some of of the projectile-project-root uses should also be converted to projectile-acquire-root).

bbatsov:
> Every function should call projectile-acquire-root at most once. Otherwise we could potentially get two prompts if I understood this correctly. I am not sure if this already holds, since projectile-acquire-root might also be called deeper down in the call chain. Maybe check some call tree visualisation?

Agreed. Not sure if there's some easy way to visualize this in Emacs. I guess for most interactive commands this should be one of the first things they do (acquire the project root).

> Every function which calls projectile-project-root or takes a project argument must check that the project is valid or throw an error otherwise. This is not the case right now, resulting in errors. (Maybe some of of the projectile-project-root uses should also be converted to projectile-acquire-root).

Yep. I spent some time just now digging through the older changes and I finally remembered what I wanted to do when I started those changes - when `projectile-require-project-root` is `nil` then the current directory becomes a fallback for the project and then `projectile-project-root` never returns `nil`. But when a project is required, obviously that's not the cases and such call sites should be guarded, unless they were preceded by projectile-acquire-root. I'd be caution to replace most calls of `projectile-project-root` with `projectile-acquire-root` as there might be some performance implications due to the extra checks performed, although I guess that should be fine in most cases.

bbatsov:
I'll merge the PR now, as it's safe and allows us to work in small increments on this.

minad:
> Agreed. Not sure if there's some easy way to visualize this in Emacs. I guess for most interactive commands this should be one of the first things they do (acquire the project root).

Actually I think the ideal solution would be if the interactive functions use the "(interactive...)" facilities to query the user. But I am not sure if this is easy to fix. This is what I had in mind when I thought about adding a "project" argument to all the functions.

Thank you for merging! I agree, incremental steps are better for now - at least until it is decided how to proceed.

PR #1607, opened 2020-11-25
Make prefix argument optional for all shells.

jellelicht:
I've made `projectile-run-{shell,eshell,ielm,term}` accept an optional prefix argument, to make them consistent with `projectile-run-vterm`. Please let me know if there was a reason for making these function behave in a different way.

Since eldev seems to not be packaged in guix yet, I was not able to run the tests. My apologies!

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3537,7 +3537,7 @@ regular expression."
     (call-interactively 'gdb)))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm fine with the suggested change. The existing difference was purely an oversight.

PR #1608, opened 2020-11-28
Switch to rebar3

k32:
This PR replaces legacy Erlang build tool `rebar`, that largely fell out of use, with the contemporary version called `rebar3`.
`rebar` script is no longer in active use in the Erlang world, so it makes sense to update the defaults.

-----------------

Thanks!


bbatsov:
Thanks!

PR #1612, opened 2020-12-03
Make projectile-require-project-root setting work on project-less files

### LINKED ISSUE #1603: Many commands fail with stringp type error if called on project-less files

thiagoa:
Closes #1603

This pull request implements the changes described in #1603. As of now, some functions will no longer throw a `stringp type error`, so they will either:

- Prompt the user to select a project when `projectile-require-project-root` is `'prompt`
- Throw a friendly error when `projectile-require-project-root` is `t`
- Use `current-directory` when `projectile-require-project-root` is `nil` (same behavior as before).

I couldn't verify any `string type error` in the following functions:

- `projectile-next-project-buffer`
- `projectile-previous-project-buffer`

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1146,7 +1146,7 @@ Controlled by `projectile-require-project-root'."
     (cond
      ((eq projectile-require-project-root 'prompt) (projectile-completing-read
                                                     "Switch to project: " projectile-known-projects))
-     (projectile-require-project-root (error "Projectile can't find a project definition in %s" dir))
+     (projectile-require-project-root (error "Projectile can’t find a project definition in %s" default-directory))
      (t default-directory))))
 
 (defun projectile-acquire-root (&optional dir)
@@ -1459,7 +1459,7 @@ this case unignored files will be absent from FILES."
 (defun projectile-project-buffers (&optional project)
   "Get a list of a project's buffers.
 If PROJECT is not specified the command acts on the current project."
-  (let* ((project-root (or project (projectile-project-root)))
+  (let* ((project-root (or project (projectile-acquire-root)))
          (all-buffers (cl-remove-if-not
                        (lambda (buffer)
                          (projectile-project-buffer-p buffer project-root))
@@ -1927,7 +1927,7 @@ project-root for every file."
 
 (defun projectile-current-project-files ()
   "Return a list of the files in the current project."
-  (projectile-project-files (projectile-project-root)))
+  (projectile-project-files (projectile-acquire-root)))
 
 (defun projectile-process-current-project-files (action)
   "Process the current project's files using ACTION."
@@ -2948,7 +2948,7 @@ The project type is cached for improved performance."
   "Display info for current project."
   (interactive)
   (message "Project dir: %s ## Project VCS: %s ## Project type: %s"
-           (projectile-project-root)
+           (projectile-acquire-root)
            (projectile-project-vcs)
            (projectile-project-type)))
 
@@ -4496,7 +4496,7 @@ See `projectile--cleanup-known-projects'."
 (defun projectile-remove-current-project-from-known-projects ()
   "Remove the current project from the list of known projects."
   (interactive)
-  (projectile-remove-known-project (abbreviate-file-name (projectile-project-root))))
+  (projectile-remove-known-project (abbreviate-file-name (projectile-acquire-root))))
 
 (defun projectile-ignored-projects ()
   "A list of projects that should not be save in `projectile-known-projects'."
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -1927,7 +1927,7 @@ project-root for every file."
 
 (defun projectile-current-project-files ()
   "Return a list of the files in the current project."
-  (projectile-project-files (projectile-project-root)))
+  (projectile-project-files (projectile-ensure-project (projectile-project-root))))
```
bbatsov, line 1930: It's better to use `projectile-acquire-project`, as its more compact.

projectile.el
```diff
@@ -1146,7 +1146,7 @@ Controlled by `projectile-require-project-root'."
     (cond
      ((eq projectile-require-project-root 'prompt) (projectile-completing-read
                                                     "Switch to project: " projectile-known-projects))
-     (projectile-require-project-root (error "Projectile can't find a project definition in %s" dir))
+     (projectile-require-project-root (error "Projectile can't find a project definition in %s" default-directory))
```
bbatsov, line 1149: Nice catch!

test/projectile-test.el
```diff
@@ -87,6 +88,27 @@ You'd normally combine this with `projectile-test-with-sandbox'."
      (spy-on 'projectile-project-root :and-return-value (file-truename (expand-file-name "project/")))
      ,@body))
 
+(defmacro assert-friendly-error-when-no-project (fn)
```
thiagoa, line 91: To avoid a macro, we could write a single `it` and iterate over a list of functions. However, it would require manually rescuing the error and manually re-throwing the error to include the function name in the error string, to make it clear which function we're dealing with. I wish buttercup would allow us to specify a custom error message, just like rspec does:

```elisp
(expect (funcall 'foo) :to-throw 'error nil "Custom error message here")
```

bbatsov, line 91: Good idea!

bbatsov: Yeah, that would have been nice. It guess it'd be nice to leave your comment here as comment before the macro, so it's clear why it was needed.

bbatsov:
Apart from my small remarks, the PR looks pretty good. Thanks for tackling this!

bbatsov:
It also seems that something is wrong with the new tests.

projectile.el
```diff
@@ -1146,7 +1146,7 @@ Controlled by `projectile-require-project-root'."
     (cond
      ((eq projectile-require-project-root 'prompt) (projectile-completing-read
                                                     "Switch to project: " projectile-known-projects))
-     (projectile-require-project-root (error "Projectile can't find a project definition in %s" dir))
+     (projectile-require-project-root (error "Projectile is unable to find a project definition in %s" default-directory))
```
thiagoa, line 1149: @bbatsov I feel really bad for having changed the error message just to make the tests pass. The reason the test was failing was because of a mismatching quote:

`can't` versus `can’t`

Note how the quote character is different. This failure only happens in CI (and not on my local) and I'm not sure why. The character both from the error message and the test is a normal single quote but for some reason the CI environment is replacing it with a `’`. Any pointers about this let me know. In the meantime, I will try to figure it out.

thiagoa:
@bbatsov All comments have been addressed and the tests are green ✅ 

Now, the only issue is the quote from the error message, as I pointed out [here](https://github.com/bbatsov/projectile/pull/1612#discussion_r535689077).

projectile.el
```diff
@@ -1146,7 +1146,7 @@ Controlled by `projectile-require-project-root'."
     (cond
      ((eq projectile-require-project-root 'prompt) (projectile-completing-read
                                                     "Switch to project: " projectile-known-projects))
-     (projectile-require-project-root (error "Projectile can't find a project definition in %s" dir))
+     (projectile-require-project-root (error "Projectile can’t find a project definition in %s" default-directory))
```
thiagoa, line 1149: @bbatsov I managed to resolve [the issue](https://github.com/bbatsov/projectile/pull/1612#discussion_r535689077) by replacing `'` with `’` here. Note that the tests only failed for a few versions of Emacs (the latest ones). This trick should make it compatible with all versions of Emacs and still allow us to write an assertion to test the error message. Let me know if you're OK with this.

bbatsov (approved):
Great work!

PR #1614, opened 2020-12-05
mark `projectile--mode-line` as a risky local variable

slippycheeze:
`projectile--mode-line` is a buffer-local variable that contains the
global and per-buffer mode line content for projectile.  It is rendered
via `(projectile-mode projectile--mode-line)` in `minor-mode-alist`.

When emacs encounters a symbol in rendering the mode-line, it inserts
the string literally.  As documented in [1]:

> Unless SYMBOL is marked as risky (i.e., it has a non-‘nil’
> ‘risky-local-variable’ property), all text properties specified in
> SYMBOL’s value are ignored.

I wanted to use a pretty icon for the fixed part of projectile mode, as
is my habit, but the properties applied to `projectile-mode-line-prefix`
were ignored.

Screenshot 2020-12-05 at 14 53 35

Marking the variable as risky when set with a file-local or
directory-local value allows Emacs to render text properties, including
in my case:

```elisp
(propertize " prj" 'display (create-image ".../icons/projectile.png")
                                          'png nil
                                          :height 18 :width 18
                                          :ascent 'center
                                          :margin 1))
```

This has no harmful effects on any other use of the variables, either
the ones that compose the mode line, or the local variable used to
cache it.

It will, in the unlikely event anyone sets `projectile--mode-line` in
a file/directory local variable, cause Emacs to prompt them to confirm
they are OK with a change that could result in Emacs executing
arbitrary code.  (which, technically, it could do.)

The usual rules for, eg, locally flagging a value safe are still
available, and this is the normal behaviour for any attempt to, eg, set
a hook variable or whatever in file or directory local variables.

[1] https://www.gnu.org/software/emacs/manual/html_node/elisp/Mode-Line-Data.html

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4845,21 +4845,52 @@ It's used by `projectile-default-mode-line'
 when using dynamic mode line lighter and is the only
 thing shown in the mode line otherwise."
   :group 'projectile
-  :type 'string
+  :type '(choice (string)
+                 (sexp :tag "custom mode-line format"))
+  ;; risky, because it can contain arbitrary mode-line constructs which means
+  ;; it can evaluate arbitrary elisp.
+  :risky t
+  :link '(variable-link mode-line-format)
+  :link '(info-link "(elisp)Mode Line Format")
   :package-version '(projectile . "0.12.0"))
 
-(defvar-local projectile--mode-line projectile-mode-line-prefix)
+(defvar-local projectile--mode-line projectile-mode-line-prefix
+  "Mode line lighter including dynamic content, cached for performance.
+Text properties on `projectile-mode-line-prefix' will be used.
+
+This is marked as a risky local variable so that text properties on strings
+are used, not ignored; there is no good reason to set it as a file- or
+directory-local variable, so it is also flagged as an ignored
+local variable.
+
+See Info node `(elisp)File Local Variables' for details on risky locals
+variables, and Info node `(elisp)Properties in Mode' for how the
+mode-line treats them.")
+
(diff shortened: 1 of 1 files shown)


slippycheeze:
note: if you wanted, you could replace all the current mode-line maintenance with the combination of always setting the `projectile-project-{name,type}` variables, and this mode-line format, which is equivalent to the logic in projectile-default-mode-line`.

```elisp
(defconst projectile-mode-line-format
  '(projectile-mode-line-prefix
    "["
    (projectile-project-name projectile-project-name "-")
    (projectile-project-type (":" projectile-project-type))
    "]"))
```

bbatsov:
I'm fine with the proposed change, although you should also add the rationale to the code and maybe mention in the docs how people can use icons with it. 

I never meant for this variable to be used directly, that's why it's not currently public. The mode line code used to be simpler (it was just relying on the standard machinery to update the modeline or buffer state changes), but as some calculations in the modeline can be expensive this was causing performance issues for users and I optimized it as much as I could. Probably there's a better way to achieve the same results, but I haven't thought much about this lately.

slippycheeze:
I went through, added comments, and updated the docstrings and custom field types to reflect the change.

bbatsov:
Seems 2 modeline tests are failing now for some reason.

slippycheeze:
-_-  ...and I was busy, so hadn't noticed yet.  I'll convert this to a draft until I'm happy that it works correctly, then bounce it back to you.  sorry for the inconvenience!

bbatsov:
No worries.

PR #1618, opened 2020-12-07
projectile-files-via-ext-command: support magic filenames

### LINKED ISSUE #1616: Commit 351e0c3b430392c73372f4129b1174c088a934b2 breaks listing files over TRAMP
## Expected behavior

```
(projectile-files-via-ext-command "/ssh:$HOSTNAME:/path/to/git/project" "git ls-files -zco")
```
Should return a list of files

## Actual behavior

Always returns nil no matter what I do.

Fairly straightforward to reproduce: just try accessing a project over TRAMP and run projectile-find-file. There will be no files available.

This looks to have been introduced in commit 351e0c3b430392c73372f4129b1174c088a934b2. Reverting the function projectile-files-via-ext-command fixes the issue for me.

Does the "with-temp-buffer" change the default-directory? I'm sorry I'm just starting out on my Emacs journey.

```
Projectile version:  20201204.954
```

### Emacs version

GNU Emacs 26.3 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.14)

### Operating system

Ubuntu 20.10

maurelio1234:
In 351e0c3b430392c73372f4129b1174c088a934b2 I switched from
`shell-command-to-string` to `call-process-shell-command` so that we
could capture only the stdout, not the stderr.

However, `call-process-shell-command` does not support magic
filenames, which breaks the use of `projectile-files-via-ext-command`
in TRAMP for example.

In this commit we swich to `shell-command`, which supports both
separating stdout and stderr and magic handlers.

-----------------

Thanks!

Fixes #1616

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1373,9 +1373,12 @@ Only text sent to standard output is taken into account."
   (when (stringp command)
     (let ((default-directory root))
       (with-temp-buffer
-        (call-process-shell-command command nil '(t nil))
-        (let ((shell-output (buffer-substring (point-min) (point-max))))
-          (split-string (string-trim shell-output) "\0" t))))))
+        (let ((stderr-buffer (current-buffer)))
+          (with-temp-buffer
+            (let ((stdout-buffer (current-buffer)))
+              (shell-command command stdout-buffer stderr-buffer)
+              (let ((shell-output (buffer-substring (point-min) (point-max))))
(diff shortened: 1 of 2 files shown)


maurelio1234:
@bbatsov  I just pushed a unit test for the magic file handlers.

bbatsov:
Thanks!

PR #1620, opened 2020-12-08
Use projectile-acquire-root in projectile-vc

fjl:
When calling projectile-vc outside of a project directory, it crashes with

    abbreviate-file-name: Wrong type argument: stringp, nil

This change makes it use projectile-acquire-root, which may prompt the
user or display a nicer error message in this case.


bbatsov:
Thanks!

Issue #1617, opened 2020-12-07
projectile-switch-project does not immediately changes the current buffer


xl666:
I think I found a bug in the latest version of Projectile.
projectile-switch-project does not immediately changes the current
buffer to the one selected in the switching, this is a problem for me
because I do the switching as part of some custom code related to my
workflow. This problem appeared after the last update.

For example, if I'm on the *scratch* buffer and I execute:

(progn (projectile-switch-project) (current-buffer))

The return value is #<buffer *scratch*> instead of whatever buffer is
opened as part of project switching

Doing some hacking, I solved the problem in my code as follows:

(progn (projectile-switch-project)
  (switch-to-buffer (car (buffer-list)))
  (current-buffer))

But is a hacky and ugly solution.

Thanks in advance.

Projectile version information:
20201204.954

Emacs version:

27.1

Operating system:

Linux kernel 5.9

bbatsov:
If I had to guess, I'd say it's probably related to this change https://github.com/bbatsov/projectile/commit/5406793bee52f7005f20a78cf01b064dcbb6dbbc I'll take a closer look at the problem when I can. In the mean time maybe @kevinjfoley or @Deewiant will beat me to it.

kevinjfoley:
@bbatsov - you're right it is related to that change.  Something about changing buffers in a temp buffer doesn't carry over in the same form.  Here's a simplified example that demonstrates the issue:
```
;; call from *scratch*
(progn
  (with-temp-buffer
    (switch-to-buffer "*Messages*")
    ;; First buffer *Messages*
    (message "First buffer %s" (buffer-name)))
  ;; Second buffer *scratch*
  (message "Second buffer %s" (buffer-name)))
```

Someone who understands the mechanics of switching buffers better and how `with-temp-buffer` impacts that may know a better fix but here's my hacky solution:
```
(progn
  (let (buffer)
    (with-temp-buffer
      (switch-to-buffer "*Messages*")
      (message "First buffer %s" (buffer-name))
      (setq buffer (current-buffer)))
    (with-current-buffer buffer
      (message "Second buffer %s" (buffer-name)))))
```

Deewiant:
`with-temp-buffer` is implemented in terms of `save-current-buffer`:

> Record which buffer is current; execute BODY; make that buffer current.

So it makes sense that buffer switches done within the body are lost. @kevinjfoley's solution is what I'd've suggested as well.

[...]

PR #1621, opened 2020-12-09
[#1617] Switch to right buffer in projectile-switch-project-by-name

Deewiant:
As discussed in #1617, manually switch to the correct buffer after exiting from `with-temp-buffer`.

Adds tests for both cases mentioned in #1617, and they fail without the change so in that sense it should be solid.

-----------------

I didn't update the changelog or readme as I'm not sure how you want to indicate this especially since it's fixing a regression from another already-logged change (#1450).

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4392,20 +4392,27 @@ With a prefix ARG invokes `projectile-commander' instead of
                                    'projectile-commander
                                  projectile-switch-project-action)))
     (run-hooks 'projectile-before-switch-project-hook)
-    (let ((default-directory project-to-switch))
-      ;; use a temporary buffer to load PROJECT-TO-SWITCH's dir-locals before calling SWITCH-PROJECT-ACTION
-      (with-temp-buffer
-        (hack-dir-local-variables-non-file-buffer)
-        ;; Normally the project name is determined from the current
-        ;; buffer. However, when we're switching projects, we want to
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -4392,20 +4392,27 @@ With a prefix ARG invokes `projectile-commander' instead of
...
+              ;; to tell the `projectile-project-name' function to ignore the
+              ;; current buffer and the caching mechanism, and just return the
+              ;; value of the `projectile-project-name' variable.
+              (let ((projectile-project-name (funcall projectile-project-name-function
+                                                      project-to-switch)))
+                (funcall switch-project-action)
+                (current-buffer)))))
+      ;; If switch-project-action switched buffers then with-temp-buffer will
+      ;; have lost that change, so switch back to the correct buffer.
+      (if (buffer-live-p switched-buffer)
```
bbatsov, line 4414: You can also use `when` here.

bbatsov:
Apart from the test that fails in CI the changes look good to me.

Deewiant:
Updated to use `when` and with a hopeful fix for the CI issue, let's see.

Deewiant:
Gah, apologies for rushing this so much. Apparently I had left a debug message in the original test so I removed that, and now it took me a few attempts to put the right change in the right test...

But it appears to work in CI too, now.

bbatsov:
Thanks!

Issue #1624, opened 2020-12-13
Projectile creates second window when switching projects


thepreacher:
## Expected behavior
With projectile installed and configured with the basic config:
```
(use-package projectile
  :ensure t
  :custom ((projectile-completion-system 'ivy))
  :init
  (projectile-mode +1)
  :bind (:map projectile-mode-map
              ("s-p" . projectile-command-map)
              ("C-c p" . projectile-command-map)))
```
When I press `C-c p p`, I expect **projectile** via **ivy** to list the available projects which it does. Now when I move the cursor on a desired project and I press **RET**, I expect projectile to enter root directory for the selected project listing it's content. This it does as well.

In addition I expect the current window prior to pressing `C-c p p` to be the only window visible. Then after I select a file by pressing RET, a new window with the content of the file, replace the previous window as the only one visible.

## Actual behavior
What actually happens is, after I hit `C-c p p`:

<img width="738" alt="Screenshot 2020-12-13 at 08 00 31" src="https://user-images.githubusercontent.com/219075/102006894-6a7eb700-3d1c-11eb-8276-5f17ecba4878.png"> 

and press RET on a desired project, projectile enters and list the content of the project root directory (as expected) BUT in addition, it creates a new window. 

<img width="738" alt="Screenshot 2020-12-13 at 08 00 49" src="https://user-images.githubusercontent.com/219075/102006915-8eda9380-3d1c-11eb-9298-0398fa73ac0e.png">

Then when I select and press RET on a file, it displays the content in this newly created window, leaving me with a second window I don't need.

## Steps to reproduce the problem

The code above is the only projectile code I have in my config.

### Projectile version information

```
Projectile version: 20201210.927
```

### Emacs version

*E.g. 27.1.50* 

### Operating system

*MacOS Big Sur*

bbatsov:
Did this start happening recently? I can't reproduce it locally, but I'm wondering if it's not something to do with recent changes to project switching  we were doing.

bbatsov:
(I'm mostly referring to https://github.com/bbatsov/projectile/pull/1621)

thepreacher:
Unfortunately I can't tell if it's a recent change causing it since I was using **doom emacs** before. I'm not imposing or anything but here is a link to my current [init.el](https://www.apaste.info/wJJq) if you may want to try it. Thanks.

thepreacher:
[...]

PR #1626, opened 2020-12-20
projectile-files-via-ext-command: avoid touching users' windows

### LINKED ISSUE #1625: projectile-find-file creates additional windows on Win10
`projectile-git-submodule-command`, with default value `"git submodule --quiet foreach 'echo $path' | tr '\\n' '\\0'"`, uses `tr` which isn't necessarily available on windows.  The missing `tr` utility causes an error which leaves an additional window open.  Setting `projectile-git-submodule-command` to `nil` resolves the problem for those of us that don't care about submodules.

`Projectile 20201214.839`

maurelio1234:
hopefully fixes #1625 and #1624 

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1372,14 +1372,15 @@ If `command' is nil or an empty string, return nil.
 This allows commands to be disabled.
 
 Only text sent to standard output is taken into account."
-  (when (stringp command)
-    (let ((default-directory root))
-      (with-temp-buffer
-        (let ((stderr-buffer (current-buffer)))
-          (with-temp-buffer
-            (shell-command command t stderr-buffer)
-            (let ((shell-output (buffer-substring (point-min) (point-max))))
-              (split-string (string-trim shell-output) "\0" t))))))))
+  (save-window-excursion
+    (when (stringp command)
+      (let ((default-directory root))
+        (with-temp-buffer
+          (let ((stderr-buffer (current-buffer)))
+            (with-temp-buffer
+              (shell-command command t stderr-buffer)
+              (let ((shell-output (buffer-substring (point-min) (point-max))))
+                (split-string (string-trim shell-output) "\0" t)))))))))
 
 (defun projectile-adjust-files (project vcs files)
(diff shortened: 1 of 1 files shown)


bbatsov:
Did you figure out which command exactly results in the extra buffer?

maurelio1234:
I tried to write a TU like that:

```elisp
          (it "does not create new windows"
              (let ((count-windows-before (count-windows))
                    (count-windows-after))
                (projectile-files-via-ext-command "" "nonExistentCommand")
                (setq count-windows-after (count-windows))
                (expect count-windows-after :to-equal count-windows-before)))
```

But it passes even without my fix... I'm not sure how we could test that.

maurelio1234:
> Did you figure out which command exactly results in the extra buffer?

shell-comand does that (around line 3720):

```elisp
	    (when (and error-file (file-exists-p error-file))
              (when (< 0 (file-attribute-size (file-attributes error-file)))
                (with-current-buffer (get-buffer-create error-buffer)
                  (let ((pos-from-end (- (point-max) (point))))
                    (or (bobp)
                        (insert "\f\n"))
                    ;; Do no formatting while reading error file,
                    ;; because that can run a shell command, and we
                    ;; don't want that to cause an infinite recursion.
                    (format-insert-file error-file nil)
                    ;; Put point after the inserted errors.
                    (goto-char (- (point-max) pos-from-end)))
                  (display-buffer (current-buffer))))```

It seems we cannot disable this 'feature'.

bbatsov:
We can always inline and tweak the built-in function. Not sure which approach is better.

maurelio1234:
> We can always inline and tweak the built-in function. Not sure which approach is better.

I guessed `save-window-excursion` was ok, because there's already anyther one (line 1276, commit: 184d3451c9258cfbbfa2b26a07c7daeee2e9a605)

bbatsov:
I think the suggestion solution is fine. My only concern was that the code is not particularly straightforward right now. Thanks for tackling this so quickly!

PR #1632, opened 2020-12-27
Display any errors from projectile files external command

### LINKED ISSUE #1631: projectile-files-via-ext-command sometimes opens a duplicate frame
## Expected behavior

Projectile find files just shows an autocompletable list of files.

## Actual behavior

I get the list of files, but also an additional emacs frame opens showing the current buffer.

* `(setq pop-up-frames 'graphic-only)`
* Change your git repo in a way that will cause `git ls-files` to print errors (e.g. `mv .git/refs-backup .git/refs`)
* Run projectile-find-file

The cause is that `projectile-files-via-ext-command` sets the errors buffer to a temporary buffer. So 
the buffer is created (and shown in a new frame), then `with-temp-buffer` cleans up the buffer leaving an empty frame.

This wouldn't be too for people who only ever use Emacs windows because there's a `save-window-excursion`
around the call, so the new window is quickly deleted.

Possible fixes:

* Write error messages to a realbuffer which will be displayed to the user if there are any errors.
* Let bind `pop-up-frames` to nil within this function.
* Add a `2>/dev/null` to the end of the command to discard stderr (and revert the various attempts at redirecting stderr from inside emacs). But I don't think this works on Windows...

I'll open a PR which does the first option because I think that's the best one, but feel free to reject if not.

@maurelio1234 Tagging you because it looks like you worked on this recently.

```
Projectile version: Projectile 2.4.0snapshot
```

### Emacs version

GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, X toolkit, Xaw scroll bars) of 2020-11-21

### Operating system

Ubuntu 18.04

davidshepherd7:
It's useful to see any error messages that these commands return.

This also fixes another issue: previously if pop-up-frames was non-nil and an
error occurred then a new frame would be opened to display the error buffer. But
the error buffer would be immediately deleted by `with-temp-buffer`, leaving an
additional frame displaying the users current buffer.

The use of `save-window-excursion` worked around this issue when only Emacs
windows were in use, which is perhaps why no-one noticed.

Fixes #1631

I don't think this change is unit-testable, so no tests. I'll update the changelog if you want to merge this.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1380,15 +1380,12 @@ If `command' is nil or an empty string, return nil.
 This allows commands to be disabled.
 
 Only text sent to standard output is taken into account."
-  (save-window-excursion
-    (when (stringp command)
-      (let ((default-directory root))
-        (with-temp-buffer
-          (let ((stderr-buffer (current-buffer)))
-            (with-temp-buffer
-              (shell-command command t stderr-buffer)
-              (let ((shell-output (buffer-substring (point-min) (point-max))))
-                (split-string (string-trim shell-output) "\0" t)))))))))
+  (when (stringp command)
+    (let ((default-directory root))
+      (with-temp-buffer
+        (shell-command command t "*projectile-files-errors*")
+        (let ((shell-output (buffer-substring (point-min) (point-max))))
+          (split-string (string-trim shell-output) "\0" t))))))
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Looks good! Thanks for tackling this!

PR #1633, opened 2020-12-29
updates git-submodules-command to fix relative paths

Gopiandcode:
Updates git-submodules-command to use $displaypath rather than $path.

$displaypath returns the path to a module relative to the current working directory whereas $path (used previously) only returns the path relative to the git root. 

This causes issues with function `projectile-get-sub-projects-command` when the git root is not the same as the project root.

-----------------

Thanks!


bbatsov:
The change seems reasonable. Thanks!

PR #1639, opened 2021-01-14
Do not ask twice for project running ielm, term and vterm

kinote:
Do not ask twice for the project root when calling
`projectile-run-ielm`, `projectile-run-term` or `projectile-run-vterm`
from outside a projectile folder.

Use `projectile-with-default-dir` at the beginning of those functions
to prevent calling `projectile-acquire-root` twice.

When running one of those functions (e.g. projectile-run-vterm) from a folder outside a projectile root (e.g. my home folder) I was being ask twice in the minibuffer to provide a projectile project root. This was caused because projectile-acquire-root was being called twice without setting the default folder.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,13 @@
 
 ### Changes
 
+* Add `project` param to `projectile-generate-process-name`.
 * [#1608](https://github.com/bbatsov/projectile/pull/1608): Use rebar3 build system by default for Erlang projects.
 * Rename `projectile-project-root-files-functions` to `projectile-project-root-functions`.
 
 ### Bugs fixed
 
+* [#1639](https://github.com/bbatsov/projectile/pull/1639): Do not ask twice for project running ielm, term and vterm.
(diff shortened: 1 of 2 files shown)


bbatsov:
Where exactly is the second call to `projectile-acquire-root` happening? That's not immediately clear looking at your patch.

kinote:
> Where exactly is the second call to `projectile-acquire-root` happening? That's not immediately clear looking at your patch.

`projectile-acquire-root` gets called the second time inside `projectile-generate-process-name`.

bbatsov:
Hmm, won't it be much better to just name the `projectile-generate-process-name` take an explicit project name argument? It will definitely make the code less brittle, as anyone can accidentally break it now.

kinote:
> Hmm, won't it be much better to just name the `projectile-generate-process-name` take an explicit project name argument? It will definitely make the code less brittle, as anyone can accidentally break it now.

Yes, I think that makes more sense. I have updated the commit.

bbatsov:
Thanks!

PR #1153, opened 2017-07-25
allow projectile-globally-ignored-dirs to operate on non-root dirs

vincekd:
Replaces this pull (https://github.com/bbatsov/projectile/pull/1146) cuz my fork got all f-ed up.

[Fix #1119] [Fix #733] [Fix #680] 

Might fix [#1041]

-----------------

Before submitting a PR make sure the following things have been done (and denote this
by checking the relevant checkboxes):

- [x] The commits are consistent with our [contribution guidelines](../CONTRIBUTING.md)
- [x] You've added tests (if possible) to cover your change(s)
- [x] All tests are passing (`make test`)
- [x] The new code is not generating bytecode or `M-x checkdoc` warnings
- [x] You've updated the changelog (if adding/changing user-visible functionality)
- [x] You've updated the readme (if adding/changing user-visible functionality)

Thanks!

PR #1640, opened 2021-01-14
Added better documentation on asterisk prefix

dmusican:
More explanation was added as to how the asterisk prefix works, based on
the fabulous documentation by ntc2, here:

https://github.com/bbatsov/projectile/pull/1153#issuecomment-401146353

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -410,6 +410,17 @@ is set to 'alien'."
     ".clangd")
   "A list of directories globally ignored by projectile.
 Regular expressions can be used.
+
+Strings that don't start with * are only ignored at the top level
+of the project. Strings that start with * are ignored everywhere
+in the project, as if there was no *. So note that * when used as
+a prefix is not a wildcard; it is an indicator that the directory
+should be ignored at all levels, not just root.
+
+Examples: \"tmp\" ignores only ./tmp at the top level of the
+project, but not ./src/tmp. \"*tmp\" will ignore both ./tmp and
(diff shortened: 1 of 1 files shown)


bbatsov:
I have to say I'm a bit surprised by this, because you can see in the code that the list of directories are expanded using shell-globbing rules:

```
(defun projectile-expand-paths (paths)
  "Expand the elements of PATHS.

Elements containing wildcards are expanded and spliced into the
resulting paths.  The returned PATHS are absolute, based on the
projectile project root."
  (let ((default-directory (projectile-project-root)))
    (projectile-flatten (mapcar
                         (lambda (pattern)
                           (or (file-expand-wildcards pattern t)
                               (projectile-expand-root pattern)))
                         paths))))
```

e.g.

```
$ (file-expand-wildcards "*.el")
=> (".dir-locals.el" "projectile.el")
```

dmusican:
Hmmm. I admit that I have not read through the code in detail; I stumbled upon this because I could not get it to work as intended, and found that the comment I linked to was the info I needed to make it work. I believe that this functionality is coming from here:

```
If ignored directory prefixed with '*', then ignore all
directories/subdirectories with matching filename,
otherwise operates relative to project root."
  (let ((ignored-files (projectile-ignored-files-rel))
        (ignored-dirs (projectile-ignored-directories-rel)))
    (cl-remove-if
```

bbatsov:
Hmm, I don't remember that change (and the reasoning behind it) and it definitely seems a bit weird to me, but I agree it explains the comment you referenced.

PR #1643, opened 2021-01-18
Use rebar3 shell to run Erlang projects

k32:
`rebar3 shell` command opens REPL with the code of the project preloaded. It is a reasonble default for `projectile-run-project` command for Erlang projects.  

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Add `project` param to `projectile-generate-process-name`.
 * [#1608](https://github.com/bbatsov/projectile/pull/1608): Use rebar3 build system by default for Erlang projects.
 * Rename `projectile-project-root-files-functions` to `projectile-project-root-functions`.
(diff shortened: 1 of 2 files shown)


bbatsov:
Perhaps, although opening a shell/REPL seems a bit weird to me. I definitely wouldn't expect this from a run command. Might be a good idea to add some `:shell` key to the project config, though, and a related command like `projectile-project-shell` or something along those lines.

k32:
Thanks for the feedback! I will take a look at this proposal later this week.

k32:
I have to admit it was easier said than done. My initial idea was to simply spawn an appropriate command (such as `rebar3 shell` or `cabal repl`) in emacs `shell`, which is not hard, but then I realized that many major language modes provide their own functions for starting REPL with additional bells and whistles (e.g. `inferior-erlang-mode`, `haskell-interactive-mode`), and projectile users might want to use these feature-rich REPLs instead. But all these modes have wildly different APIs, and the code for integrating with them may become rather hacky and unmaintainable. I have to experiment more, will close the PR for now.

PR #1655, opened 2021-03-04
Fix file paths in projectile-replace* with prefix arg

bcc32:
When prefix arg is supplied and a directory is selected, the directory
passed to projectile-dir-files is not necessarily the project root, so
using projectile-expand-root yields incorrect behavior.

For example, in a project rooted at `~/project` with files:

```
~/project/
~/project/Makefile
~/project/foo/bar.c
```

With `C-u M-x projectile-replace-regexp`, selecting directory `~/project/foo/`, projectile incorrectly attempts to read file `~/project/bar.c`, resulting in an error.

-----------------

Thanks!

When running `eldev test`, I see that two tests fail, but they also fail in master, so I don't think it has anything to do with this PR.


bbatsov:
The changes look good to me. Thanks!

PR #1657, opened 2021-03-07
Add project detection for Debian packaging directories

DanielG:
This adds project detection for unpacked Debian source packages. We detect
this type by the presence of a debian/control even though technically we
should probably check for all of debian/{rules,contro,changelog}. I found
documenting this too cumbersome so I went with just the control file.

In Debian there's two main (lightweight) ways of building packages:
dpkg-buildpackage and debuild. The latter is pretty much just a wrapper for
the former that also runs static checks across the resulting package.

Developers usually use debuild so that's what we use here. The -uc -us
options we hardcode are extremely common. They disable codesigning which
you'd usually not want to do for a simple developmnet build since it can
always be done using debsign before uploading the package.

Since tests usually run as part of the Debian package build and there isn't
a unified way to run just the tests we don't add a :test command.

Technically there's autopkgtest for running special 'as installed' tests
after the package build, but this requires setting up some virtualisation
drivers beforehand so it doesn't make much sense here either.

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2702,6 +2702,9 @@ test/impl/other files as below:
                                   :compile "bazel build"
                                   :test "bazel test"
                                   :run "bazel run")
+(projectile-register-project-type 'debian "debian/control"
+                                  :project-file "debian/control"
+                                  :compile "debuild -uc -us")
 
 ;; Make & CMake
 (projectile-register-project-type 'make '("Makefile")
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -2702,6 +2702,9 @@ test/impl/other files as below:
                                   :compile "bazel build"
                                   :test "bazel test"
                                   :run "bazel run")
+(projectile-register-project-type 'debian "debian/control"
```
bbatsov, line 2705: Something like `debian-pkg` would be a more descriptive name.

PR #1658, opened 2021-03-09
Fix rediscovery when clearing known projects

claritystorm:
Fix rediscovery in `projectile-clear-known-projects`.

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4656,6 +4656,13 @@ See `projectile--cleanup-known-projects'."
   (projectile-save-known-projects))
 
 ;;;###autoload
+(defun projectile-reset-known-projects ()
+  "Clear known projects and rediscover."
+  (interactive)
+  (projectile-clear-known-projects)
+  (projectile-discover-projects-in-search-path))
+
+;;;###autoload
 (defun projectile-remove-known-project (&optional project)
   "Remove PROJECT from the list of known projects."
   (interactive (list (projectile-completing-read
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
 * Add `projectile-update-project-type-function` for updating the properties of existing project types
+* [#1658](https://github.com/bbatsov/projectile/pull/1658): New command `projectile-reset-known-projects`.
 
 ### Changes
 
--- a/test/projectile-test.el
+++ b/test/projectile-test.el
@@ -1509,6 +1509,14 @@ Just delegates OPERATION and ARGS for all operations except for`shell-command`'.
       (projectile-clear-known-projects)
       (expect projectile-known-projects :to-equal nil))))
 
+(describe "projectile-reset-known-projects"
+  (it "resets known projects"
+    (spy-on 'projectile-clear-known-projects)
+    (spy-on 'projectile-discover-projects-in-search-path)
+    (projectile-reset-known-projects)
+    (expect 'projectile-clear-known-projects :to-have-been-called)
+    (expect 'projectile-discover-projects-in-search-path :to-have-been-called)))
+
 (describe "projectile-test-ignored-directory-p"
   (it "ignores specified literal directory values"
     (spy-on 'projectile-ignored-directories :and-return-value '("/path/to/project/tmp"))


bbatsov:
I'm not sure this constitutes as a fixs, as it was intended to work in this way. I always meant it to just remove everything, which it does. :-)

projectile.el
```diff
@@ -4529,6 +4529,7 @@ See `projectile--cleanup-known-projects'."
   "Clear both `projectile-known-projects' and `projectile-known-projects-file'."
   (interactive)
   (setq projectile-known-projects nil)
+  (projectile-discover-projects-in-search-path)
```
bbatsov, line 4532: Perhaps you can add a prefix argument to the command, which would trigger the project discovery.

claritystorm:
Partial rediscovery already occurs, via `(projectile-save-known-projects)`.

https://github.com/bbatsov/projectile/blob/1528ed4f082e7aaca19f22394eb4bed879645b7c/projectile.el#L4534

projectile.el
claritystorm: Should `(projectile-save-known-projects)` execute in both paths?

bbatsov: I don't remember if the discover function saves what it finds or not. Depends on this.

bbatsov:
I don't see save-known-projects doing any discovery. Am I missing something?

claritystorm:
The `projectile-clear-known-projects` function does remove everything from the `projectile-known-projects` variable. However, this function also immediately repopulates this variable with `(projectile-save-known-projects)`. When `projectile-mode` is re-enabled, `(projectile-discover-projects-in-search-path)` also repopulates this variable. Maybe this should occur together?

projectile.el
claritystorm: If `(projectile-discover-projects-in-search-path)` finds a project to add it does save.

claritystorm: Should neither `(projectile-save-known-projects)` or `(projectile-discover-projects-in-search-path)` execute when no prefix argument is provided?

projectile.el
```diff
@@ -4529,6 +4529,8 @@ See `projectile--cleanup-known-projects'."
   "Clear both `projectile-known-projects' and `projectile-known-projects-file'."
   (interactive)
   (setq projectile-known-projects nil)
+  (when projectile-auto-discover
+    (projectile-discover-projects-in-search-path))
```
bbatsov, line 4533: I see that this command internally updates the known projects, so the second invocation is redundant. Probably you should use `if-else` instead of a `when`.

bbatsov:
> However, this function also immediately repopulates this variable with (projectile-save-known-projects). When projectile-mode is re-enabled, (projectile-discover-projects-in-search-path) also repopulates this variable. Maybe this should occur together?

I guess I messed something up, as the intention was to save `nil` known projects (meaning to reset them). The auto-discovery was added much later, so probably I didn't test it carefully enough.

projectile.el
```diff
@@ -4527,11 +4527,16 @@ See `projectile--cleanup-known-projects'."
     (message "No projects needed to be removed.")))
 
 ;;;###autoload
-(defun projectile-clear-known-projects ()
-  "Clear both `projectile-known-projects' and `projectile-known-projects-file'."
-  (interactive)
+(defun projectile-clear-known-projects (rediscover)
```
bbatsov, line 4530: You need to make this `&optional` to avoid breaking existing clients. That's also why the tests are failing.

projectile.el
```diff
@@ -4527,11 +4527,16 @@ See `projectile--cleanup-known-projects'."
...
+  "Clear both `projectile-known-projects' and `projectile-known-projects-file'.
+
+With a prefix argument REDISCOVER rediscovers projects."
+  (interactive "P")
   (setq projectile-known-projects nil)
-  (projectile-save-known-projects))
+  (when rediscover
+    (projectile-save-known-projects)
```
bbatsov, line 4537: Setting the projects to `nil` without serializing this would result in all projects being the same as before next time you run Projectile. When I suggested the `discover` option I meant it to be something that always does auto-discovery, regardless of `projectile-auto-discover`. Alternatively we can have a `projectile-reset-projects` commands, that just always does remove + rediscovery. Maybe this is going to be more intuitive.

bbatsov:
Looks good. Thanks!

PR #1661, opened 2021-04-08
[Fix #1637] Integrate with savehist-mode

### LINKED ISSUE #1637: Commit 0d52268 has broken storing the compile history between Emacs runs
The compile history is no longer stored between Emacs runs.
The problem was introduced with commit 0d52268, issue #1601.

## Expected behavior

- `projectile-compile-project` command history should be saved on disk (by default in `~/emacs.d/history`)
- When Emacs is closed and started again, `projectile-compile-project` should load the history back from disk.

## Actual behavior

- The history from `projectile-compile-project` isn't saved.
- Even if there's previous history from before this problem started occurring, this history won't be loaded.
- The plain `compile` command loads/saves history just fine.

1. Open any .cpp file from a C++ project in Emacs (my file path, for reference: `/mnt/n/dev/4walls/4walls/src/core/ecs/flags.cpp`)
2. `projectile-compile-project`
3. Customize the compile command.
4. Execute it.
5. Close Emacs.
6. Run Emacs again, reopen the same .cpp file.
7. `projectile-compile-project`, try to go back in history.

I use Helm `20210111.1454`.

```
Projectile 20210104.1216
```

### Emacs version

```
GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.20)
 of 2020-10-07
```

### Operating system

Kubuntu 20.04

TxGVNN:
Hi @bbatsov 

I would like to add a PR to fix #1637. Could you have a look?

Thanks

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -5125,6 +5125,8 @@ Otherwise behave as if called interactively.
     (advice-remove 'compilation-find-file #'compilation-find-file-projectile-find-compilation-buffer)
     (advice-remove 'delete-file #'delete-file-projectile-remove-from-cache))))
 
+;;; savehist-mode - When `savehist-mode' is t, projectile-project-command-history will be saved.
+;; See https://github.com/bbatsov/projectile/issues/1637 for more details
 (if (bound-and-true-p savehist-loaded)
     (add-to-list 'savehist-additional-variables 'projectile-project-command-history)
   (defvar savehist-additional-variables nil)
--- a/projectile.el
+++ b/projectile.el
@@ -5125,6 +5125,13 @@ Otherwise behave as if called interactively.
     (advice-remove 'compilation-find-file #'compilation-find-file-projectile-find-compilation-buffer)
     (advice-remove 'delete-file #'delete-file-projectile-remove-from-cache))))
 
+(if (bound-and-true-p savehist-loaded)
+    (add-to-list 'savehist-additional-variables 'projectile-project-command-history)
+  (defvar savehist-additional-variables nil)
+  (add-hook 'savehist-mode-hook
+            (lambda()
+              (add-to-list 'savehist-additional-variables 'projectile-project-command-history))))
+
 ;;;###autoload
 (define-obsolete-function-alias 'projectile-global-mode 'projectile-mode "1.0")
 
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -5125,6 +5125,13 @@ Otherwise behave as if called interactively.
     (advice-remove 'compilation-find-file #'compilation-find-file-projectile-find-compilation-buffer)
     (advice-remove 'delete-file #'delete-file-projectile-remove-from-cache))))
 
+(if (bound-and-true-p savehist-loaded)
```
bbatsov, line 5128: I think it'd be best to move such setup to `projectile-mode`.

TxGVNN: Yes sure. I updated

projectile.el
```diff
@@ -5118,7 +5118,13 @@ Otherwise behave as if called interactively.
     (add-hook 'projectile-find-dir-hook #'projectile-track-known-projects-find-file-hook t)
     (add-hook 'dired-before-readin-hook #'projectile-track-known-projects-find-file-hook t t)
     (advice-add 'compilation-find-file :around #'compilation-find-file-projectile-find-compilation-buffer)
-    (advice-add 'delete-file :before #'delete-file-projectile-remove-from-cache))
+    (advice-add 'delete-file :before #'delete-file-projectile-remove-from-cache)
+    (if (bound-and-true-p savehist-loaded)
```
bbatsov, line 5122: Oh second, thought - this is not ideal as we'll keep doing the same thing every time the mode starts. Probably it's better to have this at the top-level as initially proposed, just add some comments there explaining the need for this, as it's unlikely that even I'll remember this several months down the road.

TxGVNN: As I see, even we keep doing the same thing when mode starts but seems it doesn't affect anything. Anyway, up to you. Do you want to revert this commit?

TxGVNN: HI @bbatsov, any chance?

bbatsov: Yeah, I think it's better to revert it and document better the previous solution.

TxGVNN: I updated. Hope it is good.

bbatsov:
Thanks!

PR #1662, opened 2021-05-01
Add a Function for Updating the Properties of an Existing Project Type

LaurenceWarne:
Hi! This PR adds a convenience function for changing the properties of a project type that is already registered. The use case would be if you only want to change one parameter and keep the others as they already are. Example usage:

```elisp
   'sbt
   :related-files-fn
   (list
    (projectile-related-files-fn-test-with-suffix "scala" "Spec")
    (projectile-related-files-fn-test-with-suffix "scala" "Test")))
```

Before I was editing `projectile-project-types` manually, so I hope this change is useful if there is not already a nice way to do this.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* Add `projectile-update-project-type-function` for updating the properties of existing project types
+
 ### Changes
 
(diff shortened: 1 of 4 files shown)


bbatsov:
Your changes look good to me. Thanks for working on this!

PR #1667, opened 2021-05-25
update racket project type with :install and :package

xgqt:
This commit adds:
- install - which runs "raco pkg install" to insatll the current package
- package - which runs raco pkg create --source X
  where X is the absolute path to the current package directory

Signed-off-by: Maciej Barć <xgqt@riseup.net>

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3054,7 +3054,9 @@ test/impl/other files as below:
 ;; Racket
 (projectile-register-project-type 'racket '("info.rkt")
                                   :project-file "info.rkt"
-                                  :test "raco test .")
+                                  :test "raco test ."
+                                  :install "raco pkg install"
+                                  :package "raco pkg create --source $(pwd)")
 
 ;; Dart
 (projectile-register-project-type 'dart '("pubspec.yaml")


projectile.el
```diff
@@ -3054,7 +3054,9 @@ test/impl/other files as below:
 ;; Racket
 (projectile-register-project-type 'racket '("info.rkt")
                                   :project-file "info.rkt"
-                                  :test "raco test .")
+                                  :test "raco test ."
+                                  :install "raco pkg install"
+                                  :package (concat "raco pkg create --source " (expand-file-name ".")))
```
bbatsov, line 3059: I don't think that `expand-file-name` is a good idea, as it's just going to be evaluated when Projectile loads. I guess a simple `.` is what you need.

xgqt:
`:package` doesn't work as I intended, it is executed as `"raco pkg create --source <user's home folder>"`

I tried changing to
`:package (concat "raco pkg create --source " (projectile-project-root))`
but this shows only `"raco pkg create --source"` when trying to execute `projectile-package-project`

projectile.el
xgqt: > I don't think that `expand-file-name` is a good idea, as it's just going to be evaluated when Projectile loads. I guess a simple `.` is what you need.

yea, that the catch, it has to be absolute path

xgqt: image


image

xgqt:
changed `:package` to `raco pkg create --source $(pwd)`

bbatsov:
You can't have dynamic code evaluation here, as this is simply evaluated once and becomes a string. It's not something that's going to be evaluated again and again. But I guess you already figured this out. 

I don't get how is `$(pwd)` different from `.`.

xgqt:
> You can't have dynamic code evaluation here, as this is simply evaluated once and becomes a string. It's not something that's going to be evaluated again and again. But I guess you already figured this out.
> 
> I don't get how is `$(pwd)` different from `.`.

It is different in this context, as raco will error on `.`

https://github.com/bbatsov/projectile/pull/1667#discussion_r638488722

xgqt:
Thanks for cooperation!

Issue #1650, opened 2021-02-17
SBT, maven: projectile doesn't create the test file if the project already contains one with the required name


nigredo-tori:
## Expected behavior

With `projectile-create-missing-test-files` enabled, when using `projectile-toggle-between-implementation-and-test` in a file like `.../src/main/.../foo.scala`, Projectile should always create a file `.../src/test/.../fooSpec.scala` in the **corresponding** test directory if such file doesn't exist.

## Actual behavior

If the project already contains a file with a name `fooSpec.scala` (in any directory), Projectile opens it instead.

## Steps to reproduce the problem

1. Start with this directory structure ([projectile-example.zip](https://github.com/bbatsov/projectile/files/5994899/projectile-example.zip)):

    ![image](https://user-images.githubusercontent.com/5962285/108189482-5f446f00-7143-11eb-887e-6e8a45c20f13.png)

2. Open `src/main/scala/bar/package.scala`.
3. Execute `projectile-toggle-between-implementation-and-test`.
4. You are now in `src/test/scala/foo/packageSpec.scala` rather than in `src/test/scala/bar/packageSpec.scala`.

The same behavior can be observed in a maven project. I suspect that this happens in any project large enough to have colliding test file names.

### Projectile version information

```
Projectile 20210125.726
```

### Emacs version

GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0)
 of 2021-02-11

### Operating system

Ubuntu 20.10

LaurenceWarne:
Hi @nigredo-tori, are you happy for this to be closed now #1671/#1679 have been merged?

nigredo-tori:
> Hi @nigredo-tori, are you happy for this to be closed now #1671/#1679 have been merged?

 SBT project [hasn't changed](https://github.com/bbatsov/projectile/blob/master/projectile.el#L3134-L31388), so there's a little bit left to be done here. We can try and copy Maven's settings for SBT, the file/directory structures are identical. I have cobbled a more complicated structure for myself a while back:
```scala
  (projectile-update-project-type
   'sbt
   :src-dir
   (lambda (file-path) (projectile-complementary-dir file-path "test" "main"))
   :test-dir
   (lambda (file-path) (projectile-complementary-dir file-path "main" "test")))
  )
```
, although I can't remember why I have chosen this over what we have for Maven in this project.

LaurenceWarne:
[...]

PR #1671, opened 2021-05-30
Allow Functions for :test-dir

LaurenceWarne:
Hi :wave:, this PR is a proposed solution to https://github.com/bbatsov/projectile/issues/1650, achieved by allowing the `:test-dir` option to take a function which returns the test directory given the implementation directory.

The behaviour for the project in https://github.com/bbatsov/projectile/issues/1650 with the config:

```elisp
(setq projectile-create-missing-test-files t)
   'sbt
   :test-dir
   (lambda (file-path) (projectile-complementary-dir file-path "main" "test")))
```
Is that calling `projectile-toggle-between-implementation-and-test` from `src/main/scala/bar/package.scala` sticks you in `src/main/scala/bar/packageSpec.scala` and calling from `src/test/scala/foo/package.scala` sticks you in `src/test/scala/foo/packageSpec.scala` (regardless of either test file existing).

Calling `projectile-toggle-between-implementation-and-test` from either test file however will prompt a choice between the two existing `package.scala` files. I can PR a similar change for allowing `:src-dir` to take a function if you are happy with these changes.

Setting the option to a function should override existing behaviour (for example `:related-files-fn`) , and if the option is not a function then the behaviour should be the same as before. I've added more detailed documentation in the README about the new behaviour, and also added a comparison of when I think you should use `:test-dir` over `related-files-fn` and vice versa.

Apologies for the large PR :slightly_smiling_face:, I'm not sure how it could be split up on top of omitting changes to `src-dir`.

Thanks, let me know what you think!

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#1671](https://github.com/bbatsov/projectile/pull/1671) Allow the `:test-dir` option of a project to be set to a function for more flexible test switching.
+
 ### Bugs fixed
 
 * [#1673](https://github.com/bbatsov/projectile/issues/1673): Fix CMake system-preset filename.
@@ -10,7 +14,7 @@
 
 ### New features
 
-* Add `projectile-update-project-type-function` for updating the properties of existing project types.
+* Add `projectile-update-project-type` function for updating the properties of existing project types.
 * [#1658](https://github.com/bbatsov/projectile/pull/1658): New command `projectile-reset-known-projects`.
 * [#1656](https://github.com/bbatsov/projectile/pull/1656): Add support for CMake configure, build and test presets. Enabled by setting `projectile-cmake-presets` to non-nil, disabled by default.
 
--- a/projectile.el
+++ b/projectile.el
@@ -3311,19 +3311,42 @@ PROJECT-ROOT is the targeted directory.  If nil, use
      (test-suffix (concat impl-file-name test-suffix "." impl-file-ext))
      (t (error "Project type `%s' not supported!" project-type)))))
 
-(defun projectile-create-test-file-for (impl-file-path)
-  "Create a test file for IMPL-FILE-PATH."
-  (let* ((test-file (projectile--test-name-for-impl-name impl-file-path))
-         (project-root (projectile-project-root))
-         (relative-dir (file-name-directory (file-relative-name impl-file-path project-root)))
-         (src-dir-name (projectile-src-directory (projectile-project-type)))
-         (test-dir-name (projectile-test-directory (projectile-project-type)))
-         (test-dir (expand-file-name (replace-regexp-in-string src-dir-name test-dir-name relative-dir) project-root))
-         (test-path (expand-file-name test-file test-dir)))
-    (unless (file-exists-p test-path)
-      (progn (unless (file-exists-p test-dir)
-               (make-directory test-dir :create-parents))
-             test-path))))
+(defun projectile--impl-to-test-dir (impl-dir-path)
+  "Return the directory path of a test whose impl file resides in IMPL-DIR-PATH.
+
+Occurrences of the current project type's src-dir property (which should be a
+string) are replaced with the current project type's test-dir property
+ (which should be a string) to obtain the new directory.
+
+An error is signalled if either the src-dir or test-dir properties are not
+strings."
+  (let ((test-dir (projectile-test-directory (projectile-project-type)))
+        (impl-dir (projectile-src-directory (projectile-project-type))))
+    (if (and (stringp test-dir) (stringp impl-dir))
+        (projectile-complementary-dir impl-dir-path impl-dir test-dir)
+      (error "Expected the current project's :test-dir and :impl-dir properties to be strings but found %s and %s" test-dir impl-dir))))
+
+(defun projectile-complementary-dir (dir-path string replacement)
+  "Return the \"complementary\" directory of DIR-PATH by replacing STRING in DIR-PATH with REPLACEMENT."
+  (let* ((project-root (projectile-project-root))
+         (relative-dir (file-name-directory (file-relative-name dir-path project-root))))
+    (projectile-expand-root
+     (replace-regexp-in-string string replacement relative-dir))))
+
+(defun projectile--create-directories-for (path)
+  "Create directories necessary for PATH."
+  (unless (file-exists-p path)
+    (make-directory (if (file-directory-p path)
+                        path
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,11 +2,13 @@
...
 ## master (unreleased)
 
+* Allow the `:test-dir` option of a project to be set to a function for more flexible test switching.
+
 ## 2.4.0 (2021-05-27)
 
 ### New features
 
-* Add `projectile-update-project-type-function` for updating the properties of existing project types.
+* Add `projectile-update-project-type` function for updating the properties of existing project types.
```
LaurenceWarne, line 11: Just noticed this, sorry!

bbatsov: No problem.

CHANGELOG.md
```diff
@@ -2,11 +2,13 @@
 
 ## master (unreleased)
 
+* Allow the `:test-dir` option of a project to be set to a function for more flexible test switching.
```
bbatsov, line 5: This has to go under "New features" or "Changes".

projectile.el
```diff
@@ -3342,13 +3351,14 @@ test file."
            "No matching source file found for project type `%s'"
            (projectile-project-type))))
     ;; find the matching test file
-    (let ((test-file (projectile-find-matching-test file-name)))
-      (if test-file
+    (if-let ((overriding-test-file (projectile--test-file-override file-name)))
```
bbatsov, line 3354: Doesn't it make more sense to incorporate this in the existing test lookup function instead of creating the potentially confusing test override concept? I definitely don't like how the code reads now.

projectile.el
```diff
@@ -3443,6 +3453,35 @@ Fallback to DEFAULT-VALUE for missing attributes."
         (or (string-equal prefix-name name)
             (string-equal suffix-name name))))))
 
+(defun projectile--complementary-file (file-path dir-fn filename-fn)
+  "Return `(DIR-FN dir)/(FILENAME-FN name)' where FILE-PATH = dir/name.
```
bbatsov, line 3457: You've quoted this a symbol reference, even though it's not.

projectile.el
```diff
@@ -3313,19 +3313,28 @@ PROJECT-ROOT is the targeted directory.  If nil, use
...
-         (test-dir (expand-file-name (replace-regexp-in-string src-dir-name test-dir-name relative-dir) project-root))
-         (test-path (expand-file-name test-file test-dir)))
-    (unless (file-exists-p test-path)
-      (progn (unless (file-exists-p test-dir)
-               (make-directory test-dir :create-parents))
-             test-path))))
+(defun projectile--impl-to-test-dir (impl-dir-path)
+  "Return the directory path of a test whose impl file resides in IMPL-DIR-PATH.
+
+Occurrences of 'src-dir' are replaced with 'test-dir' to obtain the new directory."
```
bbatsov, line 3319: The quoting references here a confusing, as those are not some references, but rather explanations of the implementation.

projectile.el
```diff
@@ -3313,19 +3313,28 @@ PROJECT-ROOT is the targeted directory.  If nil, use
...
+  (let ((test-dir (projectile-test-directory (projectile-project-type)))
+        (impl-dir (projectile-src-directory (projectile-project-type))))
+    (projectile-complementary-dir impl-dir-path impl-dir test-dir)))
+
+(defun projectile-complementary-dir (dir-path string replacement)
+  "Return the 'complementary' directory of DIR-PATH by replacing STRING in DIR-PATH with REPLACEMENT."
```
bbatsov, line 3325: `complementary` should be double-quoted here.

bbatsov:
>  I can PR a similar change for allowing :src-dir to take a function if you are happy with these changes.

That'd be nice indeed, otherwise it'd be super weird why we have different setups for `src-dir` and `test-dir`. This can be done in a separate PR, though. 

Overall the code looks good, but I'd like to try to simplify a bit the code - in particular it took me a while to understand the need for the overriding function.

projectile.el
LaurenceWarne: yeah, in hindsight that makes more sense.

projectile.el
```diff
@@ -3342,13 +3365,15 @@ test file."
...
     ;; find the matching test file
-    (let ((test-file (projectile-find-matching-test file-name)))
-      (if test-file
-          (projectile-expand-root test-file)
-        (if projectile-create-missing-test-files
-            (projectile-create-test-file-for file-name)
-          (error "No matching test file found for project type `%s'"
-                 (projectile-project-type)))))))
+    (let* ((test-file (projectile-find-matching-test file-name))
+           (test-file-or-fallback (or test-file (projectile--test-file-fallback file-name)))
```
LaurenceWarne, line 3369: `projectile--test-file-fallback` here does the same as `projectile-create-test-file-for` did previously, minus the creation of the test file which has been moved to this function.

I'm undecided whether moving `projectile--test-file-fallback` to `projectile-find-matching-test` (as a final `when-let` :smile:) or keeping it here would make more sense. Let me know what you think.

LaurenceWarne:
> That'd be nice indeed, otherwise it'd be super weird why we have different setups for src-dir and test-dir. This can be done in a separate PR, though.

Thanks, hopefully that should be straightforward once you're happy with this PR :+1: 

> Overall the code looks good, but I'd like to try to simplify a bit the code - in particular it took me a while to understand the need for the overriding function.

Thanks for feedback and fast response, hopefully I've simplified it a little now by integrating the precedence logic into `projectile-find-matching-test`, let me know what you think.

projectile.el
```diff
@@ -3441,13 +3466,37 @@ Fallback to DEFAULT-VALUE for missing attributes."
...
+        test-dir
+        #'projectile--test-name-for-impl-name)
+       (projectile-project-root)))))
+
 (defun projectile--find-matching-test (impl-file)
   "Return a list of test files for IMPL-FILE."
-  (if-let ((plist (projectile--related-files-plist-by-kind impl-file :test)))
-      (projectile--related-files-from-plist plist)
-    (if-let ((predicate (projectile--impl-to-test-predicate impl-file)))
+  (if-let ((test-file-from-test-dir-fn
```
LaurenceWarne, line 3492: Here is where the `:test-dir` when set to a function takes precedence over `:related-files-fn`. If it is not a function, the behaviour should be the same as before.

LaurenceWarne:
@bbatsov Do you have the availability to take another look at this? I've added some comments + justifications to the PR so hopefully it should be easier to follow, thanks.

bbatsov:
Looks good! Thanks!

Issue #1669, opened 2021-05-27
[Proposal] Should projectile-compile-project support comint-mode?


TxGVNN:
Hi @bbatsov,

Today, I ran a command need interactively then I see `projectile-compile-project` doesn't support comint-mode as `compile` command is doing.

Should we support it?

## Expected behavior

`C-u M-x projectile-compile-project` run `(compile cmd t)` as `C-u M-x compile`

## Actual behavior

`C-u M-x projectile-compile-project` has the meaning same with `(setq compilation-read-command t)`.

## Proposal

We had `compilation-read-command` to decide to read the command or not. So I propose use `(C-u) PREFIX` to decide to run `comint-mode` in compilation mode.

If you see that makes sense and agree on this change. I can take time to work on it

TxGVNN:
I switched to use `project-compile` instead, thanks.

bbatsov:
Fair enough.

bbatsov:
Fixed by https://github.com/bbatsov/projectile/pull/1672

PR #1672, opened 2021-06-01
Add projectile-comint-mode variable to allow for interactive compilation buffers

Toctave:
Hi,
This is my first PR here, I'm trying to follow the guidelines but I apologize in advance if I'm doing something wrong. I have not updated the readme or docs yet because I haven't found a suitable place to put my change and I didn't want to change the wrong files.

This is a very simple change that leverages Emacs' `compile` command argument to allow for interactive compilation buffers. This is particularly useful when using `projectile-project-run-cmd`, as it allows you to interact with your program.

-----------------

Thanks!

### DIFF
--- a/doc/modules/ROOT/pages/projects.adoc
+++ b/doc/modules/ROOT/pages/projects.adoc
@@ -717,9 +717,10 @@ your project, you could customize it with the following:
 
 By default, compilation buffers are not writable, which allows you to
 e.g.  press `g` to restart the last command. Setting
-`projectile-comint-mode` to a non-nil value allows you to make
-projectile compilation buffers interactive, letting you e.g. test a
-command-line program with `projectile-run-project`.
+`projectile-<cmd>-use-comint-mode` (where `<cmd>` is `configure`,
+`compile`, `test`, `install`, `package`, or `run`) to a non-nil value
+allows you to make projectile compilation buffers interactive, letting
+you e.g. test a command-line program with `projectile-run-project`.
 
 [source,elisp]
 ----
--- a/doc/modules/ROOT/pages/projects.adoc
+++ b/doc/modules/ROOT/pages/projects.adoc
@@ -735,3 +735,16 @@ external command or an Emacs Lisp function:
 ----
 (setq projectile-test-cmd #'custom-test-function)
 ----
+
+== Interact with a project's compilation buffer
+
+By default, compilation buffers are not writable, which allows you to e.g.
+press `g` to restart the last command. Setting `projectile-comint-mode` to a
+non-nil value allows you to make projectile compilation buffers interactive.
+This is especially useful for testing a program with `projectile-run-project`.
+
+[source,elisp]
+----
+(setq projectile-comint-mode t)
+----
+
--- a/doc/modules/ROOT/pages/projects.adoc
+++ b/doc/modules/ROOT/pages/projects.adoc
@@ -715,10 +715,11 @@ your project, you could customize it with the following:
 ((nil . ((projectile-project-name . "your-project-name-here"))))
 ----
 
-By default, compilation buffers are not writable, which allows you to e.g.
-press `g` to restart the last command. Setting `projectile-comint-mode` to a
-non-nil value allows you to make projectile compilation buffers interactive,
-letting you e.g. test a command-line program with `projectile-run-project`.
+By default, compilation buffers are not writable, which allows you to
+e.g.  press `g` to restart the last command. Setting
+`projectile-comint-mode` to a non-nil value allows you to make
+projectile compilation buffers interactive, letting you e.g. test a
+command-line program with `projectile-run-project`.
 
 [source,elisp]
 ----
--- a/doc/modules/ROOT/pages/projects.adoc
+++ b/doc/modules/ROOT/pages/projects.adoc
@@ -715,6 +715,16 @@ your project, you could customize it with the following:
 ((nil . ((projectile-project-name . "your-project-name-here"))))
 ----
 
+By default, compilation buffers are not writable, which allows you to e.g.
+press `g` to restart the last command. Setting `projectile-comint-mode` to a
+non-nil value allows you to make projectile compilation buffers interactive,
+letting you e.g. test a command-line program with `projectile-run-project`.
+
+[source,elisp]
+----
+(setq projectile-comint-mode t)
+----
+
 == Configure a Project's Lifecycle Commands
 
 There are a few variables that are intended to be customized via `.dir-locals.el`.
@@ -734,17 +744,5 @@ external command or an Emacs Lisp function:
 [source,elisp]
 ----
 (setq projectile-test-cmd #'custom-test-function)
-----
-
-== Interact with a project's compilation buffer
-
-By default, compilation buffers are not writable, which allows you to e.g.
-press `g` to restart the last command. Setting `projectile-comint-mode` to a
-non-nil value allows you to make projectile compilation buffers interactive.
-This is especially useful for testing a program with `projectile-run-project`.
-
-[source,elisp]
-----
-(setq projectile-comint-mode t)
 ----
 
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -4447,11 +4447,14 @@ project of that type"
       (projectile-read-command prompt default-cmd)
     default-cmd))
 
+(defvar projectile-comint-mode nil
```
bbatsov, line 4450: Shouldn't this be a `defcustom`?

Toctave: It should indeed. I just fixed it

bbatsov:
It's good to mention this somewhere in the docs, otherwise probably no one will find it.

Toctave:
Just added a section at the end of [Configuring projectile's behavior](https://docs.projectile.mx/projectile/2.2/projects.html#configuring-projectiles-behavior) . Hope this was the right place to do so?

Toctave:
I just noticed this was requested in #1669 . My approach is different from the one suggested in that thread. Any thoughts @bbatsov ?

projectile.el
```diff
@@ -4447,11 +4447,15 @@ project of that type"
       (projectile-read-command prompt default-cmd)
     default-cmd))
 
+(defcustom projectile-comint-mode nil
```
bbatsov, line 4450: Probably this should be named `projectile-use-compilation-shell-minor-mode` or something along those lines, as `comint-mode` is something else and that would be confusing IMO.

doc/modules/ROOT/pages/projects.adoc
```diff
@@ -665,6 +665,17 @@ your project, you could customize it with the following:
 ((nil . ((projectile-project-name . "your-project-name-here"))))
 ----
 
+By default, compilation buffers are not writable, which allows you to
+e.g.  press `g` to restart the last command. Setting
+`projectile-comint-mode` to a non-nil value allows you to make
+projectile compilation buffers interactive, letting you e.g. test a
```
bbatsov, line 671: I don't think "interactive" is the write term here.

bbatsov:
The two approaches are not mutually exclusive. I guess we can still have some prefix argument for the cases when people want this functionality on demand.

projectile.el
bbatsov: Oh, well the description is actually a bit confusing and I don't have time to test this myself:

> If optional second arg COMINT is t the buffer will be in Comint mode with
‘compilation-shell-minor-mode’.

Toctave: It is both : when a prefix argument is used, the major mode of the compilation buffer is `comint-mode`, and the minor mode `compilation-shell-minor-mode` is activated. `comint-mode` is the more important one, though. All `compilation-shell-minor-mode` does is rebind keys to avoid conflicts : 

```
When Compilation Shell minor mode is enabled, all the
error-parsing commands of the Compilation major mode are
available but bound to keys that don't collide with Shell mode.
```

Perhaps `projectile-comint` would be the best choice of name : it uses `comint`, which is the name of the argument of the `compile` function and of the major mode the compilation buffer will be in, and avoids the `-mode` prefix, which might wrongly imply that this is an actual mode. Would you be fine with that ?

doc/modules/ROOT/pages/projects.adoc
Toctave: I used that word because it is the one that is used in the Emacs documentation of `compile` (`C-h f compile <ENTER>`): 

```
Additionally, with universal prefix arg, compilation buffer will be in
comint mode, i.e. interactive.
```

projectile.el
Toctave: For completeness's sake, here are the bindings defined in `compilation-shell-minor-mode-map` :

```
C-M-n  compilation-next-error
C-M-p  compilation-previous-error
M-RET  compile-goto-error
M-{    compilation-previous-file
M-}    compilation-next-file
```

doc/modules/ROOT/pages/projects.adoc
bbatsov: Got it.

projectile.el
bbatsov: Got it. Okay, let's name this `projectile-compilation-use-comint-mode` or something like this. I just want to make it a bit more explicit that this is related to complition commands.

projectile.el
```diff
@@ -4447,11 +4447,15 @@ project of that type"
...
+  "If non-nil, launch projectile compilation buffers in interactive mode."
+  :type 'bool)
+
 (defun projectile-run-compilation (cmd)
```
bbatsov, line 4454: I think you can also add here an optional argument called `comint-mode`, so that people can triggered potentially with a prefix arg from interactive commands.

bbatsov:
You'll also need to rebase.

projectile.el
Toctave: I would, but the user level function is `projectile-compile-project`, which already has a prefix argument (used to force the command prompt). Should I remove it in favour of this one ?

Toctave:
I've rebased and added a variable for each type of compilation command (didn't want to refactor the `projectile-<cmd>-project` code, but it's quite repetitive). 

Only thing left is to (maybe?) change the use of the prefix argument for the `projectile-<cmd>-project` functions.

doc/modules/ROOT/pages/projects.adoc
```diff
@@ -715,6 +715,18 @@ your project, you could customize it with the following:
...
+By default, compilation buffers are not writable, which allows you to
+e.g.  press `g` to restart the last command. Setting
+`projectile-<cmd>-use-comint-mode` (where `<cmd>` is `configure`,
+`compile`, `test`, `install`, `package`, or `run`) to a non-nil value
+allows you to make projectile compilation buffers interactive, letting
+you e.g. test a command-line program with `projectile-run-project`.
+
+[source,elisp]
+----
+(setq projectile-comint-mode t)
```
bbatsov, line 727: You forgot to update this.

bbatsov:
> I've rebased and added a variable for each type of compilation command (didn't want to change the structure of the projectile-<cmd>-project code, but it's quite repetitive)

I meant this in the sense that all those commands use `compile` internally. :-) Probably we're fine with a single config for all commands.

projectile.el
```diff
@@ -4536,11 +4536,11 @@ project of that type"
       (projectile-read-command prompt default-cmd)
     default-cmd))
 
-(defun projectile-run-compilation (cmd)
+(defun projectile-run-compilation (cmd use-comint-mode)
```
bbatsov, line 4539: I'd make `use-comint-mode` optional (`&optional).

projectile.el
```diff
@@ -4582,9 +4582,39 @@ The command actually run is returned."
...
     (unless (file-directory-p default-directory)
       (mkdir default-directory))
-    (projectile-run-compilation command)
+    (projectile-run-compilation command use-comint-mode)
     command))
 
+(defcustom projectile-configure-use-comint-mode nil
+  "If non-nil, make the output buffer of projectile-configure-project
+interactive."
+  :type 'bool)
```
bbatsov, line 4591: There should be a few other properties here. Refer to existing `defcustoms` for those.

doc/modules/ROOT/pages/projects.adoc
Toctave: What do you mean ? I did update it to reflect the new variable names

Toctave:
> > I've rebased and added a variable for each type of compilation command (didn't want to change the structure of the projectile--project code, but it's quite repetitive)
> 
> I meant this in the sense that all those commands use `compile` internally. :-) Probably we're fine with a single config for all commands.

I'm not sure : my use case would be to make the `run` buffer interactive, and keep all other ones in compilation mode. This means I keep the syntax highlighting and links to errors in the compilation buffer, and I can interact with my program when I test it out.

bbatsov:
Got it. Well, I guess we can go with the granular defcustoms then.

PR #1696, opened 2021-07-30
fix: compilation-find-file advice

actondev:
We should check if file doesn't exist in the passed directory and not
the default directory. In cases where the compilation happens in
another directory (like ./build), the passed filename can be
relative (like ../file.cpp) and checking for its existence in the
default-directory fails

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4739,7 +4739,10 @@ If the prefix argument SHOW_PROMPT is non nil, the command can be edited."
 (defun compilation-find-file-projectile-find-compilation-buffer (orig-fun marker filename directory &rest formats)
   "Try to find a buffer for FILENAME, if we cannot find it,
 fallback to the original function."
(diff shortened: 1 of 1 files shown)


actondev:
Was late at night so I was not quite verbose :)

I asked in emacs reddit before I understood what was happening ( https://www.reddit.com/r/emacs/comments/ouad7r/compilegotoerror_wrong_file/ )

Also, if I may ask, what is the reason of this advice?
I'm thinking if in the advice would be better to set `compilation-search-path` and call the original `compilation-find-file`

bbatsov:
It seems to me this problem is already fix by https://github.com/bbatsov/projectile/pull/1691

actondev:
Nice!
I'm closing then this for now. Will test with the latest commit and if I get any problems will come back to you.

PR #1702, opened 2021-08-25
[Add #1517] Add per-project compilation buffer variable

### LINKED ISSUE #1517: Make *compilation* per-project
Right now any compilation/test/etc. happens in the single `*compilation*` buffer, so switching between projects isn't as independent as I would like (if the first project has a long compilation, I can't go work on something else and compile it while I wait for the first project to finish).

## Expected behavior

- `C-c p c` to compile the current project
- `C-c p p` to switch projects
- `C-c p c` to compile the second project, both now compiling concurrently, ideally in buffers named something like `*compilation*<project1>` and `*compilation*<project2>`

## Actual behavior

The second `C-c p c` says "A compilation process is running` kill it?"

The expected behavior steps should do it, I think. You need two projects that have a compilation time of over a few seconds. Using `yes` as the compilation command should get it to just run indefinitely.

Projectile 20191216.2105

### Emacs version

GNU Emacs 26.3 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.13)

### Operating system

NixOS 20.03

kijimaD:
Fix #1517 
continue #1535 PR

I made it easy to per-project compilation buffer configures!

To make per-project  compilation buffer, currently require user set:
```emacs lisp
(setq compilation-buffer-name-function
      #'projectile-compilation-buffer-name)
(setq compilation-save-buffers-predicate
      #'projectile-current-project-buffer-p)
```

Proposing changes can set:
```emacs lisp
(setq projectile-per-project-compilation-buffer t)
```

- Easy to set.
- Do not affect the other packages compilation buffer.

Thank you!

-----------------

Thanks!

### DIFF
--- a/doc/modules/ROOT/pages/configuration.adoc
+++ b/doc/modules/ROOT/pages/configuration.adoc
@@ -449,14 +449,11 @@ This affects all commands built on top of `projectile--run-project-cmd` like:
 
 Normally, the buffers created by those commands would be shared (overwritten)
 between projects, but it's also possible to make the compilation buffer names
-project-specific. Currently, this requires that the user explicitly set:
+project-specific. This requires that the user set:
 
 [source,elisp]
 ----
-(setq compilation-buffer-name-function
-      #'projectile-compilation-buffer-name)
-(setq compilation-save-buffers-predicate
-      #'projectile-current-project-buffer-p)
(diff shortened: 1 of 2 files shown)


bbatsov:
Nice improvement!

PR #1706, opened 2021-09-03
[Fix #1703] Improve projectile-generic-command default

### LINKED ISSUE #1703: Issue with projectile-generic-command
Recently found an issue with generic find command. When issued it returns paths with leading `./`, e.g.:

```
~/configs/emacs$ find .
.
./init.el
./setup.el
```

While `git ls-files` and `fd` does not. Because of leading dot slash projectile does not remove ignored directories and files.

I've also found that `fd` is called `fdfind` in ubuntu. Locally I resolved it this way:

```elisp
  (setq projectile-generic-command
        (if (executable-find "fdfind")
            "fdfind . -0 --type f --color=never"
          "find . -type f -print0 | cut -c3-"))
```

Think that dot slash should be removed in `projectile-files-via-ext-command` function.

vitek:
* Use fd/fdfind when available
* Strip leading ./ from paths produced by find

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -690,9 +690,11 @@ Set to nil to disable listing submodules contents."
   :type 'string)
 
 (defcustom projectile-generic-command
-  (if (executable-find "fd")
-      "fd . -0 --type f --color=never"
-    "find . -type f -print0")
+  (cond ((executable-find "fd")
+         "fd . -0 --type f --color=never")
+        ((executable-find "fdfind")
+         "fdfind . -0 --type f --color=never")
+        (t "find . -type f | cut -c3- | tr '\\n' '\\0'"))
   "Command used by projectile to get the files in a generic project."
   :group 'projectile
   :type 'string)
--- a/test/projectile-test.el
+++ b/test/projectile-test.el
@@ -747,7 +747,7 @@ Just delegates OPERATION and ARGS for all operations except for`shell-command`'.
         ;; After listing all the files, the cache should have been updated.
         (projectile-current-project-files)
         ;; find returns the leading ./ therefore the somewhat odd notation here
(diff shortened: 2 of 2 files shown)


projectile.el
```diff
@@ -1419,7 +1421,8 @@ Only text sent to standard output is taken into account."
       (with-temp-buffer
         (shell-command command t "*projectile-files-errors*")
         (let ((shell-output (buffer-substring (point-min) (point-max))))
-          (split-string (string-trim shell-output) "\0" t))))))
+          (mapcar (lambda (pat) (string-remove-prefix "./" pat))
```
bbatsov, line 1424: I'm afraid we can't do this, as it's going to kill performance for any larger project. It's very important to avoid mapping over all the files in Elisp when doing `alien` indexing (it's pretty slow).

bbatsov:
There are a few more ideas for `find` to consider here https://stackoverflow.com/questions/2596462/how-to-strip-leading-in-unix-find If only `macOS` was using GNU find... :-)

projectile.el
vitek: Ok, then I can change default find command to `find . -type f | cut -c3- | tr '\n' '\0'`. Would that be ok?

Tried it on large project there is minimal difference:

```
$  find . -type f | wc -l
184034
$ time find . -type f -print0 > /dev/null

real	0m2,112s
user	0m0,102s
sys	0m0,998s
$ time find . -type f | cut -c3- | tr '\n' '\0'> /dev/null

real	0m2,126s
user	0m0,201s
sys	0m0,989s
```

bbatsov: That looks reasonable.

vitek: Ok, updated PR

bbatsov:
Thanks!

PR #1707, opened 2021-09-09
fix byte compilation warnings

markgdawson:
Minor fixes to remove byte compilation warnings on projectile.el

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1049,7 +1049,7 @@ discover projects there."
           (dolist (dir (directory-files directory t))
             (when (and (file-directory-p dir)
                        (not (member (file-name-nondirectory dir) '(".." "."))))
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -1050,7 +1049,7 @@ discover projects there."
           (dolist (dir (directory-files directory t))
             (when (and (file-directory-p dir)
                        (not (member (file-name-nondirectory dir) '(".." "."))))
-              (projectile-discover-projects-in-directory dir (1- depth))))
+              (projectile-discover-projects-in-directory dir)))
```
bbatsov, line 1052: Why did you change this?

markgdawson: You're right. This shouldn't be here. Reverted.

PR #1713, opened 2021-09-30
[#1709] Fix `projectile-project-buffer-p` contain current buffer

### LINKED ISSUE #1709: projectile-switch-buffers with prefix argument plus k also kills current buffer
## Expected behavior
<kbd>C-u C-c p p Enter k</kbd> (i.e. prefix argument plus `projectile-switch-project`, select project, plus <kbd>k</kbd> for kill) should only kill buffers of the subsequently selected project. In particular, do not kill the current buffer if it is not of that project (otherwise, I wouldn't have to use the prefix command; I could just run `projectile-kill-buffers`).

## Actual behavior
The current buffer is killed, as well, even though it is not part of the selected project.

Open one or more buffers of project 'a'. Switch to a buffer of project 'b'. Enter prefix <kbd>C-u</kbd> and run `projectile-switch-project` (e.g., via <kbd>C-c p p</kbd>), select a project, and press <kbd>k</kbd> for kill.

```
Projectile 2.6.0-snapshot
```

### Emacs version
GNU Emacs 27.2 (build 1, x86_64-apple-darwin18.7.0, Carbon Version 158 AppKit 1671.6) of 2021-03-27

### Operating system
macOS Big Sur Version 11.5.2 (20G95)

kijimaD:
Fix #1709

## Cause

If `projectile-project-buffer-p` is executed for a non-current project, `default-directory` return current directory.
Since `projectile-project-buffers` contains current buffer, kill current buffer unintentionally.

## Solution

`default-directory` → `(file-name-directory buffer-file-name)`

Seem to it works correctly. 
→ not working https://github.com/bbatsov/projectile/pull/1713#issuecomment-930844738

If `buffer-file-name` is nil, use `default-directory`.

## How to confirm

- `C-u` `projectile-switch-project` {select project}  `k` 
not contain current buffer✔

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1547,13 +1547,16 @@ If PROJECT is not specified the command acts on the current project."
 (defun projectile-project-buffer-p (buffer project-root)
   "Check if BUFFER is under PROJECT-ROOT."
   (with-current-buffer buffer
-    (and (not (string-prefix-p " " (buffer-name buffer)))
-         (not (projectile-ignored-buffer-p buffer))
-         buffer-file-name
-         (string-equal (file-remote-p (file-name-directory buffer-file-name))
-                       (file-remote-p project-root))
-         (not (string-match-p "^http\\(s\\)?://" (file-name-directory buffer-file-name)))
-         (string-prefix-p project-root (file-truename (file-name-directory buffer-file-name)) (eq system-type 'windows-nt)))))
+    (let ((directory (if buffer-file-name
+                         (file-name-directory buffer-file-name)
+                       default-directory)))
+      (and (not (string-prefix-p " " (buffer-name buffer)))
+           (not (projectile-ignored-buffer-p buffer))
+           directory
+           (string-equal (file-remote-p directory)
+                         (file-remote-p project-root))
+           (not (string-match-p "^http\\(s\\)?://" directory))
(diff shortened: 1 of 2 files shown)


projectile.el
```diff
@@ -1549,11 +1549,11 @@ If PROJECT is not specified the command acts on the current project."
   (with-current-buffer buffer
     (and (not (string-prefix-p " " (buffer-name buffer)))
          (not (projectile-ignored-buffer-p buffer))
-         default-directory
-         (string-equal (file-remote-p default-directory)
+         buffer-file-name
```
bbatsov, line 1552: This won't work for special buffers - those that are not backed by files.

bbatsov:
> Seem to it works correctly.

The problem is this won't work for special buffers (e.g. a `dired`/shell buffer for a project).

kijimaD:
Oh, that's right...
Thank you for your review and I'm sorry!

kijimaD:
> The problem is this won't work for special buffers (e.g. a dired/shell buffer for a project).

I try to fix it on c8c6921!  How about?

bbatsov:
Yeah, that looks like a reasonable fix to me. Thanks!

PR #1716, opened 2021-10-31
Allow completions for projectile-run{-async,}-shell-command-in-root

### LINKED ISSUE #1692: Tab-completion in projectile-run-shell-command-in-root
## Expected behavior
Completion by tab in projectile-run-shell-command-in-root instead of printing tab
## Actual behavior
Pressing tab prints tab

1. Install last version from melpa
2. Press hotkey(default is `C-c p !`)
3. Press tab

*This is extremely important! Providing us with a reliable way to reproduce
a problem will expedite its solution.*

Arch Linux
Emacs 27.2
Projectile 20210701.441 from Melpa

This had crashed after reinstalling plugin, i tried to delete everything in `~/.emacs.d/`, that doesn't help.

I tried it on another computer, but with old version, all works, when i press `C-c p !` it require to pick project

I found the problem: pull request #1675 disabled completion, because it doesn't run `shell-command` directly.

daanturo:
By using `read-shell-command`.

Fixes #1692 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 ### Changes
 
 * [#1447](https://github.com/bbatsov/projectile/issues/1447): Restructure the menu.
-
+* [#1716](https://github.com/bbatsov/projectile/pull/1716): Allow completions when reading shell-commands.
 ## 2.5.0 (2021-08-10)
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #1717, opened 2021-11-03
Show a warning if the projectile cache is not writable

gizmomogwai:
I had a misconfiguration where my cache-file variable was pointing to nowhere (actually not even the parent folder was there, so the file was also not createable which leads `file-writeable-p` to always return false).
The cache was always invalid and recalculated, because the file was not there
and `projectile-maybe-invalidate-cache` would always call `projectile-invalidate-cache`.

With this patch at least a warning is printed to `*Messages*` that something is not quite right.

-----------------

- [-] You've added tests (if possible) to cover your change(s)
- [-] The new code is not generating bytecode or `M-x checkdoc` warnings
- [-] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)
- [-] You've updated the readme (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -860,9 +860,10 @@ specify a project explicitly via the optional PROJECT param."
   "Serialize DATA to FILENAME.
 
 The saved data can be restored with `projectile-unserialize'."
-  (when (file-writable-p filename)
+  (if (file-writable-p filename)
     (with-temp-file filename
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

PR #1734, opened 2021-12-13
Make projectile--find-matching-test Use src-dir/test-dir Properties

LaurenceWarne:
Hi!  This PR is in response to a discussion in https://github.com/bbatsov/projectile/issues/1650#issuecomment-987977517, basically it's a fix for the `src-dir` and `test-dir` properties being prematurely circumvented if they are set to strings when switching from/to test/impl files.  Essentially the function for finding a test/impl file using `src-dir` and `test-dir` (set as strings) was called in `projectile--find-matching-test`, but a fallback already existed in `projectile-find-implementation-or-test` which was being used in preference.

This PR hopes to fix and simplify the above by making impl/test switching wholly contained in `projectile--find-matching-file`/`projectile--find-matching-test`.

I've also updated the doc tweaking the `src-dir`/`test-dir` as a function example since I think that situation is better handled by setting `src-dir`/`test-dir` to strings (as pointed out by the issue author in the link).

Thanks!

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3385,8 +3385,10 @@ string) are replaced with the current project type's src-dir property
  (which should be a string) to obtain the new directory.
 
 Nil is returned if either the src-dir or test-dir properties are not strings."
-  (let ((test-dir (projectile-test-directory (projectile-project-type)))
-        (impl-dir (projectile-src-directory (projectile-project-type))))
+  (let ((test-dir (projectile-project-type-attribute
+                   (projectile-project-type) 'test-dir))
+        (impl-dir (projectile-project-type-attribute
+                   (projectile-project-type) 'src-dir)))
     (when (and (stringp test-dir) (stringp impl-dir))
       (projectile-complementary-dir test-dir-path test-dir impl-dir))))
 
@@ -3398,8 +3400,10 @@ string) are replaced with the current project type's test-dir property
  (which should be a string) to obtain the new directory.
 
 Nil is returned if either the src-dir or test-dir properties are not strings."
-  (let ((test-dir (projectile-test-directory (projectile-project-type)))
-        (impl-dir (projectile-src-directory (projectile-project-type))))
+  (let ((test-dir (projectile-project-type-attribute
+                   (projectile-project-type) 'test-dir))
+        (impl-dir (projectile-project-type-attribute
+                   (projectile-project-type) 'src-dir)))
     (when (and (stringp test-dir) (stringp impl-dir))
       (projectile-complementary-dir impl-dir-path impl-dir test-dir))))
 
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -3565,15 +3596,25 @@ concatenated with FILENAME-FN applied to the file name of FILE-PATH."
...
     (if-let ((plist (projectile--related-files-plist-by-kind impl-file :test)))
         (projectile--related-files-from-plist plist)
-      (when-let ((predicate (projectile--impl-to-test-predicate impl-file)))
-        (projectile--best-or-all-candidates-based-on-parents-dirs
-         impl-file (cl-remove-if-not predicate (projectile-current-project-files)))))))
+      (if-let ((test-file (projectile--test-file-from-test-dir-str impl-file)))
+          (list test-file)
+        (when-let ((predicate (projectile--impl-to-test-predicate impl-file)))
+          (projectile--best-or-all-candidates-based-on-parents-dirs
+           impl-file (cl-remove-if-not predicate (projectile-current-project-files))))))))
```
LaurenceWarne, line 3617: Basically the main change (and in `projectile--find-matching-file`).

test/projectile-test.el
```diff
@@ -1163,8 +1163,23 @@ Just delegates OPERATION and ARGS for all operations except for`shell-command`'.
...
+  (it "finds matching test with dirs and inexistent test file"
+    (projectile-test-with-sandbox
+      (projectile-test-with-files-using-custom-project
+          ("project/src/main/scala/bar/package.scala"
+           "project/src/main/scala/foo/package.scala"
+           "project/src/test/scala/foo/packageSpec.scala")
+          (:test-suffix "Spec" :test-dir "test" :src-dir "main")
+        (expect (projectile--find-matching-test
+                 "project/src/main/scala/bar/package.scala")
+                :to-equal '("src/test/scala/bar/packageSpec.scala")))))
```
LaurenceWarne, line 1182: This would fail before this change, since https://github.com/bbatsov/projectile/blob/2bb7ec28b1275bbce7cac743ee9e7b2cf41c5bbd/projectile.el#L3574 would take precedence over using the `src-dir`/`test-dir` properties.

LaurenceWarne:
Ah earlier Emacs version compat caught me out again :facepalm:, back to :hammer:

bbatsov:
Great PR! Thanks! 🙇‍♂️

LaurenceWarne:
Thanks!

dklimov-allstreet:
it looks like someone fucked things up

```elisp
(let* ((error-msg (format
                   "No matching test file found for project type `%s'"
                   (projectile-project-type)))
       (test-file (or (projectile-find-matching-test file-name)
                      (error error-msg))) ; <- bam!
       (expanded-test-file (projectile-expand-root test-file)))
  (cond ((file-exists-p expanded-test-file) expanded-test-file) ; we never get here if there's no test file 
        (projectile-create-missing-test-files 
         (projectile--create-directories-for expanded-test-file)
         expanded-test-file)
        (t (progn (error error-msg)))))
```

ever tried to test the code with `projectile-create-missing-test-files` enabled?

CC @bbatsov

bbatsov:
@dklimov-allstreet feel free to open a new issue or just send a PR addressing this. Looking at the problem the fix should be relatively simply.

PR #1745, opened 2022-01-29
Allow for Customisation of Project Type Precedence with projectile-update-project-type

LaurenceWarne:
Hi!  This PR allows you to change the precedence (ie if marker files match multiple project types) of a project type via another keyword argument.  This is not particularly sophisticated, done by just sticking the updated project type at the start of `projectile-project-types` and at the end for `'low`.  For example:

```elisp
(projectile-update-project-type 'sbt :precedence 'high)
```

(often I find my sbt projects get their project type set as bloop projects)  I've also added the minor change of allowing nil values to be passed to `projectile-update-project-type` to set project type attributes to nil (using `cl-defun` kwargs to distinguish between values not passed and values passed as nil).

-----------------

Thanks!


projectile.el
```diff
@@ -2763,35 +2776,44 @@ test/impl/other files as below:
...
-                                  p))
-                    projectile-project-types))
-    (error "No existing project found for: %s" project-type)))
+      (cl-flet* ((project-filter (p) (eq project-type (car p)))
+                 (project-map (p) (if (project-filter p) project-type-elt p)))
+        (setq projectile-project-types
+              (if precedence
+                  (let ((filtered-types
+                       (cl-remove-if #'project-filter projectile-project-types)))
+                    (setq projectile-project-type-cache (make-hash-table))
```
LaurenceWarne, line 2810: Invalidate the project type cache here, I think this is what we would want since otherwise users may have to remove projects from `projectile-project-type-cache` manually.

LaurenceWarne:
Thanks!
> I've been thinking that numeric precedences might work even better, as it will allow more granularity into the ordering, but that's probably an overkill given the people should normally run into such situation quite infrequently.

It could always be modified so that `precedence` can take a number but I agree, it's probably too niche a use case :slightly_smiling_face:

PR #1760, opened 2022-03-12
Fix wide docstrings

larslj:
The byte compiler in Emacs 28 show warnings for "docstring wider than 80 characters" in multiple places. These warnings are show every time the package is updated from list-packages so it's visible for the user.

I mostly just used fill-paragraph to fix these docstrings.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1230,7 +1230,8 @@ explicitly."
   (projectile-project-root (or dir default-directory)))
 
 (defun projectile-default-project-name (project-root)
-  "Default function used create project name to be displayed based on the value of PROJECT-ROOT."
+  "Default function used create project name to be displayed based on the
+value of PROJECT-ROOT."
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -1230,7 +1230,8 @@ explicitly."
   (projectile-project-root (or dir default-directory)))
 
 (defun projectile-default-project-name (project-root)
-  "Default function used create project name to be displayed based on the value of PROJECT-ROOT."
+  "Default function used create project name to be displayed based on the
```
bbatsov, line 1233: Checkdoc wants the first line of the docstring to be a complete sentence. This way we're solving one problem, but introducing another one.

larslj: There are numerous other checkdoc warnings as well. But I agree we should not make it worse. Some of the others ones are really easy to fix but I'll leave that out of this pull request.

bbatsov:
Thanks!

PR #1762, opened 2022-03-13
Escape periods in "projectile-globally-ignored-directories"

maxtrussell:
The unescaped periods were causing unexpected directories to be ignored.

E.g. for the following go project, the configitem directory was ignored because it matched the regex ".git":
```
$ find pkg -type f
pkg/metrics/argus.go
pkg/secrets/secrets.go
pkg/dynamo/dynamo.go
pkg/utils/types.go
pkg/utils/utils.go
pkg/utils/errors.go
pkg/configitem/configitem.go
pkg/api/handlers.go
pkg/api/middleware.go
```

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### Bugs fixed
 
+* [#1762](https://github.com/bbatsov/projectile/pull/1762): Fix `projectile-globally-ignored-directories` unescaped regex
 * [#1713](https://github.com/bbatsov/projectile/issues/1731): Fix `projectile-discover-projects-in-directory` reordering known projects.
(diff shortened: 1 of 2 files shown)


bbatsov:
Can you elaborate on this a bit? I was under the impression the items on this list were used for literal matches, but if they are used as regular expressions we can make them more specific.

bbatsov:
> It's probably worth adding start and end anchors to projectile-globally-ignored-directories so that wacky directory names like "foo.git.bar" won't be ignored

Agreed.

maxtrussell:
> > It's probably worth adding start and end anchors to projectile-globally-ignored-directories so that wacky directory names like "foo.git.bar" won't be ignored
> 
> Agreed.

Done

bbatsov:
Thanks!

Issue #1600, opened 2020-11-17
Error when using git submodules with projectile


sonictk:
# Expected behavior
That `projectile-find-file` works.
## Actual behavior

I get an error. It seems like formatting of the final command to call has a bug with it on Windows:

```
Debugger entered--Lisp error: (file-missing "Setting current directory" #("No such file or directory" 0 25 (charset windows-1252)) "c:/Users/REDACTED/Git/REDACTED/'echo: -c: line 0: unexpe...")
  call-process("C:/emacs27_1/libexec/emacs/27.1/x86_64-w64-mingw32..." nil t nil "-c" "git submodule --quiet foreach 'echo $path' | tr '\\...")
  call-process-shell-command("git submodule --quiet foreach 'echo $path' | tr '\\..." nil t)
  shell-command("git submodule --quiet foreach 'echo $path' | tr '\\..." t)
  shell-command-to-string("git submodule --quiet foreach 'echo $path' | tr '\\...")
  (split-string (shell-command-to-string command) "\0" t)
  (let ((default-directory root)) (split-string (shell-command-to-string command) "\0" t))
  (progn (let ((default-directory root)) (split-string (shell-command-to-string command) "\0" t)))
  (if (stringp command) (progn (let ((default-directory root)) (split-string (shell-command-to-string command) "\0" t))))
  projectile-files-via-ext-command("c:/Users/REDACTED/Git/REDACTED/'echo: -c: line 0: unexpe..." "git submodule --quiet foreach 'echo $path' | tr '\\...")
  (mapcar #'(lambda (s) (file-name-as-directory (expand-file-name s path))) (projectile-files-via-ext-command path (projectile-get-sub-projects-command vcs)))
  (let* ((vcs (projectile-project-vcs path)) (submodules (mapcar #'(lambda (s) (file-name-as-directory (expand-file-name s path))) (projectile-files-via-ext-command path (projectile-get-sub-projects-command vcs)))) (project-child-folder-regex (concat "\\`" (regexp-quote path)))) (cl-remove-if-not #'(lambda (submodule) (string-match-p project-child-folder-regex submodule)) submodules))
  projectile-get-immediate-sub-projects("c:/Users/REDACTED/Git/REDACTED/'echo: -c: line 0: unexpe...")
  (let ((submodules (projectile-get-immediate-sub-projects project))) (cond ((null submodules) nil) (t (nconc submodules (projectile-flatten (mapcar #'(lambda ... ...) submodules))))))
[...]

PR #1764, opened 2022-03-23
Fix Git submodule command on Windows

shivjm:
Projectile produces an error when finding project files in a Git repository with submodules on Windows (#1600). This is because the command to find the submodules is hard-coded, including the quotes. This PR adds a `projectile-get-git-sub-projects-command` function to generate the command with the correct quoting at runtime. It also adds `projectile-git-submodule-command-function`, which is set to `projectile-get-git-sub-projects-command` by default, obsoleting `projectile-git-submodule-command`.

I’m not sure I can add tests for this since it’s specific to the combination of Git, submodules, and Windows. I did run `checkdoc` but it showed a whole lot of unrelated errors.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -672,6 +672,17 @@ Set to nil to disable listing submodules contents."
   :group 'projectile
   :type 'string)
 
+(make-obsolete-variable 'projectile-git-submodule-command "Configure `projectile-git-submodule-command-function' instead." "2.6.0")
+
+(defcustom projectile-git-submodule-command-function 'projectile--get-git-sub-projects-command
+  "Function generate command to list submodules of a given git repository.
+Set to nil to disable listing submodules contents.
+
+This has to be a function to generate the command with the correct quotes."
+  :group 'projectile
+  :type 'function
+  :package-version '(projectile . "2.6.0"))
+
 (defcustom projectile-git-ignored-command "git ls-files -zcoi --exclude-standard"
   "Command used by projectile to get the ignored files in a git project."
   :group 'projectile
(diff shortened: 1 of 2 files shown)


bbatsov:
I'm not particularly concerned about the tests here, but I'm wondering if there's a simpler way to fix this. Also we'll need to be extra clear in the function docstrings as to why this is not a simple string as most other commands. The public docs have to be updated as well to reflect the change.

projectile.el
```diff
@@ -654,6 +654,14 @@ Set to nil to disable listing submodules contents."
   :group 'projectile
   :type 'string)
 
+(make-obsolete-variable 'projectile-git-submodule-command "Configure `projectile-git-submodule-command-function' instead." "2.6.0")
+
+(defcustom projectile-git-submodule-command-function 'projectile-get-git-sub-projects-command
+  "Function generate command to list submodules of a given git repository.
+Set to nil to disable listing submodules contents."
+  :group 'projectile
+  :type 'function)
```
bbatsov, line 663: This also have a `package-version` property.

projectile.el
```diff
@@ -1342,9 +1350,16 @@ Fallback to a generic command when not in a VCS-controlled project."
 Currently that's supported just for Git (sub-projects being Git
 sub-modules there)."
   (pcase vcs
-    ('git projectile-git-submodule-command)
+    ('git (and projectile-git-submodule-command-function (funcall projectile-git-submodule-command-function)))
     (_ "")))
 
+(defun projectile-get-git-sub-projects-command ()
```
bbatsov, line 1356: Probably this should be marked as private and also explain why the command has to be generated at runtime.

bbatsov:
One thing that's not clear to me is whether this needs to be generated every time or can be generated just the first time (which would mean you can simply move the `format` in the original defcustom.

shivjm:
> One thing that's not clear to me is whether this needs to be generated every time or can be generated just the first time (which would mean you can simply move the `format` in the original defcustom.

I also thought of this, and the only concern I had is that (if I’ve understood correctly) it would cause problems if the same configuration were reused across platforms. Otherwise I definitely think it would be better to just generate it once.

projectile.el
shivjm: Does this make sense?
```emacs-lisp
(defun projectile--get-git-sub-projects-command ()
  "Get the sub-projects command for Git.
Defined as a function in order to generate the properly-quoted command at
runtime."
```

bbatsov:
> I also thought of this, and the only concern I had is that (if I’ve understood correctly) it would cause problems if the same configuration were reused across platforms.

I don't quite understand what you mean by this. Can you elaborate a bit?

shivjm:
My understanding is that the custom variable will be processed once. Say you’re running Emacs on Linux. The custom variable will be set accordingly. If you then reuse that config, including the custom settings, for a Windows machine, the custom variable won’t be updated—it’ll remain in Linux form.

Assuming this is so, that might simply be an edge case we can ignore. I don’t know how important it is.

bbatsov:
Your diagnosis was correct, but the quoting was only half the Windows problem (the `tr` pipeline still breaks stock cmd.exe). Master now lists submodules without any shell at all, via `git config -z --file .gitmodules` (69d4d20), which fixes #1600 on every platform.

shivjm:
Ah, I see. Well, thanks for the investigation and the fix!

Issue #1676, opened 2021-06-17
Unable to completing-read CMake preset the second time


laurynas-biveinis:
## Expected behavior

`projectile-configure-project`, when run second time in a project, has some way to select a (different) CMake preset again.

## Actual behavior

`Configure command: cmake . --preset <previously-select-preset>` prompt in minibuffer without completion for presets.

## Steps to reproduce the problem

- Have a CMake project with at least two configure presets
- `s-p C`, select one of the presets
- `s-p C` again, wishing to select the other preset

## Environment & Version information

CMake 3.20.4

### Projectile version information

```
Projectile 2.4.0
```

With a workaround for https://github.com/bbatsov/projectile/issues/1673 applied:
```elisp
(defun dotfiles--projectile--cmake-all-command-presets (command-type)
  "Get CMake user and system COMMAND-TYPE presets."
  (projectile-flatten
   (mapcar (lambda (filename)
             (projectile--cmake-command-presets filename command-type))
           '("CMakeUserPresets.json" "CMakePresets.json"))))
(advice-add #'projectile--cmake-all-command-presets :override
            #'dotfiles--projectile--cmake-all-command-presets)
```

### Emacs version

27.2

### Operating system

macOS 11.4

jehelset:
i think this is simply how the project-type-agnostic`projectile-configure-command` works (https://github.com/bbatsov/projectile/blob/master/projectile.el#L4337). it will prefer to use the cached command. to get the behaviour you want, one would have to be able to disable caching of configuration commands. not sure if this is possible currently.

jehelset:
as a quickfix you could use something like this:
```elisp

(defun projectile-reconfigure-command (compile-dir)
  "Retrieve the configure command for COMPILE-DIR without considering history.

The command is determined like this:

- first we check for `projectile-project-configure-cmd' supplied
via .dir-locals.el

- finally we check for the default configure command for a
project of that type"
  (or projectile-project-configure-cmd
      (let ((cmd-format-string (projectile-default-configure-command (projectile-project-type))))
        (when cmd-format-string
          (format cmd-format-string (projectile-project-root) compile-dir)))))

(defun projectile-reconfigure-project (arg)
  "Run project configure command without considering command history.

Normally you'll be prompted for a compilation command, unless
variable `compilation-read-command'.  You can force the prompt
with a prefix ARG."
  (interactive "P")
  (let ((command (projectile-reconfigure-command (projectile-compilation-dir))))
[...]

PR #1768, opened 2022-04-02
Add variable to enable or disable command caching.

jehelset:
For a CMake-project with a preset-driven workflow, the caching of commands is not necessary (https://github.com/bbatsov/projectile/issues/1676). This PR adds functionality for disable caching.

We add a variable `projectile-project-enable-cmd-caching` so that the user can enable or disable caching on a per-project basis by setting this variable in `.dir-locals.el`. If the variable is set to `nil`, then `nil` is passed as command map to `projectile--run-project-cmd` so that no commands are cached. The variable is `t` by default.

The variable is checked with a function `projectile--cache-project-commands-p` which ensures that directory local variables for the project are loaded.

I didn't update the readme, but dropped a line in `projects.adoc`.

-----------------

Thanks!

### DIFF
--- a/doc/modules/ROOT/pages/projects.adoc
+++ b/doc/modules/ROOT/pages/projects.adoc
@@ -774,3 +774,7 @@ external command or an Emacs Lisp function:
 ----
 (setq projectile-test-cmd #'custom-test-function)
 ----
+
+In addition caching of commands can be disabled by setting the variable
+`projectile-project-enable-cmd-caching` is to `nil`. This is useful for
+preset-based CMake projects.
--- a/projectile.el
+++ b/projectile.el
@@ -4599,6 +4599,16 @@ directory to open."
   (make-hash-table :test 'equal)
   "A mapping between projects and the last run command used on them.")
 
+(defvar projectile-project-enable-cmd-caching t
+  "Enables command caching for the project.  Set to nil to disable.
+Should be set via .dir-locals.el.")
+
+(defun projectile--cache-project-commands-p ()
+  "Whether to cache project commands."
+  (with-temp-buffer
+    (hack-dir-local-variables-non-file-buffer)
+    projectile-project-enable-cmd-caching))
+
 (defvar projectile-project-configure-cmd nil
   "The command to use with `projectile-configure-project'.
 It takes precedence over the default command for the project type when set.
@@ -4923,8 +4933,9 @@ Normally you'll be prompted for a compilation command, unless
 variable `compilation-read-command'.  You can force the prompt
 with a prefix ARG."
   (interactive "P")
-  (let ((command (projectile-configure-command (projectile-compilation-dir))))
-    (projectile--run-project-cmd command projectile-configure-cmd-map
+  (let ((command (projectile-configure-command (projectile-compilation-dir)))
+        (command-map (if (projectile--cache-project-commands-p) projectile-configure-cmd-map)))
+    (projectile--run-project-cmd command command-map
                                  :show-prompt arg
                                  :prompt-prefix "Configure command: "
                                  :save-buffers t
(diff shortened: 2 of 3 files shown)


jehelset:
I guess it would still be nice if the previously used preset was what the `completing-read` would default to, but didn't see an easy way to implement that.

jehelset:
Hm, this doesn't seem to work if you run a projectile command when the compilation-buffer is active, because it doesn't seem to pick up the directory-local variables then. So the command gets cached. 

After this, even if you do a command with a file-buffer active, because the command retrieval functions are not parametrized on the command-map, they will do unconditional lookups and always find the old command.

Not sure how to deal with that in a good way.

jehelset:
Ok, I found a hack that uses `with-temp-buffer` in `projectile-switch-project-by-name` that seems to work.

projectile.el
```diff
@@ -4493,6 +4493,16 @@ directory to open."
   (make-hash-table :test 'equal)
   "A mapping between projects and the last run command used on them.")
 
+(defvar projectile-project-enable-cmd-caching t
+  "Enables command caching for the project.  Set to nil to disable.
+Should be set via .dir-locals.el.")
+
+(defun projectile--cache-project-commands-p ()
```
bbatsov, line 4500: Is this really necessary? Seems to me that the local variable would work fine in most cases even without it.

doc/modules/ROOT/pages/projects.adoc
```diff
@@ -774,3 +774,6 @@ external command or an Emacs Lisp function:
 ----
 (setq projectile-test-cmd #'custom-test-function)
 ----
+
+In addition caching of commands can be disabled by setting the variable
```
bbatsov, line 778: Might be good to mention that CMake projects are most likely to benefit from this.

bbatsov:
Also we can maybe disable the command caching with a double prefix argument or something like this, so people can do this on demand.

doc/modules/ROOT/pages/projects.adoc
jehelset: Added a line.

projectile.el
jehelset: my understanding is that it will fail in the case where the f.ex. *compilation* buffer is selected. this
is not too uncommon for me personally, and noticed it quite fast with my initial attempt which did
not do this.

it's not necessary, but it's more ergonomic. up to you.

jehelset:
> Also we can maybe disable the command caching with a double prefix argument or something like this, so people can do this on demand.

do you mean instead of a variable? yes, that might work. however i do think it would be nice to
configure the default (preferrably on a command-type-by-command-type-basis). with presets you usually do not
want to cache the configure command (the original issue was about this: #1676). 

i think for presets it would be more beneficial to cache the last selected preset and pass it as
`INITIAL-INPUT` to `completing-read` rather than caching the command itself.

bbatsov:
I meant this as something additional - a quick way to avoid the command cache if you don't need it (sometimes the default is a good default, but you still want to do something different occasionally). Might be an overkill, though.

projectile.el
bbatsov: Okay, fair enough. We can keep this.

bbatsov:
@jehelset I guess we can keep this as is, in the interest of moving forward with the PR. Just get the PR to pass and we're in business.

powel-jeh:
> @jehelset I guess we can keep this as is, in the interest of moving forward with the PR. Just get the PR to pass and we're in business.

ah, i have been a bit busy. i think the double prefix is a nice idea, and i it on my todo list. i can do that in a separate pr though.

jehelset:
> @jehelset I guess we can keep this as is, in the interest of moving forward with the PR. Just get the PR to pass and we're in business.

ah! i have been a bit busy. i think double-prefix was a nice idea, and i had it on my todo list. i can do that in a separate pr though.

bbatsov (approved):

bbatsov:
Yeah, it can be implemented in a separate PR. Thanks for tacking this!

Issue #1765, opened 2022-03-24
`projectile-toggle-between-implementation-and-test results` in `or: No matching test file found for project type ‘lein-test’` when the corresponding test file doesn't exist


dklimov-allstreet:
## Expected behavior

projectile creates a test file

## Actual behavior

I get an error message `or: No matching test file found for project type ‘lein-test’` when there's no test file 

## Steps to reproduce the problem

- create a clojure project using `lein new` 

- find a new clj file deep inside `src/` dir

- run M-x `projectile-toggle-between-implementation-and-test results`

### Projectile version information
```
Projectile version: 2.6.0-snapshot
```

### Emacs version
```
GNU Emacs 26.3 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.14) of 2020-03-26, modified by Debian
```

### Operating system

Ubuntu 20.04

LaurenceWarne:
I think an error here is right, but the error message is wrong.  Given an impl file without a corresponding test (ie the situation you describe) to determine a test file location we need some way of determining a test file location: so one of `test-dir` (with `src-dir`) and/or `related-files-fn` should be set for that project type.  But none of these are set for the `lein-test` project type:
```elisp
(projectile-register-project-type 'lein-test '("project.clj")
                                  :project-file "project.clj"
                                  :compile "lein compile"
                                  :test "lein test"
                                  :test-suffix "_test")
```

The following for example would string replace `src` for `test` to get the test file name.
```elisp
(projectile-update-project-type 'lein-test
                                :src-dir "src"
                                :test-dir "test")
```
Testing the above locally with `projectile-create-missing-test-files` set inside `foo/src/core.clj` sticks me in `foo/test/foobar_test.clj`.  If this is the standard for `lein` projects (I don't know if it is) the I'd say this should be set (alongside a better error message, something like `'projectile-create-missing-test-files' is set, but cannot determine directory of test file for 'foo.bar', set one of 'foo'/'bar' for this project type to make this explicit`).

But there was a behaviour change in https://github.com/bbatsov/projectile/pull/1734, prior to that PR if a project type didn't have a `src-dir` and `test-dir` property, these were assumed to be `src` and `test` respectively (apologies I don't recall if it was deliberate or if it slipped my mind at the time).  If that's the desired behaviour, I'm happy to PR it back in.

dklimov-allstreet:
> If that's the desired behaviour, I'm happy to PR it back in.

[...]

PR #1769, opened 2022-04-04
Resurrect Default src-dir and test-dir for project-types

LaurenceWarne:
Hi, this is a proposed fix for #1765, it (re) adds a default value for the `src-dir` and `test-dir` properties for project types (= `"src/"` and `"test/"`).  The main effect is for `projectile-toggle-between-implementation-and-test`, so these defaults will be used as a fallback if no other method can be found for finding an impl/test file.

There was a breaking change in https://github.com/bbatsov/projectile/pull/1734 (sorry!) which removed this fallback, so project types without a `src-dir` or `related-files-fn` property wouldn't have a method for finding corresponding test/impl files, as shown in #1765.  This fallback is not always particularly useful however, for example when your implementation file doesn't contain `"src/"`, this PR throws an error in such cases.

I've also tweaked some error messages around this area to hopefully make debugging a little easier:

| Error | Old Message | New Message |
| --- | --- | --- |
|No method was found for determining the (base) file name for a test/impl file | `"Project type foo not supported!"` | `"Cannot determine a test file name, one of "test-suffix" or "test-prefix" must be set for project type foo"`(or an equivalent message for test -> impl) |
|An appropriate test file was found for an impl file, but `projectile-create-missing-test-files` is not set |`"No matching test file found for project type foo"` | ``"Determined test file to be "foo", which does not exist. Set `projectile-create-missing-test-files' to allow `projectile-find-implementation-or-test' to create new files"`` |
|`src-dir` and `test-dir` are set for a project type, but the impl/test file does not contain these as substrings|No error|`"Attempted to find a test file by switching this project type's (foo) src-dir property "src" with this project type's test-dir property "test", but foo/bar does not contain "src"` (or an equivalent message for test -> impl) |

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -530,6 +530,22 @@ See also `projectile-update-mode-line'."
   :type 'function
   :package-version '(projectile . "2.0.0"))
 
+(defcustom projectile-default-src-directory "src/"
+  "The default value of a project's src-dir property.
+
+It's used as a fallback in the case the property is not set for a project
+type when `projectile-toggle-between-implementation-and-test' is used."
+  :group 'projectile
+  :type 'string)
+
+(defcustom projectile-default-test-directory "test/"
+  "The default value of a project's test-dir property.
+
+It's used as a fallback in the case the property is not set for a project
+type when `projectile-toggle-between-implementation-and-test' is used."
+  :group 'projectile
+  :type 'string)
+
 
 ;;; Idle Timer
 (defvar projectile-idle-timer nil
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -3674,6 +3749,25 @@ to a custom function, else return nil."
         #'projectile--test-name-for-impl-name)
        (projectile-project-root)))))
 
+(defmacro projectile--acond (&rest clauses)
```
LaurenceWarne, line 3752: I've added `acond` from `anaphora.el` (though with a tweak for emacs 25 compat) in an effort to make `projectile--find-matching-file` and `projectile--find-matching-test` a bit nicer.  If you prefer it I can remove it and go back to using the chain of `if-let`s.

projectile.el
```diff
@@ -530,6 +530,22 @@ See also `projectile-update-mode-line'."
   :type 'function
   :package-version '(projectile . "2.0.0"))
 
+(defcustom projectile-default-src-directory "src/"
+  "The default value of a project's src-dir property.
```
LaurenceWarne, line 534: Sorry I'm not sure the convention for `:package-version`, do you wait until a new release before adding this?

bbatsov:
The fix looks good. Thanks for tackling this!

LaurenceWarne:
Thanks :slightly_smiling_face:

Issue #1654, opened 2021-03-03
History in projectile-run-project lists all duplicates


doctordesh:
(disclaimer, I don't know if this is a bug or intentional behaviour, but since this was a bug template I used it)

## Expected behavior

* Given a history of different commands run with `projectile-run-project`, where some commands have been used multiple times in a row (like: `<cmd1>`, `<cmd1>`, `<cmd1>`, `<cmd2>`, `<cmd1>`, `<cmd1>` ...)
* When using `M-p` and `M-n` to move in the history
* Then projectile should only move between "unique" entries and ignores the duplicates

**Informal example**

I made 1 call to do `go mod vendor` then ~15 calls to `go run main.go`. I had to use `M-p` 15 times to get back to `go mod vendor`

## Actual behavior

Projectile keeps all the commands in the history, regardless of uniqueness. 

## Steps to reproduce the problem

* Run `<cmd1>` 1 time
* Run `<cmd2>` 10 times
* Use `M-p` 10 times to move up to <cmd1>

### Projectile version information

```
Projectile version: 20210104.1216
```

### Emacs version

27.1

### Operating system

macOS

bbatsov:
Projectile simply delegates to `complilation-mode` for the command execution. Perhaps there's some way to instruct it to keep only unique commands in its history.

PR #1774, opened 2022-05-02
Don't add Duplicate Commands to Command History

LaurenceWarne:
Hi, this PR makes it so that duplicate commands are not added to the command history (a command is only added to the history ring if it doesn't equal the most recent command).  

Related is: https://github.com/bbatsov/projectile/issues/1654, but it has the same effect for `projectile-test-project`, `projectile-compile-project` and so on.

Thanks!

-----------------

Thanks!


bbatsov:
Thanks!

PR #1778, opened 2022-05-29
Support `projectile-replace` to select file extension on C-u

taquangtrung:
Hi,

I created this PR to enable users to select file name pattern and extension when performing `projectile-replace` with `C-u`.

This is to prevent `projectile-replace` from searching on very large files of a project (my Emacs hangs when projectile search on a 1MB Javascript library file, while I only wanted to search and replace on Rust code).

Currently, the file extension filtering is supported only by `ag` (using the option `-G <file-ext-pattern>`). 

I don't use `rg`, hence I haven't been able to add the corresponding option for it (and also for `ack` and `grep`).

Can you take a look and advise if this PR can be merged?

Thank you!

-----------------

- ~~[ ] You've added tests (if possible) to cover your change(s)~~
- [x]  The new code is not generating bytecode or `M-x checkdoc` warnings

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4468,8 +4468,8 @@ files in the project."
 (defun projectile-replace (&optional arg)
   "Replace literal string in project using non-regexp `tags-query-replace'.
 
-With a prefix argument ARG prompts you for a directory on which
-to run the replacement."
+With a prefix argument ARG prompts you for a directory and file name patterns
+on which to run the replacement."
   (interactive "P")
   (let* ((directory (if arg
                         (file-name-as-directory
--- a/projectile.el
+++ b/projectile.el
@@ -4410,11 +4410,12 @@ to run the replacement."
                          (read-directory-name "Replace in directory: "))
                       (projectile-acquire-root)))
          (file-ext (if arg
-                        (if (fboundp #'helm-grep-get-file-extensions)
-                            (car (helm-grep-get-file-extensions (list directory)))
-                          (read-string
-                           (format "Replace in files with extension: ")))
-                      nil))
+                       (if (fboundp #'helm-grep-get-file-extensions)
+                           (car (helm-grep-get-file-extensions (list directory)))
+                         (read-string
+                          (projectile-prepend-project-name
+                           "Replace in files with extension: ")))
+                     nil))
          (old-text (read-string
                     (projectile-prepend-project-name "Replace: ")
                     (projectile-symbol-or-selection-at-point)))
--- a/projectile.el
+++ b/projectile.el
@@ -4358,7 +4358,7 @@ Returns a list of expanded filenames."
     ;; -I: no binary files
     (grep . "grep -rHlI %s .")))
 
-(defun projectile--rg-construct-command (file-ext search-term)
+(defun projectile--rg-construct-command (search-term &optional file-ext)
   "Construct Rg option to search files by the extension FILE-EXT."
   (if (stringp file-ext)
       (concat (cdr (assoc 'rg projectile-files-with-string-commands))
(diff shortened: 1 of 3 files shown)


bbatsov:
I like the idea in general, but I think it's a very bad idea to implement something just for one search tool, as it's confusing for the end users if something is available with `ag`, but not with other tools. Do the other tools rely only on shell globbing for filtering?

taquangtrung:
@bbatsov: Thanks for the suggestion! I agree on that point.

I added the support for file extension filtering for all searchers now.

I also tested all of them manually in my Linux machine, and they all work well.

projectile.el
```diff
@@ -4348,17 +4348,77 @@ Returns a list of expanded filenames."
...
+    (ack . "ack --literal --nocolor -l ")
     (git . "git grep -HlI ")
     ;; -r: recursive
     ;; -H: show filename for each match
     ;; -l: show only file names with matches
     ;; -I: no binary files
     (grep . "grep -rHlI %s .")))
 
-(defun projectile-files-with-string (string directory)
+(defun projectile--rg-construct-command (file-ext search-term)
```
bbatsov, line 4361: I think the params should be the other way around and `file-ext` should be optional in all of them (`&optional`).

projectile.el
```diff
@@ -4398,13 +4454,20 @@ to run the replacement."
                         (file-name-as-directory
                          (read-directory-name "Replace in directory: "))
                       (projectile-acquire-root)))
+         (file-ext (if arg
```
bbatsov, line 4457: Perhaps instead of reusing the other arg you want to use double prefix arg or something like this here? At any rate you'll also have to update the command doc, otherwise that would seem weird.

bbatsov:
The changes look reasonable to me, but you'll also have to mention them somewhere in the online docs, otherwise it's unlikely people will find this behavior. 

And the tests seem to be failing now, so this has to be fixed.

projectile.el
taquangtrung: Thanks! I updated the new code.

projectile.el
taquangtrung: I think that using the double prefix arg is generally not natural: the user needs to pause to think about whether to use `C-u` or `C-u C-u`.

Instead, I revise the code to print a more suggestive message saying that if the users just need to enter an empty string to select all file extensions:

```
(file-ext (if arg
                       (if (fboundp #'helm-grep-get-file-extensions)
                           (car (helm-grep-get-file-extensions (list directory)))
                         (read-string
                          (projectile-prepend-project-name
                           "With file extension (empty string means all files): ")))
                     nil))
```

How do you think about that?

taquangtrung:
> The changes look reasonable to me, but you'll also have to mention them somewhere in the online docs, otherwise it's unlikely people will find this behavior.
> 
> And the tests seem to be failing now, so this has to be fixed.

Sorry, it was my mistake in the previous commit that I accidentally commented out a line of code, which lead to the failed tests.

Can you run the GitHub action again?

I also updated the CHANGELOG file. Where can I update the online documentation?

bbatsov:
> I also updated the CHANGELOG file. Where can I update the online documentation?

It's generated from the .adoc files in the `docs` folder.

bbatsov:
Ping :)

taquangtrung:
@bbatsov: sorry for the long delay! I just updated the documentation.

bbatsov:
@taquangtrung It seems the reference to `helm-grep-get-file-extensions` is breaking the CI. You might have to declare it or something like this.

taquangtrung:
Thanks! I just updated it.

bbatsov:
Thanks!

PR #1779, opened 2022-06-01
[Fix #1755] Cache failure to find project root

### LINKED ISSUE #1755: projectile-project-root is slow when not in project - not caching negative results?
## Expected behavior
Calling `projectile-project-root` in a deep subdir that's not in a project should be fast, especially after the first call

## Actual behavior
It can be slow; .5 to 1 sec per call.

I have a dir `/mnt/c/files/consulting/project/projectA/` containing `foo.txt`. In that buffer, invoke `(projectile-project-root)`. It takes about half a second on a fast machine.
The reason seems to be that the negative project result does not get cached, so every time it's called it calls `projectile-root-top-down`, which is slow.

The code in `projectile.el` has a comment indicating that negative results should be cached as `'none`, but that does not seem to be actually happening:
```lisp
                  ;; set cached to none so is non-nil so we don't try
                  ;; and look it up again
```

### To reproduce: 
Start `emacs -q -l /tmp/emacs-test-startup.el` with this `emacs-test-startup.el`:
```lisp
;;; Set up package system -- straight.el
(defvar bootstrap-version)
(or (boundp 'native-comp-deferred-compilation-deny-list)
    (setq native-comp-deferred-compilation-deny-list '()))

(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

;;; Meta-package system: use-package. Auto-installs and configures packages.
(straight-use-package 'use-package)
(defvar straight-use-package-by-default)
(setq straight-use-package-by-default t) ; make use-package use straight

(use-package projectile
  :config
  (projectile-mode))
```
Then create and find the file `/mnt/c/files/consulting/project/projectA/foo.txt`. Invoke `(projectile-project-root)`; it should take around half a second. Then invoke it again; it will still take half a second.

If you edebug `projectile-project-root` and step through it, you can see it never caches the `'none` value so each time it tries all the `projectile-project-root-functions`, and some of them (e.g. `projectile-root-top-down`, which calls `projectile-locate-dominating-file`) are slow.

Linux Emacs nightly of 2022-01-24, with native comp enabled.
Running in WSL2 on Windows 11 with an AMD Ryzen Threadripper CPU.

25.0, latest release

*Include here the version string displayed by `M-x
projectile-version`.*

```
Projectile nil
```
(I use straight.el; I have the tag `25.0`.)

### Emacs version

GNU Emacs 29.0.50 (build 3, x86_64-pc-linux-gnu, GTK+ Version 3.24.20,
cairo version 1.16.0) of 2022-01-24

### Operating system
WSL2, Ubuntu 20.04, on Windows 11

tomheon:
When `projectile-project-root` was unable to find a project root for a
directory, it would not cache the negative result, so future invocations would
repeat the (often expensive) search for a project root every time.

With this change, `projectile-project-root` caches failure to find a project
root, when that failure is expected to be permanent, and consults that cache
for speed on subsequent invocations.  "Expected to be permanent" means that
either we're trying to find the project root of a local directory, or we're
successfully connected to a remote directory via TRAMP.  If the directory isn't
local, but we can't connect to it, we consider that a transient failure and
don't cache it.

Under the hood, this change uses the same `projectile-project-root-cache`
that's used to cache successful attempts to find a project root, but with a new
key `projectilerootless-{dir}`.  This should allow cache invalidation to work
as expected.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * [#1765](https://github.com/bbatsov/projectile/issues/1765): Fix `src-dir`/`test-dir` not defaulting to `"src/"` and `"test/"` with `projectile-toggle-between-implementation-and-test`.
 * Fix version extraction logic.
 * [1654](https://github.com/bbatsov/projectile/issues/1654) Fix consecutive duplicates appearing in command history
+* [#1755](https://github.com/bbatsov/projectile/issues/1755) Cache failure to find project root
 
 ### Changes
 
--- a/projectile.el
+++ b/projectile.el
@@ -1189,42 +1189,53 @@ topmost sequence of matched directories.  Nil otherwise."
 (defun projectile-project-root (&optional dir)
   "Retrieves the root directory of a project if available.
 If DIR is not supplied its set to the current directory by default."
-  ;; the cached value will be 'none in the case of no project root (this is to
-  ;; ensure it is not reevaluated each time when not inside a project) so use
-  ;; cl-subst to replace this 'none value with nil so a nil value is used
-  ;; instead
   (let ((dir (or dir default-directory)))
     ;; Back out of any archives, the project will live on the outside and
     ;; searching them is slow.
     (when (and (fboundp 'tramp-archive-file-name-archive)
                (tramp-archive-file-name-p dir))
       (setq dir (file-name-directory (tramp-archive-file-name-archive dir))))
+    ;; the cached value will be 'none in the case of no project root (this is to
+    ;; ensure it is not reevaluated each time when not inside a project) so use
+    ;; cl-subst to replace this 'none value with nil so a nil value is used
+    ;; instead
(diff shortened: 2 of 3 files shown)


bbatsov:
@tomheon The changes look good. Don't forget to reflect them in the changelog.

tomheon:
@bbatsov Changelog updated, let me know if there's anything else that needs attention.

bbatsov:
Thanks!

PR #1781, opened 2022-06-18
Switch rails-rspec src-dir

lackita:
Rails puts most of its source files under app rather than lib. This
updates the src-dir to allow easy switching between source and test
files in this case.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Bugs fixed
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Makes sense. I guess we should do the same for `rails-minitest` as well.

lackita:
Good call, I'll try to change that tomorrow

lackita:
Alright, I think that's all set now. Let me know if you want me to change anything else.

bbatsov:
Thanks!

Issue #1783, opened 2022-07-01
Indexing not ignoring folders.


acristoffers:
## Expected behavior

`projectile-find-file` should ignore files in the directory `build` if I add `"build"` to the variable `projectile-globally-ignored-directory`.

## Actual behavior

When the indexing mode is `alien`, entries like `"build"` do not work whereas `"*build"` works.

## Steps to reproduce the problem

Run `(add-to-list 'projectile-globally-ignored-directories "build")` and then try to search for files using `projectile-find-file` on a project with a populated `build` folder. The files in that folder should be ignored, but they won't.

## Environment & Version information

```
GNU Emacs     v28.1            nil
Doom core     v3.0.0-dev       HEAD -> master c2f8476c8 2022-06-29 18:12:43 +0200
Doom modules  v22.06.0-dev     HEAD -> master c2f8476c8 2022-06-29 18:12:43 +0200
```

### Projectile version information

```
Projectile version: 2.6.0-snapshot
```

### Emacs version

GNU Emacs     v28.1

### Operating system

macOS 12.4

### Possible problem root

The function `projectile-dir-files-alien` returns a list of files starting with `./`, but then the function `projectile-remove-ignored` tries to compare it with the name of the folders as found in the variable, that is, it is comparing (using `string-prefix-p`, line 1504 for me) `vendor` with `./vendor/filename`, and therefore falling. If `projectile-dir-files-alien` returned `vendor/filename` or the `./` was removed before the comparison, it should work.

bbatsov:
> When the indexing mode is alien, entries like "build" do not work whereas "*build" works.

Hmm, but alien intentionally doesn't do any filtering of anything by defign. The ignore rules are consulted only for hybrid and native indexing.

acristoffers:
This is was fixed already with #1784, we forgot to close this issue.

PR #1784, opened 2022-07-01
Fixes projectile-dir-files-alien on macOS

acristoffers:
As I describe in #1783, the function `projectile-dir-files-alien` is not returning the same value in macOS as the function `projectile-dir-files-native` because it returns values with a leading "./". This PR fixes this, ensuring both functions return values without a leading dot directory and fixing the issue reported in https://github.com/doomemacs/doomemacs/issues/6504.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -716,10 +716,10 @@ Set to nil to disable listing submodules contents."
   (cond
    ;; we prefer fd over find
    ((executable-find "fd")
-    "fd . -0 --type f --color=never")
+    "fd . -0 --type f --color=never --strip-cwd-prefix")
    ;; fd's executable is named fdfind is some Linux distros (e.g. Ubuntu)
    ((executable-find "fdfind")
-    "fdfind . -0 --type f --color=never")
+    "fdfind . -0 --type f --color=never --strip-cwd-prefix")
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -1358,11 +1358,13 @@ IGNORED-DIRECTORIES may optionally be provided."
 (defun projectile-dir-files-alien (directory)
   "Get the files for DIRECTORY using external tools."
   (let ((vcs (projectile-project-vcs directory)))
-    (cond
+    (-map
+     (lambda (x) (string-remove-prefix "./" x))
```
bbatsov, line 1362: Such a check would be extremely expensive on a big project, therefore I'd recommend to solve this at a different level (e.g. the command invocation itself)

acristoffers: Agree. At first, I thought that, since you have at least 8 commands which can be used, doing it here would be safe to make sure all of them are consistent. However, given the comment above the `find` command, it seems that care to make sure they were consistent was taken, and `fd` not following the pattern was just a slip.

acristoffers:
Are the new changes better?

bbatsov:
Yeah, this looks good to me.

Issue #1788, opened 2022-07-28
`--strip-cwd-prefix` for `fd` doesn't work on versions <v8.3.0


J3RN:
## Expected behavior

Running `projectile-find-file` with older (< v8.3.0) versions of `fd` installed prompts the user with a list of project files.

## Actual behavior

Running `projectile-find-file` with older version of `fd` shows the user an error.

![image](https://user-images.githubusercontent.com/2058614/181598425-aba18e68-75a7-43f1-998b-e899050e5dc8.png)

## Steps to reproduce the problem

1. Install a version of `fd` <v8.3.0 (e.g. use Fedora 37 or older)
2. Run `projectile-find-file` in Emacs with `projectile` installed.

### Projectile version information

```
Projectile 20220719.446
```
(Melpa)

### Emacs version

GNU Emacs 27.2

### Operating system

Fedora Silverblue 37; inside Fedora 36 toolbox.

## Related
- [Source](https://github.com/bbatsov/projectile/blob/dc6e7ff658789e4c6cf7c32a4b4fdf22ef9bc5a3/projectile.el#L718-L722)
- [`fd` PR introducing `--strip-cwd-prefix`](https://github.com/sharkdp/fd/pull/861)
- [Issue requesting `--strip-cwd-prefix`](https://github.com/bbatsov/projectile/issues/1749)
- [Doom bug report](https://github.com/doomemacs/doomemacs/issues/6600)
- [Doom override PR](https://github.com/doomemacs/doomemacs/pull/6597)

J3RN:
A workaround is to add the following line to your configuration:
```
(setq projectile-generic-command "fd . -0 --type f --color=never")

**Comment by krydos** [comment]
It looks like `--strip-cwd-prefix` is a little more complicated option than initially expected... Looking at `fd` repo I can see some discussion going on about it and how it's conflicting with some other options (and even does nothing in some cases). 

Would it make more sense to, instead of using `--strip-cwd-prefix`, use something similar to what we do for `find` command (pipe it to `cut -c3-`)?

It looks like if we don't pass a path (last argument of fd) it will not add that leading cwd prefix which will break the output if we pipe it to `cut -c3c` BUT if we pass a path as `.` then it will add it as leading prefix hence `cut` would work. 

```
fd . --type f --color=never . | cut -c3-
```
Should do the trick. It should work for all fd versions. 

[...]

PR #1792, opened 2022-08-15
Make any version of `fd` work

krydos:
# HOLD ON THIS PLEASE (seems like there is an issue)

Update `fd` in `projectile-generic-command` to NOT use `--strip-cwd-prefix` option because this option appeared in `fd@8.3.0`
which makes projectile to throw an error.  See - https://github.com/bbatsov/projectile/issues/1788

As you can see I also removed `-0` option which was a replacement of `tr '\\n' '\\0'`. We need to remove it in this case because `cut` needs those `\n`s

Those `cut` and `tr` pipes are repeating among all 3 commands which doesn't look great but I decided not to refactor them into a postfix variable. Who knows what other commands will be there and what sort of pipes they will require. 

Tested on the latest `fd@8.4.0` and on `fd@8.2.1`

Feel free to let me know to change/improve anything. 
Thanks. 

P.S. 
I wish I can write some tests for this but I have never used `eldev`. Also feel free to force me to do so

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -716,10 +716,10 @@ Set to nil to disable listing submodules contents."
   (cond
    ;; we prefer fd over find
    ((executable-find "fd")
-    "fd . -0 --type f --color=never --strip-cwd-prefix")
+    "fd . --type f --color=never . | cut -c3- | tr '\\n' '\\0'")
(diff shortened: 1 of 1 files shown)


krydos:
I messed up with it. I'll reopen when I'm 100% sure it works. 

Sorry about the noise.

PR #1797, opened 2022-10-02
Add Local Overrides for more Project Attributes

LaurenceWarne:
Hi!  This PR adds variables intended to be used to be used as local overrides (in the same vein as https://docs.projectile.mx/projectile/projects.html#configure-a-projects-lifecycle-commands) for the project type attributes test-prefix, test-suffix, related-files-fn, src-dir and test-dir.  The intention is that you would no longer have to register a custom project type and set `projectile-project-type` in your `.dir-locals.el` just because you want to override say the test suffix for one particular project.

Thanks!

-----------------

Thanks!


bbatsov:
Great improvement! Thanks!

PR #1801, opened 2022-10-06
Allow specifying project name in the dirconfig file

Thuna-Cing:
This allows specifying the project name in the .projectile file.

It is backwards compatible as the older versions will parse the name as a comment.

All tests except one that also fails on origin are passing so I marked them as such.
There should be a new test added to actually check that the project name is being 
parsed properly but it can wait until the rest of the change is finalized, similarly for the
readme and the changelog.

----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1308,8 +1308,18 @@ explicitly."
 (defun projectile-default-project-name (project-root)
   "Default function used to create the project name.
 The project name is based on the value of PROJECT-ROOT."
-  (or (let ((default-directory project-root))
-        (nth 3 (projectile-parse-dirconfig-file)))
+  (or (when-let*
+          ((comment-prefix projectile-dirconfig-comment-prefix)
+           (name-prefix projectile-dirconfig-name-prefix)
+           (default-directory project-root)
+           (dirconfig (projectile-dirconfig-file)))
+        (when (projectile-file-exists-p dirconfig)
+          (with-temp-buffer
+            (insert-file-contents dirconfig)
+            (when (re-search-forward
+                   (format "^%c%c\\(.*\\)$" comment-prefix name-prefix)
+                   nil t)
+              (match-string 1)))))
       (file-name-nondirectory (directory-file-name project-root))))
 
 (defun projectile-project-name (&optional project)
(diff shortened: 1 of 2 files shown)


bbatsov:
Isn't it much easier to set a custom project name using `.dir-locals.el`?

Thuna-Cing:
> Isn't it much easier to set a custom project name using `.dir-locals.el`?

`.dir-locals.el` doesn't work reliably in non-file buffers.

bbatsov:
But at the cost of adding more complexity to an already complex thing. That's why I'm a bit on the fence about this. I wonder if it won't be simpler to introduce a new file like `.projectile-info` or something where additional project data can be stored aside from ignore/include stuff.

Thuna-Cing:
Does projectile use (or intend to use) any information other than the project name?  If it will also contain the projectile configuration in `.dir-locals.el`, then it might be worth it to create a different `.projectile-info` file but it seems overkill otherwise

bbatsov:
At some point I was thinking of making some more structured configuration file to replace `.projectile`, but I put this idea on the backburner as with `.dir-locals.el` we've managed to address 90+% of what was needed essentially for free. And it solves the most complex problem - keep in sync the configuration file and the runtime configuration.

projectile.el
```diff
@@ -1299,7 +1308,9 @@ explicitly."
 (defun projectile-default-project-name (project-root)
   "Default function used to create the project name.
 The project name is based on the value of PROJECT-ROOT."
-  (file-name-nondirectory (directory-file-name project-root)))
+  (or (let ((default-directory project-root))
+        (nth 3 (projectile-parse-dirconfig-file)))
```
bbatsov, line 1312: As the project name is constantly recalculated in the modeline parsing the config file all the time can make things extremely slow. That's one more reason I'm wary about the proposed change.

Thuna-Cing:
Hmm, I didn't realize just how often `projectile-project-name` was called.  Then, instead of `projectile-parse-dirconfig-file`, a function that only checks for the name should be fast enough.  I just made a new commit as a proof of concept, maybe that will be better.

bbatsov:
The 3.0 dirconfig parser rewrite made this patch obsolete, and `projectile-project-name` is marked `:safe` these days, so a committed `.dir-locals.el` gives you a shareable project name without a second mechanism. Thanks for the patch and the discussion regardless.

PR #1805, opened 2022-10-14
[Fix #1708] Search hidden files with projectile-ripgrep

### LINKED ISSUE #1708: `projectile-ripgrep` do not search into hidden files
The default behavior of `ripgrep` is to not search into hidden files (files beginning with a dot) by default. As a result, using `projectile-ripgrep` do not search into hidden files. This is from my point of view an unexpected behavior in the context of a project search.

## Expected behavior

`projectile-ripgrep` should search into hidden files by default.

## Actual behavior

`projectile-ripgrep` ignore hidden files.

Search with `projectile-ripgrep` a patterm you have in a hidden file of your project. You will not found it in the results.

Projectile 2.4.0

### Emacs version

Emacs 27.1 (Doom Emacs)

### System info

GNU/Linux, Ubuntu 20.04
ripgrep 11.0.2

NickSavage:
This closes #1708. This is a small behaviour change to make the behaviour consistent across different methods of searching files. The issue raised was on how `projectile-ripgrep` did not search through hidden files, while `projectile-grep` does. This also still respects the `.gitignore` and the various excluded directories, so the only change is that it now includes non-specifically excluded hidden files.

Interestingly enough, the behaviour behind using the different ripgrep libraries `rg` and `ripgrep` were actually different here, where `rg` would include hidden files while `ripgrep` would exclude them.

----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 * [#1692](https://github.com/bbatsov/projectile/issues/1692): Enable minibuffer completions when reading shell-commands.
 * Change the Grails project marker to `application.yml`.
 * [#1789](https://github.com/bbatsov/projectile/pull/1789): Progress reporter for recursive progress discovery
+* [#1708](https://github.com/bbatsov/projectile/issues/1708): `projectile-ripgrep` now consistently searches hidden files.
 
 ## 2.5.0 (2021-08-10)
 
--- a/projectile.el
+++ b/projectile.el
@@ -4168,7 +4168,7 @@ installed to work."
                            (projectile-acquire-root)
                            (if arg
                                args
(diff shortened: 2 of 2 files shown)


bbatsov:
> Interestingly enough, the behaviour behind using the different ripgrep libraries rg and ripgrep were actually different here, where rg would include hidden files while ripgrep would exclude them.

Pretty weird indeed. Thanks for the fix!

efroemling:
Just chiming in here quickly: I'm currently finding that projectile with rg *isn't* searching for hidden files for me, which seems to be counter to the findings here. Perhaps something changed since then?
If I do a `(setq rg-command-line-flags '("--hidden"))` then I do get them.

PR #1806, opened 2022-10-17
Projectile-project-type could return the wrong project type when called with argument

fanshi1028:
## Expected behavior

(projectile-project-type DIR) should return the right project type.

## Actual behavior

(projectile-project-type DIR) sometimes return the wrong project type

### Why 

The problem is, some logic branches inside `projectile-project-type` involve calling 
`(projectile-detect-project-type)` which use the `default-directory` instead of the DIR we supply.

### Potential Fix

1. Amend `projectile-detect-project-type` to take optional argument DIR(just like `projectile-project-type`).

2. Pass along the input from the projectile-project-type to `projectile-detect-project-type`

Let DIR1 and DIR2 be 2 directories from projects of different types

Call (projectile-project-type DIR2) from a DIR1, you will get DIR1's project type instead of DIR2's(given that DIR2's project type is not cached)

```
Projectile 2.6.0-snapshot
```

### Emacs version

GNU Emacs 29.0.50 (build 1, x86_64-apple-darwin19.6.0, NS appkit-1894.60 Version 10.15.6 (Build 19G2021))

### Operating system

macOS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 
 ### Bugs fixed
 
+* [#1806](https://github.com/bbatsov/projectile/pull/1806): Fix `projectile-project-type` to return the correct project type even when we pass it the DIR arg. As a result of the fix, `projectile-expand-root`, `projectile-detect-project-type`, `projectile-verify-files` and `projectile-verify-file` now also takes an &optional DIR arg to specify the directory it is acting on. 
 * [#1781](https://github.com/bbatsov/projectile/pull/1781): Fix `rails-rspec` and `rails-test` to use `app` instead of `lib` as `src-dir`.
 * [#1762](https://github.com/bbatsov/projectile/pull/1762): Fix `projectile-globally-ignored-directories` unescaped regex.
 * [#1713](https://github.com/bbatsov/projectile/issues/1731): Fix `projectile-discover-projects-in-directory` reordering known projects.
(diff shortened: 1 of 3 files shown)


fanshi1028:
I could have just bind the default-directory locally in the middle of the stack to fix the issue. But currently I chose to go deeper down the function stack to make those functions taking the same optional dir as argument, kind of give them a more unified interface. Which way do you prefer?

bbatsov:
Your suggestions seem reasonable, but I'm not quite sure what exactly is the problem that you've encountered (and I don't see any new tests that's illustrate it). I got that somewhere the project detection is failing, because it's not acting on the right folder, but you didn't say where exactly.

fanshi1028:
> Your suggestions seem reasonable, but I'm not quite sure what exactly is the problem that you've encountered (and I don't see any new tests that's illustrate it). I got that somewhere the project detection is failing, because it's not acting on the right folder, but you didn't say where exactly.

Here is a branch with just the commit for the tests for the issues.
[master + tests](https://github.com/fanshi1028/projectile/tree/Fix_%231806_new_tests_on_master)

Briefly, what I did was calling `(projectile-project-type DIR1)` while I was at `default-directory` of `DIR2`. In this case I thought I would get the `projectile-project-type` of `DIR1` instead of `DIR2`, but it failed my expectation.

bbatsov:
Got it. Thanks for working on this! 🙇‍♂️

Issue #1729, opened 2021-12-01
`projectile-root-top-down` searches bottom-up


philiprodrigues:
## Expected behavior

With the following directory structure, I expect `(projectile-root-top-down "~/projectile-find/subdir")` to return `~/projectile-find`, based on the existence of `~/projectile-find/Makefile`

```
~/projectile-find/
├── Makefile
├── .projectile
└── subdir
    ├── Makefile
    └── .projectile
```
(The .projectile files are for the benefit of comparing to `projectile-root-bottom-up`, below)

## Actual behavior

`projectile-root-top-down` appears to search in the same order as `projectile-root-bottom-up`.

## Steps to reproduce the problem

In the scratch buffer, with the directory tree described above:

```
(projectile-root-top-down "~/projectile-find/subdir")                                                                                                                                                                                              
"/home/me/projectile-find/subdir/"

(projectile-root-bottom-up "~/projectile-find/subdir")                                                                                                                                                                                             
"/home/me/projectile-find/subdir/"                                                                                                                                                                                                           
```

I think the reason is that `projectile-root-top-down` calls `projectile-locate-dominating-file` which does a bottom-up search.

I got the behaviour I wanted locally with:

```lisp
(defun my/projectile-locate-dominating-file-top-down (file name)
  "Look up the directory hierarchy from FILE for a directory containing NAME.
Return the highest parent directory containing a file NAME,
and return the directory.  Return nil if not found.
Instead of a string, NAME can also be a predicate taking one argument
\(a directory) and returning a non-nil value if that directory is the one for
which we're looking."
  (setq file (abbreviate-file-name file))
  (let ((root nil)
        try)
    (while (not (or (equal file "/")
                    (null file)
                    (string-match locate-dominating-stop-dir-regexp file)))
      (setq try (if (stringp name)
                    (projectile-file-exists-p (expand-file-name name file))
                  (funcall name file)))
      (if try (setq root file))
      (setq file (file-name-directory
                  (directory-file-name file)))
      (setq try nil)
      )
    (and root (expand-file-name (file-name-as-directory root)))))

(defun my/projectile-root-top-down (dir &optional list)
  "Identify a project root in DIR by top-down search for files in LIST.
If LIST is nil, use `projectile-project-root-files' instead.
Return the first (topmost) matched directory or nil if not found."
  (my/projectile-locate-dominating-file-top-down
   dir
   (lambda (dir)
[...]

PR #1810, opened 2022-11-05
Find project files bottom-up

### LINKED ISSUE #1796: `projectile-root-bottom-up` doesn’t always find bottom-most file
## Expected behavior

Given a directory structure like
```
~/projectile-example
├── .git
└── subdir
    └── .pijul
```
I’d expect `(projectile-root-bottom-up "~/projectile-example/subdir")` to return `"~/projectile-example/subdir/"`.

## Actual behavior

It returns `"~/projectile-example/"`.

I think that’s covered above. To take out some environmental variability, we can be explicit about the files we’re searching for: `(projectile-root-bottom-up "~/projectile-example/subdir" '(".git" ".pijul"))`.

However, if we reverse the order, `(projectile-root-bottom-up "~/projectile-example/subdir" '(".pijul" ".git"))`, it does the right thing. But if the directory structure is instead 
```
~/projectile-example
├── .pijul
└── subdir
    └── .git
```
the reversed one now returns `~/projectile-example/` and the original returns the correct result.

The problem seems to be that `projectile-root-bottom-up` is actually returning the bottom-most directory _for the first name in the list that it finds at all_. Rather than using `cl-some` here, `projectile-root-bottom-up` should explore the full list and return the longest result.

Projectile 20220430.800

### Emacs version

GNU Emacs 28.1 (build 1, aarch64-apple-darwin21.6.0, NS appkit-2113.60 Version 12.6 (Build 21G115))

### Operating system

macOS 12.6 (21G115)

sellout:
Previously, the directory it found would depend on the order of
`projectile-project-root-files-bottom-up` (or the provided list).

given a directory structure
```
foo
├── .a
└── bar
    └── .a
    └── baz
        └── .b
```
`(projectile-root-bottom-up ".../foo/bar/baz/" '(".a" ".b"))` would return
`".../foo/bar/"`, because the current definition looks for any `".a"` before it
looks for ".b".

With this change, it traverses the directory structure once, looking for any
element of the list before moving up a level. So now the same call returns
`".../foo/bar/baz/"`, which matches the documented behavior of returning the
bottommost matched directory.

Fixes #1796

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -1206,10 +1206,9 @@ Return the first (bottommost) matched directory or nil if not found."
   (projectile-locate-dominating-file
    dir
    (lambda (directory)
-     (when (file-readable-p directory)
-       (let ((files (mapcar (lambda (file) (expand-file-name file directory))
-                            (or list projectile-project-root-files-bottom-up))))
-         (cl-some (lambda (file) (and file (file-readable-p file))) files))))))
+     (let ((files (mapcar (lambda (file) (expand-file-name file directory))
+                          (or list projectile-project-root-files-bottom-up))))
+       (cl-some (lambda (file) (and file (file-exists-p file))) files)))))
 
 (defun projectile-root-top-down-recurring (dir &optional list)
   "Identify a project root in DIR by recurring top-down search for files in LIST.
--- a/projectile.el
+++ b/projectile.el
@@ -1203,8 +1203,13 @@ Return the first (topmost) matched directory or nil if not found."
   "Identify a project root in DIR by bottom-up search for files in LIST.
 If LIST is nil, use `projectile-project-root-files-bottom-up' instead.
 Return the first (bottommost) matched directory or nil if not found."
-  (cl-some (lambda (name) (projectile-locate-dominating-file dir name))
-           (or list projectile-project-root-files-bottom-up)))
+  (projectile-locate-dominating-file
+   dir
+   (lambda (directory)
+     (when (file-readable-p directory)
+       (let ((files (mapcar (lambda (file) (expand-file-name file directory))
+                            (or list projectile-project-root-files-bottom-up))))
+         (cl-some (lambda (file) (and file (file-readable-p file))) files))))))
 
 (defun projectile-root-top-down-recurring (dir &optional list)
   "Identify a project root in DIR by recurring top-down search for files in LIST.
--- a/test/projectile-test.el
+++ b/test/projectile-test.el
@@ -63,7 +63,10 @@ You'd normally combine this with `projectile-test-with-sandbox'."
      ,@(mapcar (lambda (file)
                  (if (string-suffix-p "/" file)
                      `(make-directory ,file t)
-                   `(with-temp-file ,file)))
+                   `(let ((dir (file-name-directory ,file)))
+                      (when dir
+                        (make-directory dir t))
+                      (with-temp-file ,file))))
                files)
      ,@body))
 
@@ -615,25 +618,34 @@ Just delegates OPERATION and ARGS for all operations except for`shell-command`'.
   (it "identifies the root directory of a project by bottom-up search"
     (projectile-test-with-sandbox
      (projectile-test-with-files
-      ("projectA/.svn/"
+      ("projectA/.git/"
+       "projectA/.projectile"
+       "projectA/.svn/"
        "projectA/src/.svn/"
-       "projectA/src/html/.svn/"
-       "projectA/.git/"
-       "projectA/src/html/"
-       "projectA/src/framework/lib/"
        "projectA/src/framework/framework.conf"
-       "projectA/src/html/index.html"
-       "projectA/.projectile")
+       "projectA/src/framework/lib/"
+       "projectA/src/html/"
+       "projectA/src/html/.svn/"
+       "projectA/src/html/index.html")
       (expect (projectile-root-bottom-up "projectA/src/framework/lib" '(".git" ".svn"))
+              :to-equal
+              (expand-file-name "projectA/src/"))
+      (expect (projectile-root-bottom-up "projectA/src/framework/lib" '(".git"))
               :to-equal
               (expand-file-name "projectA/"))
       (expect (projectile-root-bottom-up "projectA/src/html" '(".git" ".svn"))
               :to-equal
-              (expand-file-name "projectA/"))
+              (expand-file-name "projectA/src/html/"))
       (expect (projectile-root-bottom-up "projectA/src/html" '(".svn" ".git"))
               :to-equal
               (expand-file-name "projectA/src/html/"))
       (expect (projectile-root-bottom-up "projectA/src/html" '(".projectile" "index.html"))
+              :to-equal
+              (expand-file-name "projectA/src/html/"))
+      (expect (projectile-root-bottom-up "projectA/src/html" '("index.html" ".projectile"))
+              :to-equal
+              (expand-file-name "projectA/src/html/"))
+      (expect (projectile-root-bottom-up "projectA/src/html" '(".projectile"))
               :to-equal
               (expand-file-name "projectA/"))))))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Bug fixed
 
+* [#1796](https://github.com/bbatsov/projectile/issues/1796): Fix `projectile-root-bottom-up` doesn’t always find bottom-most file
 * [#1799](https://github.com/bbatsov/projectile/pull/1799): Fix `projectile-open-projects` lists projects for which all buffers are closed.
 * [#1806](https://github.com/bbatsov/projectile/pull/1806): Fix `projectile-project-type` to return the correct project type even when we pass it the DIR arg. As a result of the fix,
 `projectile-expand-root`, `projectile-detect-project-type`, `projectile-verify-files` , `projectile-verify-file` `projectile-verify-file-wildcard`, `projectile-cabal-project-p`,


projectile.el
```diff
@@ -1203,8 +1203,13 @@ Return the first (topmost) matched directory or nil if not found."
   "Identify a project root in DIR by bottom-up search for files in LIST.
 If LIST is nil, use `projectile-project-root-files-bottom-up' instead.
 Return the first (bottommost) matched directory or nil if not found."
-  (cl-some (lambda (name) (projectile-locate-dominating-file dir name))
-           (or list projectile-project-root-files-bottom-up)))
+  (projectile-locate-dominating-file
+   dir
+   (lambda (directory)
+     (when (file-readable-p directory)
```
bbatsov, line 1209: Do we really need those permissions checks? I get their purpose, but we generally assume that most of the time you'd be dealing with something readable.

sellout: I’m happy to get rid of this one. The one below here should probably be `file-exists-p` anyway. We don’t really care if, say, `.git` is readable, if it exists, it marks a project.

sellout:
The CI failures seem unrelated to the change.

Also, there are some checkdoc failures in the code base, but none as a result of this change.

bbatsov:
I'll handle the unrelated test failures myself. Thanks for tackling this issue! 🙇‍♂️

thisch:
Hi this change unfortunately broke support for detecting my project root.

I have a main project `/main` which is just a folder of git repos. I've put a `.projectile` file in this folder to declare it as my desired project root. I've compared the two versions of the projectile-root-bottom-up function

```
(projectile-root-bottom-up-old "/main/sub/x")
"/main/"

(projectile-root-bottom-up-new "/main/sub/x")
"/main/sub/"
```

So the old version is working as intended and the new version isn't.

bbatsov:
@thisch The new behavior is the correct one, as it was supposed to find the closest root to the current file. I think that's also related with https://github.com/bbatsov/projectile/issues/1729 (the other function should find the top-most project, which is what you're after). You can adjust `projectile-root-functions` accordingly, as the order of the functions there matters for the project discovery. I'll have to rename the other function, as it's super confusing right now.

thisch:
Thx for this info @bbatsov.  Related to that would it be possible to provide two variants of `projectile-find-file` one that takes the top most project root and the other that takes the bottom most project root or should I write my own and `let`-bind `projectile-root-functions`?

bbatsov:
See https://docs.projectile.mx/projectile/projects.html#customizing-project-detection for a short-term fix. 

> Related to that would it be possible to provide two variants of projectile-find-file one that takes the top most project root and the other that takes the bottom most project root

This can't be in `projectile-find-file`, but rather we need two discovery functions that people can opt to prioritize for their projects. I was under the impression that our existing functions were doing exactly this, but alas they were not. I think we need to rename and fix the other function mentioned in #1729 to behave as you want.

bbatsov:
(a top-down search from projects doesn't make sense anyways, as you can have many projects at the same level, so I assume some just misnamed this function and it actually was supposed to find the top-most project, which is what you want)

sellout:
I think the “top-down” one (which I agree is confusingly named) is meant for resolving roots for projects which have the same “project” file in each subdirectory, e.g., `Makefile`.

I feel like the “bottom-up” ones are generally more reliable (since they track repos), and the “top-down” ones can make mistakes more easily. But … perhaps `.projectile` should be given first priority, since it’s a special signal to Projectile. It basically behaved that way before my change, since it was the first file checked in the “bottom-up” list, but it’s now lost that status.

So, remove `".projectile"` from `projectile-project-root-files-bottom-up, then search for `.projectile` before other bottom-up roots, like:

```elisp
(defcustom projectile-project-root-functions
  '(projectile-root-local
    (lambda (dir) (projectile-root-bottom-up dir '(".projectile")))
    projectile-root-bottom-up
    projectile-root-top-down
    projectile-root-top-down-recurring)
  "A list of functions for finding project root folders.
The functions will be ran until one of them returns a project folder.
Reordering the default functions will alter the project discovery
algorithm."
  :group 'projectile
  :type '(repeat function))
```

I’d be happy to submit that change if it looks like the right thing. (And @thisch, adding that lambda to the list should also work as a temporary fix for you.)

sellout:
tl;dr – I think `.projectile` is still a “bottom-up” file, as you want the deepest one if they are nested, but it should override the “natural” project roots when present in the tree.

bbatsov:
> I think the “top-down” one (which I agree is confusingly named) is meant for resolving roots for projects which have the same “project” file in each subdirectory, e.g., Makefile.

I think `projectile-top-down-recurring` is meant to do check for those. That being said - so much time has passed since those were introduced that even I don't remember how they were supposed to work without looking at the code. :D 

> I feel like the “bottom-up” ones are generally more reliable (since they track repos), and the “top-down” ones can make mistakes more easily. But … perhaps .projectile should be given first priority, since it’s a special signal to Projectile. It basically behaved that way before my change, since it was the first file checked in the “bottom-up” list, but it’s now lost that status.

Yeah, I agree that `.projectile` should always have special status, as it's supposed to override whatever detection is going to happen without it - perhaps its own root look function that's before the others in the list, so it can be ignored by everything else. 

> So, remove ".projectile" from projectile-project-root-files-bottom-up, then search for .projectile before other bottom-up roots, like:

I suggested above before seeing your suggestion, so we're clearly on the same page. :D I'll just suggest not using `lambda` for this. A PR would be welcome!

test/projectile-test.el
```diff
@@ -615,25 +618,34 @@ Just delegates OPERATION and ARGS for all operations except for`shell-command`'.
...
       (expect (projectile-root-bottom-up "projectA/src/framework/lib" '(".git" ".svn"))
+              :to-equal
+              (expand-file-name "projectA/src/"))
+      (expect (projectile-root-bottom-up "projectA/src/framework/lib" '(".git"))
               :to-equal
               (expand-file-name "projectA/"))
       (expect (projectile-root-bottom-up "projectA/src/html" '(".git" ".svn"))
               :to-equal
-              (expand-file-name "projectA/"))
+              (expand-file-name "projectA/src/html/"))
```
mniestroj, line 638: This is wrong here, as it changes expected results. `"projectA/src/html/"` should be returned (and is returned) by `projectile-root-top-down`.

test/projectile-test.el
```diff
@@ -615,25 +618,34 @@ Just delegates OPERATION and ARGS for all operations except for`shell-command`'.
...
        "projectA/src/framework/framework.conf"
-       "projectA/src/html/index.html"
-       "projectA/.projectile")
+       "projectA/src/framework/lib/"
+       "projectA/src/html/"
+       "projectA/src/html/.svn/"
+       "projectA/src/html/index.html")
       (expect (projectile-root-bottom-up "projectA/src/framework/lib" '(".git" ".svn"))
+              :to-equal
+              (expand-file-name "projectA/src/"))
```
mniestroj, line 632: If you copy-paste the same test to `projectile-root-top-down`, then it will pass as well. This is wrong, as both functions should traverse file system in the opposite direction.

The expected result here should be `"projectA/"`.

mniestroj (commented):
@bbatsov @sellout 

This PR is changing behavior of `projectile-root-bottom-up` to traverse file system in the same direction as `projectile-root-top-down`. This is a regression and IMHO should be reverted. Please see my comments inline.

mniestroj:
Unfortunately old version of `projectile-root-bottom-up` is also broken, so revert is not a solution here.

mniestroj:
> I am wondering when it’s the case that you ever want to search starting from the root. I don’t doubt that there are cases where a project root file commonly exists in multiple non-adjacent places in the project tree, but I don’t know of any off the top of my head.

So here is my use case. I develop Zephyr RTOS using a tool called [west](https://github.com/zephyrproject-rtos/west). This tool manages several git repositories, a bit similar to how [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) and [repo](https://gerrit.googlesource.com/git-repo/+/refs/heads/master/README.md) (from Android) work. My directory tree is something like:

```
.
├── app
│   ├── ...
│   ├── .git/
│   └── ...
├── TAGS
├── .west/
└── zephyr
    ├── ...
    ├── .git/
    └── ...
```
In this case `app/` is the application directory, which I develop. `zephyr/` is RTOS code, which contains most of the APIs and subsystems that I use within `app/` implementation. There are also other components in this tree, which I did not mention to simplify things. Overall there are many git repositories, but I want to treat the whole tree which has `.west/` directory as a single project. So what I was using so far was adding `.west` to the `projectile-project-root-files-bottom-up` so it would have following value `(".projectile" ".west" ".project" ".git")`. It worked fine, because `.west` was prioritized over `.git`. However after this PR all the items in the list has the same priority (which is good, of course), so it stopped working.

After thinking about it for longer time, looking at your explanation, at the current implementation, etc. I think I need following priorities:

mniestroj:
> My inclination is to remove `projectile-root-bottom-up`, moving everything from `projectile-project-root-files-bottom-up` into `projectile-project-root-files`. Then rename `projectile-root-top-down` to `projectile-find-root` and `projectile-root-top-down-recurring` to `projectile-find-root-recurring`, updating the documentation to avoid “top”/“bottom” terminology. (With obsolete aliases for the time being.)

This sounds like the right approach to me.

> 
> That change however exposes a second issue – this PR made it so that it would search across the whole list. Which means that if “.git” and “TAGS” are in the same list, a “TAGS” file _not_ at the root of a project would cause it to define the project root, rather than trusting the more reliable “.git”. “.projectile” is already pulled out as a higher-priority root marker, and I think that’s a good model to follow. The root files should be a list of lists – each list having files of the same priority, so like
> 
> ```emacs-lisp
> '((".projectile") ; explicit marker file
>   (".git" "_darcs" ".pijul") ; version control project files
>   ("TAGS")) ; less reliable project root indicators
> ```

This list of lists looks like what we need, yes. But we kind of have this "list of lists" here:
```
(defcustom projectile-project-root-functions
  '(projectile-root-local
    projectile-root-marked
    projectile-root-bottom-up
    projectile-root-top-down
    projectile-root-top-down-recurring)
```

Keep in mind that we need some way to inject files/dirs and lists to this "list of lists", when user just wants to slightly change the behavior/priorities and not entirely replace this "list of lists".

> 
> Beyond that, it might be good to include recurring items in the same list, since I would rather trust a “Makefile” than a “TAGS”, like
> 
> ```emacs-lisp
> '((".projectile")
>   (".git" "_darcs" ".pijul")
>   (:recurring "Makefile") ; keyword at the front changes the behavior
>   ("TAGS"))
> ```

I don't think `Makefile` recurring search is more reliable. There are many `Makefile` projects which will have `root/Makefile` and `root/a/b/Makefile`, but they **won't** have `root/a/Makefile`. In this case recurring search won't find `root/` if it starts somewhere under `root/a/b/*`.

mniestroj:
So I found a solution for my use case:
```
(defun projectile-root-west (dir)
    "Identify a project root in DIR by search for \".west\"."
    (projectile-root-bottom-up dir '(".west")))
  (let ((part (member 'projectile-root-marked projectile-project-root-functions)))
    (setcdr part (append '(projectile-root-west) (cdr part))))
```
this basically results in following `projectile-project-root-functions`:
```
  '(projectile-root-local
    projectile-root-marked
    projectile-root-west
    projectile-root-bottom-up
    projectile-root-top-down
    projectile-root-top-down-recurring)
```

sellout:
Yeah, I think we’re on the same page. The way to achieve this currently is … and you posted the same answer before I finished writing this 😂
```elisp
(defun my-projectile-root-west (dir)
  (projectile-root-bottom-up dir '(".west")))

(setq projectile-project-root-functions
  '(projectile-root-local
    projectile-root-marked
    my-projectile-root-west ; new entry
    projectile-root-bottom-up
    projectile-root-top-down
    projectile-root-top-down-recurring))
```
But yeah, I’d like that list-of-lists to be a bit more obvious – although allowing arbitrary functions for how to search for the elements in each list is better than my idea to use a keyword in the list.

Re: “Makefile”, you’re probably right. I only rely on version control files and “.projectile”. Maybe “Makefile” _is_ the use case for parentmost.

mniestroj:
> But yeah, I’d like that list-of-lists to be a bit more obvious – although allowing arbitrary functions for how to search for the elements in each list is better than my idea to use a keyword in the list.

I agree, it is more flexible to allow functions, which can implement any custom user behavior.

bbatsov:
I'd be fine to change this if one of you is willing to work on a PR in that direction. Historically Projectile had a single project lookup function that was aligned with my own use of Projectile and admittedly the addition of the newer lookup functions wasn't thought out carefully. I didn't think much about how those would fit into the bigger picture as I never needed any of them. 

Just one suggestion from me if we go down the route with a single list of functions - it'd be nice to also have some priorities associated with them, so people won't have to override the entire list all the time if they want some function to take precedence. Anyways, that's not something essential - just a nice to have IMO.

sellout:
I'd be happy to. I'll think on it a bit more a put up a more concrete proposal to make sure it meets everyone's needs.

PR #1812, opened 2022-11-12
Add a projectile-root-marked function for finding roots

sellout:
Prioritizes .projectile above other bottom-up root files.

Also makes the name of .projectile customizable.

Related to #1810.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [#1591](https://github.com/bbatsov/projectile/issues/1591): Add `project.el` integration that will make Projectile the default provider for project lookup.
 * Add new command `projectile-find-references` (bound to `C-c C-p ?` and `C-c C-p s x`).
 * [#1737](https://github.com/bbatsov/projectile/pull/1737): Add helpers for `dir-local-variables` for 3rd party use. Functions `projectile-add-dir-local-variable` and `projectile-delete-dir-local-variable` wrap their built-in counterparts. They always use `.dir-locals.el` from the root of the current Projectile project.
+* Add a new defcustom (`projectile-dirconfig-file`) controlling the name of the file used as Projectile’s root marker and configuration file.
 
 ### Bug fixed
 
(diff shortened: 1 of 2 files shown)


sellout:
Here’s the fix, @thisch.

bbatsov:
The proposed changes look good to me. Thanks!

P.S. This should probably be mentioned in the docs as well, but I'll do this myself.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * [#1591](https://github.com/bbatsov/projectile/issues/1591): Add `project.el` integration that will make Projectile the default provider for project lookup.
 * Add new command `projectile-find-references` (bound to `C-c C-p ?` and `C-c C-p s x`).
 * [#1737](https://github.com/bbatsov/projectile/pull/1737): Add helpers for `dir-local-variables` for 3rd party use. Functions `projectile-add-dir-local-variable` and `projectile-delete-dir-local-variable` wrap their built-in counterparts. They always use `.dir-locals.el` from the root of the current Projectile project.
+* Add a new defcustom (`projectile-dirconfig-file`) controlling the name of the file used as Projectile’s root marker and configuration file.
```
thisch, line 11: Shouln't the new projectile-root-marked function be mentioned in the changelog as well?

thisch (commented):
Thx a lot for the new function!

CHANGELOG.md
bbatsov: Yeah, it should. Those things should also be covered somewhere in the docs, so it's easier to discover but I'll do this myself when I can.

bbatsov:
@thisch See https://github.com/bbatsov/projectile/commit/7d414ea3421689a766383b03e566c5399419c566

This part of the docs definitely needs love, but at least it reminded me why we have `projectile-root-top-down`:

> * `projectile-root-top-down` is similar, but it will return the top-most (farthest from the current directory) match. It's configurable via `projectile-project-root-files` and all project manifest markers like `pom.xml`, `Gemfile`, `project.clj`, etc go there.

I guess it's for those situations where it's common to have nested projects, although as the markers are essentially the same as for the other function there's little point is having them together in the `projectile-root-functions` hook. We'll have to think about this a bit.

PR #1813, opened 2022-11-12
Allow multiple project-files per project type, allow wildcards in the project-file specs

toshokan:
Some project types can have different project files name patterns (for example, dotnet projects can have `csproj` or `fsproj` depending on whether they're written in C# or F#)

Related to those two examples, sometimes the project-file does not have a constant name. A C# project named `Foo` will have a project-file named `Foo.csproj`. 

This PR allows project type registrations to specify wildcards for their filenames (like `"*.csproj"`) and multiple project-file specs (like `'("*.csproj" "*.fsproj")`).

I've added some tests to make sure this works and keeps working as expected, and adjusted the type registrations for dotnet projects which benefit directly from this feature.

-----------------

Thanks!

### DIFF
--- a/doc/modules/ROOT/pages/projects.adoc
+++ b/doc/modules/ROOT/pages/projects.adoc
@@ -162,14 +162,27 @@ initialization code
 What this does is:
 
 . add your own type of project, in this case `npm` package.
-. add a list of files and/or folders in a root of the project that helps to identify the type, in this case it is only `package.json`.
-. add _project-file_, which is typically the primary project configuration file. In this case that's `package.json`.
+. add a list of files and/or folders in a root of the project that helps to identify the type, in this case it is only `package.json`. This can also be a function which takes a project root as argument and verifies whether that directory has the correct project structure for the type.
+. add _project-file_, which is typically the primary project configuration file. In this case that's `package.json`. The value can contain wildcards and/or be a list containing multiple project files to look for.
 . add _compile-command_, in this case it is `npm install`.
 . add _test-command_, in this case it is `npm test`.
 . add _run-command_, in this case it is `npm start`.
 . add test files suffix for toggling between implementation/test files, in this case it is `.spec`, so the implementation/test file pair could be `service.js`/`service.spec.js` for example.
 
 Let's see a couple of more complex examples.
+
+[source,elisp]
+----
+;; .NET C# or F# projects
+(projectile-register-project-type 'dotnet #'projectile-dotnet-project-p
+                                  :project-file '("?*.csproj" "?*.fsproj")
+                                  :compile "dotnet build"
+                                  :run "dotnet run"
+                                  :test "dotnet test")
+----
+
+This example uses _projectile-dotnet-project-p_ to validate the project's structure.
+Since C# and F# project files have names containing the name of the project, it uses a list of wildcards to specify the different valid _project-file_ name patterns.
 
 [source,elisp]
 ----
--- a/projectile.el
+++ b/projectile.el
@@ -1181,7 +1181,7 @@ which we're looking."
                     (null file)
                     (string-match locate-dominating-stop-dir-regexp file)))
       (setq try (if (stringp name)
-                    (projectile-file-exists-p (expand-file-name name file))
+                    (projectile-file-exists-p (projectile-expand-file-name-wildcard name file))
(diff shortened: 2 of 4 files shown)


bbatsov:
Looks like a good change, hopefully it doesn't affect negatively the lookup performance. Please, document the new functionality in the docs. 

If I recall correctly there are a few other project types that would benefit from something like this (e.g. Go projects).

projectile.el
```diff
@@ -2963,6 +2967,15 @@ it acts on the current project."
   (or (projectile-verify-file "Eldev" dir)
       (projectile-verify-file "Eldev-local" dir)))
 
+(defun projectile-expand-file-name-wildcard (name-pattern dir)
+  "Expand the potentially wildcard-containing NAME-PATTERN in
+DIR. If there are results expanding a wildcard, get the first
+result, otherwise expand NAME-PATTERN in DIR ignoring wildcards."
+  (let ((expanded (expand-file-name name-pattern dir)))
```
bbatsov, line 2974: Basically, that's the part of the code I'm concerned about from a perspective standpoint, as we are doing 2 file name expansions for everything. Not sure that's really a problem, but I expect that `file-expand-wildcards` is much slower than a regular `expand-file-name`.

bbatsov:
This also needs a changelog entry.

projectile.el
toshokan: Would you prefer if we checked if `NAME-PATTERN` contained `*`, `?`, `[` or `]` before doing the second check?

toshokan:
Would you prefer additional commits for docs, changelog, etc or would you rather I squash them together with the implementation commit?

projectile.el
bbatsov: I was wondering about this. In practice will we ever have a pattern different from `*.ext`? I was thinking this way we can have the fastest check - that of the first character. I'm not sure if the `file-expand-wildcards` doesn't do such checks itself, so it might be useful to check its code.

bbatsov:
> Would you prefer additional commits for docs, changelog, etc or would you rather I squash them together with the implementation commit?

I commit is better for me.

toshokan:
I've modified the docs and added a changelog entry

projectile.el
toshokan: I looked at the implementation of `file-expand-wildcards`. It looks like the way it works is by transforming the wildcard to a regex, and then calling a C function which gets a directory listing and filters for entries matching the regex.

If the pattern doesn't have any wildcard characters, the regex construction is trivialized, but it still does a directory listing and filter, just with a simpler filter.

bbatsov: Well, then I guess it'd be prudent for us to restrict the wildcards we care about only to the bare mininum - `project-name.someext` and filter which something like `string-prefix-p`.

doc/modules/ROOT/pages/projects.adoc
```diff
@@ -163,7 +163,7 @@ What this does is:
 
 . add your own type of project, in this case `npm` package.
 . add a list of files and/or folders in a root of the project that helps to identify the type, in this case it is only `package.json`.
-. add _project-file_, which is typically the primary project configuration file. In this case that's `package.json`.
+. add _project-file_, which is typically the primary project configuration file. In this case that's `package.json`. The value can contain wildcards and/or be a list containing multiple project files to look for.
```
bbatsov, line 166: I'd also add the C# example in the same project definitions (there are some about Ruby projects if I recall correctly), so that's even clearer.

projectile.el
toshokan: I thought just prefix matching was a little too restrictive and would be tricky to explain properly in the docs, so I checked for a match to `[[*?]` to decide whether to perform the wildcard expansion.
I also added a test to document this performance concern.

doc/modules/ROOT/pages/projects.adoc
toshokan: I've added the C#/F# registration to the docs in that section and explained its structure as an example. It does use the function form for `MARKER-FILES`, but I just left that opaque.

I also added to the docs that using a function for markers is actually supported, which was implemented a long time ago but didn't seem to be in the docs.

bbatsov:
Thanks!

PR #1824, opened 2023-02-27
Mention how to set default cmd to (projectile-compile-project)

Hi-Angel:
Upon starting to use `projectile-compile-project`, currently the only
way to find out how to set default command is to look at the
code. Let's mention the necessary variable in the function docs instead.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -5172,7 +5172,8 @@ with a prefix ARG."
 
 Normally you'll be prompted for a compilation command, unless
 variable `compilation-read-command'.  You can force the prompt
-with a prefix ARG."
+with a prefix ARG.  Per project default command can be set through
+`projectile-project-compilation-cmd'."
   (interactive "P")
   (let ((command (projectile-compilation-command (projectile-compilation-dir)))
         (command-map (if (projectile--cache-project-commands-p) projectile-compilation-cmd-map)))


projectile.el
```diff
@@ -5172,7 +5172,8 @@ with a prefix ARG."
 
 Normally you'll be prompted for a compilation command, unless
 variable `compilation-read-command'.  You can force the prompt
-with a prefix ARG."
+with a prefix ARG.  Per project default command can be set through
+`projectile-project-compilation-cmd'"
```
bbatsov, line 5176: You forgot the final `.` here.

Hi-Angel: Thx, done!

bbatsov:
> currently the only way to find out how to set default command is to look at the code.

It's already in the docs (https://docs.projectile.mx/projectile/projects.html#configure-a-projects-lifecycle-commands-and-other-attributes). :-)

Hi-Angel:
> > currently the only way to find out how to set default command is to look at the code.
>
> It's also in the docs (https://docs.projectile.mx/projectile/projects.html#configure-a-projects-lifecycle-commands-and-other-attributes). :-)

Well, when I use some function of a Emacs addon that I know little about, the first thing I try is "monkey-typing" — I press <kbd>M-x</kbd> and try to type keywords I'm interested in. In this case I know Emacs has `compile` command, but projectile makes things work per project, so I assumed it has some some similar `projectile.*compile` function. So I typed inside <kbd>M-x</kbd> menu those words *(I have `ido-mode` fuzzy completion in mini-buffer)* — and lo and behold, I found the function. Then I read docs for it, which doesn't mention the essential for this function variable, which is why I thought it might be good idea to have it mentioned :)

bbatsov (approved):

PR #1827, opened 2023-03-05
Make projectile-run-vterm open vterm in other window

sathishmanohar:
opens Vterm in other window instead of opening in the same window while invoking Vterm with `projectile-run-vterm`

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -83,7 +83,7 @@
 (declare-function ggtags-update-tags "ext:ggtags")
 (declare-function ripgrep-regexp "ext:ripgrep")
 (declare-function rg-run "ext:rg")
-(declare-function vterm-other-window "ext:vterm-other-window")
+(declare-function vterm-other-window "ext:vterm")
 (declare-function vterm-send-return "ext:vterm")
 (declare-function vterm-send-string "ext:vterm")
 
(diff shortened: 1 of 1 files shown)


bbatsov:
You'll have to declare this function to avoid the compilation warning. (update the declaration in place for the function you've replaced)

projectile.el
```diff
@@ -83,7 +83,7 @@
 (declare-function ggtags-update-tags "ext:ggtags")
 (declare-function ripgrep-regexp "ext:ripgrep")
 (declare-function rg-run "ext:rg")
-(declare-function vterm "ext:vterm")
+(declare-function vterm-other-window "ext:vterm-other-window")
```
bbatsov, line 86: You should not change the ext part - that's the library name (`vterm.el`).

sathishmanohar: Sorry about that. I haven't written emacs lisp more than a few simple functions for my config.

Thanks a lot for the feedback. Fixed now.

hexedpackets:
Is this configurable? I'd like to restore the old behavior for my setup but I'm struggling to see how to do so.

sathishmanohar:
It is not configurable at the moment. `vterm` function was replaced with `vterm-other-window`

bbatsov:
> Is this configurable? I'd like to restore the old behavior for my setup but I'm struggling to see how to do so.

A configuration option would be trivial to add. Just open an issue and I'll do when I have a bit of time.

PR #1833, opened 2023-03-16
Detect Julia projects

berquist:
Julia projects were getting detected as `dotnet-sln`. This template follows the structure created by [PkgTemplates.jl](https://github.com/JuliaCI/PkgTemplates.jl) which is pretty standard, so I didn't bother to be more specific than `julia`.

I also added a Julia detection test and another that confirms only a `src` dir present matches a .NET solution.

-----------------

~~- [ ] You've updated the readme (if adding/changing user-visible functionality)~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #1840, opened 2023-04-20
Obey `projectile-project-test-dir` (and similarly for src dir) when toggling between impl and test

### LINKED ISSUE #1839: Toggle between test/impl doesn't obey `projectile-project-test-dir`
My project `.dir-locals.el` sets `projectile-project-test-dir` to "tests/" but this is being ignored by `projectile-toggle-between-implementation-and-test`

## Expected behavior

Test files should be found under "tests/"

## Actual behavior

Tests are being created/looked for under "test/" (which is the default for the project type)

Create a .dir-locals.el that sets  `projectile-project-test-dir` to a value different from the default for the project type.
Open a src file, call `projectile-toggle-between-implementation-and-test`.

```
Projectile 2.8.0-snapshot
```

### Emacs version

29.0.90

### Operating system

Linux

Lenbok:
Uses the existing `projectile-test-directory` function when looking up test directory (and right beside these are calls looking up the src-dir attribute that should also be using the existing `projectile-src-directory` function.

Note, that this seems to require the trailing slash on the `projectile-project-test-dir` value. I'm not sure if that is OK. It seems consistent with what is declared in the various project types, but is not mentioned in the variable documentation.

Fixes #1839
-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Changes
 
+* [#1839](https://github.com/bbatsov/projectile/issues/1839): Ensure `projectile-toggle-between-implementation-and-test` also obeys `projectile-project-test-dir` and `projectile-project-src-dir`.
 * [#1285](https://github.com/bbatsov/projectile/pull/1825): By default, use [fd](https://github.com/sharkdp/fd) in Git repositories instead of `git ls-files` when it is installed, in order to solve the problem where deleted files were still shown in `projectile-find-file` until their deletions were staged. The user-facing behavior should be the same, although potentially with different performance characteristics in large Git repositories. The old behavior can be reclaimed by setting `projectile-git-use-fd` to nil.
 * [#1831](https://github.com/bbatsov/projectile/issues/1831): Enable the project.el integration only when `projectile-mode` is active.
 
--- a/projectile.el
+++ b/projectile.el
@@ -3675,8 +3675,8 @@ string) are replaced with the current project type's src-dir property
 
 Nil is returned if either the src-dir or test-dir properties are not strings."
   (let* ((project-type (projectile-project-type))
-         (test-dir (projectile-project-type-attribute project-type 'test-dir))
-         (impl-dir (projectile-project-type-attribute project-type 'src-dir)))
+         (test-dir (projectile-test-directory project-type))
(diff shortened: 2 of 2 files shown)


bbatsov:
Your change seems reasonable, but it has broken some tests. Please, look into them. 

Also mention your fix in the changelog.

Lenbok:
I have updated the changelog. 

I have no idea about how eldev works. I tried installing it and running it from my projectile repo as described, and there's a different test failure to what appears in the github checks:
```
========================================
projectile-add-dir-local-variable Fails when there is no projectile project

Traceback (most recent call last):
  buttercup-fail("%s" "Expected `(projectile-add-dir-local-variable nil 'fooo 1)' to throw a child signal of `error', but instead it evaluated successfully, returning value `t'")
  signal(buttercup-failed "Expected `(projectile-add-dir-local-variable nil 'fooo 1)' to throw a child signal of `error', but instead it evaluated successfully, returning value `t'")
FAILED: Expected `(projectile-add-dir-local-variable nil 'fooo 1)' to throw a child signal of `error', but instead it evaluated successfully, returning value `t'

========================================
```
and the `.dir-locals.el` file in the repository gets modified. The tests also hang waiting for input. If I press enter, they continue to run and I get the two failures that are seen on the github actions.

Regarding the two test failures that seem relevant, I don't understand how my changes would cause them to fail unless those tests are setting `projectile-project-test-dir` and/or `projectile-project-src-dir` somewhere, but if they are I have no fricken idea.

LaurenceWarne:
I think the errors arise from the function `projectile-test-directory` falling back to `projectile-default-test-directory` (which is non-nil by default) when the variable `projectile-test-directory` is `nil` and the project type doesn't set the `test-dir` property.

I think it may be safe to remove these fallbacks from `projectile-test-directory` and `projectile-src-directory`, doing so doesn't appear to fail any tests anyway - though maybe there's a nicer way.

bbatsov:
> I think it may be safe to remove these fallbacks from projectile-test-directory and projectile-src-directory, doing so doesn't appear to fail any tests anyway - though maybe there's a nicer way.

I agree that's a better solution.

Lenbok:
Have made @LaurenceWarne 's suggestion.

PR #1842, opened 2023-05-02
Follow includes in CMake files

kacprzak:
I use conan dependency manager to setup c++ cmake project. It creates/modifies CMakeUserPresets.json by adding includes to generated CMakePresets.json files. The problem is that projectile does not follow those includes when parsing presets files.

Warning: this is my first ever PR in elisp.

-----------------

No, but no cmake related code is testes or I was unable to find any tests.
No, I had one error but I also had it without my changes.
Had many warnings but non related to new code.
Not sure if this change qualifies for last two checkboxes.

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4546,7 +4546,7 @@ Switch to the project specific term buffer if it already exists.
 
 Use a prefix argument ARG to indicate creation of a new process instead."
   (interactive "P")
-  (project--vterm arg))
+  (projectile--vterm arg))
 
 ;;;###autoload
 (defun projectile-run-vterm-other-window (&optional arg)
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -3091,10 +3091,17 @@ it acts on the current project."
...
-    (cl-remove-if
-     (lambda (preset) (equal (gethash "hidden" preset) t))
-     (gethash (projectile--cmake-command-preset-array-id command-type) preset))))
+  (when-let ((filename (projectile-expand-root filename))
+             (preset (projectile--cmake-read-preset filename)))
+    (append
+     (cl-remove-if
+      (lambda (preset) (equal (gethash "hidden" preset) t))
+      (gethash (projectile--cmake-command-preset-array-id command-type) preset))
+     (mapcar
```
bbatsov, line 3100: Perhaps you should make this is a separate private function, so it's clearer what it does exactly.

kacprzak: I kept the old implementation unchanged, just changed the name (added `-shallow` to name). Now `projectile--cmake-command-presets` calls the old one, and contains only new stuff. Should be more clear.

bbatsov:
Sorry for the slow turnaround here. I had totally forgotten about this PR.

Issue #1847, opened 2023-05-29
`projectile-complementary-dir` mangles casing


Lenbok:
I have defined a projectile type for R projects, which typically have their source code under a directory called `R/`, and unit tests in a directory called `tests/testthat/`. 

I have added a project type like this:
```lisp
  (projectile-register-project-type 'my-ess-r '("R")
                                    :compile "make"
                                    :test "make test"
                                    :src-dir "R/"
                                    :test-dir "tests/testthat/"
                                    :test-prefix "test_")
```

I want to use the functionality to toggle between impl and test, but it is broken.

## Expected behavior

Jumping from impl `R/foo.R` should take me to `tests/testthat/test_foo.R`

## Actual behavior

Jumping from impl `R/foo.R` takes me to `TESTS/TESTTHAT/test_foo.R`

## Steps to reproduce the problem

Create file structure:
```
R
R/foo.R
tests/testthat/test_foo.R
```

### Projectile version information

```
Projectile 2.8.0-snapshot
```

### Emacs version

GNU Emacs 29.0.90

### Operating system

Linux Mint.

PR #1848, opened 2023-05-30
Use literal directory name casing during impl / test toggle

Lenbok:
`projectil-complementary-dir` was using `replace-regexp-in-string` allowing case manging during the replacement. That's not what we want here. See #1847

Fixes: #1847 
-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3746,7 +3746,8 @@ Replace STRING in DIR-PATH with REPLACEMENT."
   (let* ((project-root (projectile-project-root))
          (relative-dir (file-name-directory (file-relative-name dir-path project-root))))
     (projectile-expand-root
-     (replace-regexp-in-string string replacement relative-dir))))
+     ;; TODO: Use string-replace once we target emacs 28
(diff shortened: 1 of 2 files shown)


Lenbok:
Pretty sure that one CI failure on snapshot isn't this PR.

projectile.el
```diff
@@ -3746,7 +3746,7 @@ Replace STRING in DIR-PATH with REPLACEMENT."
   (let* ((project-root (projectile-project-root))
          (relative-dir (file-name-directory (file-relative-name dir-path project-root))))
     (projectile-expand-root
-     (replace-regexp-in-string string replacement relative-dir))))
+     (replace-regexp-in-string string replacement relative-dir t))))
```
bbatsov, line 3749: We should also leave some TODO to use `(string-replace FROMSTRING TOSTRING INSTRING)` (introduced in Emacs 28) once we start targeting Emacs 28 in the distant future.

Lenbok: I've updated the PR to include a TODO comment.

PR #1851, opened 2023-06-08
Add ripgrep as a commander method.

ska2342:
Very small PR to add ripgrep as a method to the projectile commander just like Ag is.

No tests added, because I didn't see any suitable tests for commander or Ag.

On the choice of keybinding: I'd prefer `r` but that's already taken and I don't want to introduce a breaking change. The 'p' in 'ripgrep' stands out phonetically so it seemed the best choice to me. Any lower case character would do, I guess.

Changelog not yet adjusted. Would wait for a PR/issue ID to reference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #1865, opened 2023-10-11
[Fix #1853] `projectile-generic-command` should use `projectile-fd-executable` to find the path for fd

### LINKED ISSUE #1853: `projectile-generic-command` will try to use `fd` even if  `projectile-fd-executable` is set to `nil`
`projectile-generic-command` is set to use `fd` or `fdfind` if they are available, without checking that the user didn't explicitly ask to not use them.
https://github.com/bbatsov/projectile/blob/31b87151b1fe43221736ded957a1123a54e32531/projectile.el#L754-L768

I used projectile on a machine with Ubuntu 20.04, it had `fd` 7.4.0, which complained about the `--strip-cwd-prefix` flag.
I tried to work around that by setting `projectile-fd-executable` to `nil`, but projectile kept using `fdfind`.

## Workaround
Explicitly set `projectile-generic-command` to a command that doesn't use `fd` or `fdfind`

## Expected behavior
* Set `projectile-fd-executable` to `nil`
* projectile will not try to use `fd` or `fdfind`

## Actual behavior
* Set `projectile-fd-executable` to `nil`
* projectile will try to use `fd` or `fdfind`

miles170:
Closes #1853 

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 
 * Fix `fd` inserting color control sequences when used over tramp.
 * [#1835](https://github.com/bbatsov/projectile/issues/1835): Reopening existing vterm buffer in other window
+* [#1865](https://github.com/bbatsov/projectile/pull/1865): `projectile-generic-command` should use `projectile-fd-executable` to find the path for fd.
 
 ## 2.7.0 (2022-11-22)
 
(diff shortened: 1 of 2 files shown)


miles170:
I'm not sure why the `CI / test (snapshot) (pull_request)` failed.

bbatsov:
Thanks for tackling this! The CI failure seems unrelated to your changes.

Issue #1868, opened 2023-11-25
Loading projectile.el breaks magit/lsp in emacs 27.1


Lunderberg:
Currently, projectile provides a `(cl-defmethod project-root ...)` to integrate with `project.el`.  For emacs 28.0 and up, this is an additional overload that works alongside the default overloads provided in `project.el` (commit [`5044c190`](https://github.com/emacs-mirror/emacs/commit/5044c19001fe608f2eac621add2e05cbca6c804b)).  For earlier versions of emacs, this is the only overload present.

Many packages check `(fboundp 'project-root)` in order to determine whether they should call `project-root` or `project-roots`.  By defining the single overload for `project-root`, `projectile.el` causes `(fboundp 'project-root)` to return true, and the incorrect function is called.

## Expected behavior

Loading `projectile.el` should have no impact on the usage of either `(magit-project-status)` or `(lsp-find-references)`.

## Actual behavior

Loading `projectile.el` causes later use of `(magit-project-status)` or `(lsp-find-references)` to fail with the error message `'cl-no-applicable-method: No applicable method: project-root, (vc . "~/misc/rust-analyzer/")'`

## Steps to reproduce the problem

1. Load `projectile.el`
2. Attempt to use `(magit-project-status)`.

### Projectile version information

```
Projectile 20231013.1509
```

### Emacs version

Emacs 27.1

### Operating system

Linux Mint 20

PR #1869, opened 2023-11-25
Check for project-root before adding overload

Lunderberg:
Prior to emacs 28, the `project.el` package did not provide `project-root`.  By providing an overload for `project-root`, any compatibility checks of `(fboundp 'project-root)` done by other packages will return true, even for versions that do not have `project-root`.

Fixes https://github.com/bbatsov/projectile/issues/1868.

-----------------

- [NA] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)
- [NA] You've updated the readme (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -6155,8 +6155,11 @@ when opening new files."
 ;; it's safe to require this directly, as it was added in Emacs 25.1
 (require 'project)
 
-(cl-defmethod project-root ((project (head projectile)))
-  (cdr project))
+;; Only define an override for project-root if the method exists.  For versions
+;; before emacs 28, project.el provided project-roots instead of project.root.
+(if (fboundp 'project-root)
+    (cl-defmethod project-root ((project (head projectile)))
+      (cdr project)))
 
 (cl-defmethod project-files ((project (head projectile)) &optional _dirs)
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -6155,8 +6155,11 @@ when opening new files."
 ;; it's safe to require this directly, as it was added in Emacs 25.1
 (require 'project)
 
-(cl-defmethod project-root ((project (head projectile)))
-  (cdr project))
+;; Only define an override for project-root if the method exists.  For versions
+;; before emacs 28, project.el provided project-roots instead of project.root.
+(if (fboundp 'project-root)
```
bbatsov, line 6160: Shouldn't we override the old method in the else branch for completeness?

Lunderberg: Probably, though I haven't directly used the integration, and so I wouldn't have been able to verify that the changes work as intended.  In principle, the overload would just be the overload for `project-root`, but returning a list of length 1 instead of a single element.

```lisp
(cl-defmethod project-roots ((project (head projectile)))
      (list (cdr project))))
```

PR #1874, opened 2024-01-12
If the file already exists, attempt to navigate to it directly instea…

sonictk:
This tiny PR proposes to improve how `compilation-find-file-projectile-find-compilation-buffer` works by directly navigating to the file if already present on disk. On projects with a very large number of directories (i.e. where the projectile cache exceeds 2GB, as is common when working on Unreal projects), this can make the difference between navigating to a file from the compilation buffer instantly VS waiting for potentially upwards of 30s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master (unreleased)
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Fine by me. Just mention this in the changelog, please.

sonictk:
Thanks for taking a look at this. I've updated the `CHANGELOG.md` as requested.

PR #1877, opened 2024-01-21
Add new custom variable 'projectile-cmd-hist-ignoredups'

LaurenceWarne:
Hi!  This PR adds new custom variable `projectile-cmd-hist-ignoredups`, which can be used to tweak how duplicates are dealt with in projectile's command history.  The custom variable is more or less identical in behaviour to `eshell-hist-ignoredups`.

Specifically, the existing default behavior is maintained with the value of `t`, which means consecutive duplicates are ignored.  A value of `'erase` means only the last duplicate is kept, whilst a value of `nil` means all duplicates are kept.

(PR is an iteration on https://github.com/bbatsov/projectile/pull/1774)

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -874,6 +874,17 @@ If the value is nil, there is no limit to the opend buffers count."
   :type 'integer
   :package-version '(projectile . "2.2.0"))
 
+(defcustom projectile-cmd-hist-ignoredups t
+  "Controls when inputs are added to projectile's command history.
+
+A value of t means consecutive duplicates are ignored.
+A value of `erase' means only the last duplicate is kept.
+A value of nil means nothing is ignored."
+  :type '(choice (const :tag "Don't ignore anything" nil)
+                 (const :tag "Ignore consecutive duplicates" t)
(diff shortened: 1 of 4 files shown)


bbatsov:
The proposed change is fine by me. Might be a good idea to mention this configuration option somewhere in the docs as well, otherwise I doubt many people will find it.

projectile.el
```diff
@@ -874,6 +874,16 @@ If the value is nil, there is no limit to the opend buffers count."
   :type 'integer
   :package-version '(projectile . "2.2.0"))
 
+(defcustom projectile-cmd-hist-ignoredups t
+  "Controls when inputs are added to projectile's command history.
+
+A value of t means consecutive duplicates are ignored.
+A value of `erase' means only the last duplicate is kept.
+A value of nil means nothing is ignored."
+  :type '(choice (const :tag "Don't ignore anything" nil)
```
bbatsov, line 883: Don't forget to add a `:package-version` property here.

bbatsov:
Looks great! Thanks!

Issue #1881, opened 2024-02-11
projectile-recentf returns empty list when project was not picked


gessen:
## Expected behavior

Calling `projectile-recentf` with no active project should prompt (by default) for a project and list all recentf files from that project.

## Actual behavior

Calling `projectile-recentf` with no active project prompts for a project but the returned list of files is empty.

## Steps to reproduce the problem

Call `projectile-recentf` from e.g. scratch buffer

## Environment & Version information

X11

### Projectile version information

```
Projectile 2.8.0
```

### Emacs version

*29.2*

### Operating system

*Arch Linux*

gessen:
I think this is because the `projectile-ensure-project`, if project is active, returns a directory after calling `file-truename`. If the project is prompted no such operation takes place. With this change it works correctly:

``` diff
diff --git c/projectile.el w/projectile.el
index d422632..1c6a7f1 100644
--- c/projectile.el
+++ w/projectile.el
@@ -1350,8 +1350,8 @@ See also `projectile-acquire-root'."
   (if dir
       dir
     (cond
-     ((eq projectile-require-project-root 'prompt) (projectile-completing-read
-                                                    "Switch to project: " projectile-known-projects))
+     ((eq projectile-require-project-root 'prompt) (file-truename (projectile-completing-read
+                                                                   "Switch to project: " projectile-known-projects)))
      (projectile-require-project-root (error "Projectile cannot find a project definition in %s" default-directory))
      (t default-directory))))
```

bbatsov:
Hmm, I'm slightly puzzled by this, as I'm not sure why the true name would matter would matter here, as the paths in `projectile-known-projects` are absolute as well. Can you share more details about the project that's causing this issue for you?

gessen:
It's every project that's in home directory (partitition). `projectile-known-project` keeps projects with `~/` prefix, but inside `projectile-recentf-files` [there's a call](https://github.com/bbatsov/projectile/blob/master/projectile.el#L4940) to `expand-file-name` which changes `~/` to `/home/user` which then fails the comparison with `string-prefix-p`.

You can evaluate `projectile-acquire-root` inside any project and it should return expanded (with `/home/user`) path but if you do that inside non-project buffer (like scratch) and pick any project when prompted, it will return a path with `~/`. This small difference is what causes the `projectile-recentf` to return empty file list when called outside of any project.

bbatsov:
[...]

PR #1882, opened 2024-02-11
Fix projectile-recentf with inactive project (#1881)

gessen:
projectile-recentf-files expands recentf files to use canonicalized versions, /home/example/project instead of ~/project. However, if the project is not active and projectile-recentf was called, it would return non-canonicalized directory which then fails the comparison with recentf list.

-----------------

Thanks!


bbatsov:
Thanks!

PR #1890, opened 2024-05-09
Support OSC-SCM

Thaodan:
-------

Support the OSC SCM as used in the open build service.
OSC can be used with vcs-osc in Emacs.

The code is wip but I wanted to post this PR to get some feedback
and show progress working on it.

Osc works a little different than other scms where there two different
types of repositories, one containing packages and the other being a package.
The type that is package works very much like a regular scm where commands
such as those to open project files work as a user would expect.

But for the project type the commands referring to files target the packages the project
contains.

I've implemented the functionality using the builtin XML functionality as that is the easiest
since osc's own files use XML.

I'm not exactly sure how to add test, any help is appreciated.

-----------------

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -41,6 +41,8 @@
 (require 'ibuf-ext)
 (require 'compile)
 (require 'grep)
+(require 'lisp-mnt)
+(require 'xml)
 (eval-when-compile
   (require 'find-dired)
   (require 'subr-x))
@@ -350,6 +352,7 @@ See `projectile-register-project-type'."
     ".pijul"      ; Pijul VCS root dir
     ".sl"         ; Sapling VCS root dir
     ".jj"         ; Jujutsu VCS root dir
+    ".osc"        ; Osc VCS root dir
     )
   "A list of files considered to mark the root of a project.
 The bottommost (parentmost) match has precedence."
@@ -445,7 +448,8 @@ is set to `alien'."
     ".cache"
     ".clangd"
     ".sl"
-    ".jj")
+    ".jj"
+    "*\\.osc$")
   "A list of directories globally ignored by projectile.
 
 Strings that don't start with * are only ignored at the top level
@@ -800,6 +804,10 @@ Set to nil to disable listing submodules contents."
   "Command used by projectile to get the files in a svn project."
   :group 'projectile
   :type 'string)
+(defcustom projectile-osc-command #'projectile-osc-command-files
+    "Command used by projectile to get the files in a svn project."
+  :group 'projectile
+  :type '(function string))
 
 (defcustom projectile-generic-command
   (cond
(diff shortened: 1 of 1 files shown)


bbatsov:
> I'm not exactly sure how to add test, any help is appreciated.

Tests are not mandatory, but if you want to add some you'll need to basically create a dummy project in which to run the project detection commands. There are plenty of similar tests already. 

I'll have to take a closer look at the code, but overall the PR seems reasonable to me.

projectile.el
```diff
@@ -43,6 +43,7 @@
 (require 'compile)
 (require 'grep)
 (require 'lisp-mnt)
+(require 'xml)
```
bbatsov, line 46: I'd suggest moving this in the function that uses it, so we don't slow down loading Projectile needlessly.

projectile.el
```diff
@@ -767,6 +771,10 @@ Set to nil to disable listing submodules contents."
   "Command used by projectile to get the files in a svn project."
   :group 'projectile
   :type 'string)
+(defcustom projectile-osc-command #'projectile-osc-command-files
```
bbatsov, line 774: Please add a blank line before this.

projectile.el
```diff
@@ -767,6 +771,10 @@ Set to nil to disable listing submodules contents."
...
+    "Command used by projectile to get the files in a svn project."
+  :group 'projectile
+  :type '(function string))
```
bbatsov, line 777: Don't forget to add a :package-version property.

projectile.el
```diff
@@ -1489,8 +1497,22 @@ Fallback to a generic command when not in a VCS-controlled project."
...
     ('sapling projectile-sapling-command)
     ('jj projectile-jj-command)
+    ('osc projectile-osc-command)
     (_ projectile-generic-command)))
 
+
+(defun projectile-osc-command-files (&optional root)
+  "Return files of osc vcs, return either packages or files belonging to package."
+  (or root (setq root default-directory))
+  (let* ((files_ (car (xml-parse-file (if (file-exists-p ".osc/_files")
```
bbatsov, line 1507: Why `files_` instead of `files`?

bbatsov: It seems to me a `cond` might be a better fit than the nested `if`s you've opted for.

projectile.el
```diff
@@ -1587,12 +1609,15 @@ If `command' is nil or an empty string, return nil.
 This allows commands to be disabled.
 
 Only text sent to standard output is taken into account."
-  (when (stringp command)
+  (when (or (stringp command)
+             (functionp command))
     (let ((default-directory root))
+      (if (functionp command)
```
bbatsov, line 1615: I'm puzzled by this code, as you call the command, but don't do anything with the result.

projectile.el
Thaodan: > Why `files_` instead of `files`?

Because the file containing the content is called `_files` and `_files` as variable name didn't exactly work.

bbatsov:
Let me know when you feel this is ready for another review.

projectile.el
Thaodan: I do something with the result. The external command if a function is called, returns the list of files and that variable is returned.

projectile.el
Thaodan: Will check using cond for this.

bbatsov:
FWIW, minimal osc support landed on master (17ddffc): `.osc` is now a VCS marker, a top-down-recurring root marker, and globally ignored, with file listing handled by the generic indexing command. That deliberately avoids parsing `.osc` metadata, which would miss files staged with `osc add` and unversioned files.

bbatsov:
Minimal osc support landed on master (17ddffc): `.osc` is now a VCS marker, a top-down-recurring root marker, and globally ignored, with file listing handled by the generic indexing command. That avoids parsing `.osc` metadata, which would miss `osc add`-ed and unversioned files. Thanks for the original patch and the discussion.

Thaodan:
Bozhidar Batsov ***@***.***> writes:

> which would miss `osc add`-ed and unversioned
> files. Thanks for the original patch and the discussion.

Then there was something wrong with the parsing of the file.

The unversioned i.e. files not part of the project are also not taken
into account for project tracked through i.e. git.

Thaodan:
Also osc markers can be nested and files aren't files for osc projects.

Projects have packages (like git submodules kinda) but can't have files.

PR #1894, opened 2024-06-21
Add 'front option to projectile-current-project-on-switch

jtamagnan:
# Foreword

I originally created #1879 but have since deleted that GitHub account,
as such the associated repository and pull request were closed.

This PR reopens that previous PR. #1895 on the other hand implements
Solution 2 of this pull request which I prefer.

# Problem

expect

# Solution

There are two possible solutions that come immediately to mind. I
want to mention both of them for consideration though naturally

 1. Add a new `'front` option for
    `projectile-current-project-on-switch` such that the first item is
    the project associated with the current buffer.
 2. Modify projectile-mode to add the hook to
    `buffer-list-update-hook` such that the current buffer is always
    added to the list of known-projects.

This PR implements solution one as was done in #1879.
#1895 implements Solution 2

## Comparision

The main drawback with Solution 1 is that the newly selected project
will not be added to the projects list.

The main drawback of Solution 2 is that it will not work on versions
of emacs earlier than 28.1 because of how `buffer-list-update-hook`
behaves on those older versions.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#1875](https://github.com/bbatsov/projectile/pull/1875): Add support for Sapling VCS.
 * [#1876](https://github.com/bbatsov/projectile/pull/1876): Add support for Jujutsu VCS.
 * [#1877](https://github.com/bbatsov/projectile/pull/1877): Add custom variable `projectile-cmd-hist-ignoredups`.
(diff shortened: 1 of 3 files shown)


jtamagnan:
Closing in favor of https://github.com/bbatsov/projectile/pull/1895 as per https://github.com/bbatsov/projectile/pull/1895#issuecomment-2289308636

PR #1895, opened 2024-06-21
Add #'projectile-track-known-projects-find-file-hook to 'buffer-list-update-hook

jtamagnan:
# Foreword

I originally created #1879 but have since deleted that GitHub account,
as such the associated repository and pull request were closed. The
new version of that PR can be found at #1894.

This PR implements Solution 2 as described in #1894.

# Problem

expect

# Solution

There are two possible solutions that come immediately to mind. I
want to mention both of them for consideration though naturally

 1. Add a new `'front` option for
    `projectile-current-project-on-switch` such that the first item is
    the project associated with the current buffer.
 2. Modify projectile-mode to add the hook to
    `buffer-list-update-hook` such that the current buffer is always
    added to the list of known-projects.

This PR implements Solution 2.
#1894 implements Solution 1.

## Comparision

The main drawback with Solution 1 is that the newly selected project
will not be added to the projects list.

The main drawback of Solutionk 2 is that it will not work on versions
of emacs earlier than 28.1 because `buffer-list-update-hook` works
differently.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -6284,14 +6284,24 @@ Otherwise behave as if called interactively.
       (projectile-discover-projects-in-search-path))
     (add-hook 'project-find-functions #'project-projectile)
     (add-hook 'find-file-hook 'projectile-find-file-hook-function)
-    (add-hook 'projectile-find-dir-hook #'projectile-track-known-projects-find-file-hook t)
-    (add-hook 'dired-before-readin-hook #'projectile-track-known-projects-find-file-hook t t)
+    ;; Add hooks to track which buffer is currently active.
+    ;; Note - In Emacs 28.1 `buffer-list-update-hook' was modified to no
+    ;; longer run on temporary buffers, this allows us to use it to track
+    ;; changes to the active buffer instead of relying on more specific hooks
+    ;; such as `dired-before-readin-hook'.
+    (if (version<= "28.1" emacs-version)
+        (add-hook 'buffer-list-update-hook #'projectile-track-known-projects-find-file-hook t)
+      (add-hook 'projectile-find-dir-hook #'projectile-track-known-projects-find-file-hook t)
+      (add-hook 'dired-before-readin-hook #'projectile-track-known-projects-find-file-hook t t))
+
     (advice-add 'compilation-find-file :around #'compilation-find-file-projectile-find-compilation-buffer)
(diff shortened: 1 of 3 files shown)


projectile.el
```diff
@@ -6272,14 +6272,19 @@ Otherwise behave as if called interactively.
       (projectile-discover-projects-in-search-path))
     (add-hook 'project-find-functions #'project-projectile)
     (add-hook 'find-file-hook 'projectile-find-file-hook-function)
-    (add-hook 'projectile-find-dir-hook #'projectile-track-known-projects-find-file-hook t)
-    (add-hook 'dired-before-readin-hook #'projectile-track-known-projects-find-file-hook t t)
+    (if (version<= "28.1" emacs-version)
```
bbatsov, line 6275: I'd add a comment here explaining the need for this check and the hook in general.  (e.g. something similar to what you've written in the PR's description)

bbatsov:
This solution is preferable to me, so let's go with it. I find it to be cleaner, and most Emacs users are likely using Emacs 28+ anyways.

projectile.el
jtamagnan: Great point, I've gone ahead and added a summary. I can expand on it a bit more and include a link to the (relevant NEWS entry)[https://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS.28?h=master#n2956] but hopefully this is clear enough.

jtamagnan:
Thank you for the review

---

> This solution is preferable to me, so let's go with it. I find it to be cleaner, and most Emacs users are likely using Emacs 28+ anyways.

Agreed, this one seems more robust. I've closed out the other pull-request.

bbatsov:
Thanks!

jtamagnan:
@bbatsov thank you for the package, the conversation, and the reviews

dgutov:
Hi!

FWIW, this might be suboptimal because the "temporary" buffers that Emacs 28+ treats specially are just the ones that had been created with non-nil second (new and optional) argument to `get-buffer-create`.

Which probably means that the contents of this hook will still run in many cases of buffers created for "temporary" work. See https://github.com/company-mode/company-mode/pull/1492 or the Org related commit above referencing this issue.

PR #1896, opened 2024-06-29
Fix tests to not modify top level .dir-locals.el file

jtamagnan:
# Problem

When running the projectile tests locally I noticed that my top-level
`.dir-locals.el` file would get a `fooo . 1` added to it.

# Solution

Modify the test `projectile-add-dir-local-variable` such that it
creates its own `.dir-locals.el` file which it can modify. This was a
little bit finicky because `projectile-add-dir-local-variable` doesn't
restore the selected window and buffer after executing so I've fixed
that function to restore the windows as they were beforehand. When
applying a similar fix to the `...delete...` function I had to improve
that test as well

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2503,9 +2503,10 @@ With a prefix arg INVALIDATE-CACHE invalidates the cache first."
 Parameters MODE VARIABLE VALUE are passed directly to `add-dir-local-variable'."
   (let ((inhibit-read-only t)
         (default-directory (projectile-acquire-root)))
-    (add-dir-local-variable mode variable value)
-    (save-buffer)
-    (kill-buffer)))
+    (save-selected-window
+      (add-dir-local-variable mode variable value)
+      (save-buffer)
+      (kill-buffer))))
 
 ;;;###autoload
 (defun projectile-delete-dir-local-variable (mode variable)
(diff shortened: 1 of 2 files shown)


jtamagnan:
@bbatsov, I was wondering if you had had some time to look at this. I would really appreciate some feedback here, I've found it helpful when running tests for some other proposed changes. Sorry for the bump on this.

bbatsov:
Your proposed solution looks good to me. Sorry about the slow response.

bbatsov:
Btw, it seem this change broke the tests on Emacs 29.1. Please, check the CI report.

jtamagnan:
Thanks for the integration, really appreciate it. I've pushed this same commit to my fork of the repository and it seems to work without issue. https://github.com/jtamagnan/projectile/actions/runs/10392430247 We can see that the git SHA are the same. I also can't reproduce this locally.

I'm not sure what the source of the flakiness is or what exactly is failing in the test. Maybe we can rerun the action here to see if it'll help. I can investigate further though to see what is causing this inconsistency.

bbatsov:
Thanks for looking into this. Might just be flakiness indeed.

PR #1405, opened 2019-04-15
Add Bloop Configuration

isomarcte:
Bloop is a Scala Build Server. Projects using bloop will likely have another build system referenced locally, e.g. a `build.sbt` file, but the ideal method to compile/test will be through the `bloop` command to the `bloop` daemon.

-----------------

Before submitting a PR make sure the following things have been done (and denote this
by checking the relevant checkboxes):

- [x] The commits are consistent with our [contribution guidelines](../CONTRIBUTING.md)
- [ ] You've added tests (if possible) to cover your change(s)
- [x] All tests are passing (`make test`)
- [x] The new code is not generating bytecode or `M-x checkdoc` warnings
- [x] You've updated the changelog (if adding/changing user-visible functionality)
- [ ] You've updated the readme (if adding/changing user-visible functionality)

Thanks!

PR #1901, opened 2024-08-23
Remove Bloop from Project File List

ZachFontenot:
**Bloop file is always present in Home Directory**

Hey Bbatsov, love all of your work. I was having an issue with my home directory being indexed on multiple computers, and I couldn't figure out why until I manually went through and checked against all of the `projectile-project-root-files`, since I checked against the obvious suspects and couldn't find it.

Anyway, the culprit was a `.bloop` directory in my home directory, so I checked on why this was there. According to the Bloop [documentation](https://scalacenter.github.io/bloop/docs/server-reference), this is where the bloop server places its config by default.

This was the most obvious solution to me, which is to remove this by default, since mill and sbt projects are already covered. 

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -3496,13 +3496,7 @@ a manual COMMAND-TYPE command is created with
 (projectile-register-project-type 'clojure-cli '("deps.edn")
                                   :project-file "deps.edn"
                                   :test-suffix "_test")
-(projectile-register-project-type 'bloop '(".bloop")
-                                  :project-file ".bloop"
-                                  :compile "bloop compile root"
-                                  :test "bloop test --propagate --reporter scalac root"
-                                  :src-dir "src/main/"
-                                  :test-dir "src/test/"
-                                  :test-suffix "Spec")
+
 ;; Ruby
 (projectile-register-project-type 'ruby-rspec '("Gemfile" "lib" "spec")
                                   :project-file "Gemfile"
--- a/projectile.el
+++ b/projectile.el
@@ -3477,6 +3477,14 @@ a manual COMMAND-TYPE command is created with
                                   :test "mill __.test"
                                   :test-suffix "Test")
 
+(projectile-register-project-type 'bloop '(".bloop/bloop.settings.json")
+                                  :project-file ".bloop/bloop.settings.json"
+                                  :compile "bloop compile root"
+                                  :test "bloop test --propagate --reporter scalac root"
+                                  :src-dir "src/main/"
+                                  :test-dir "src/test/"
+                                  :test-suffix "Spec")
+
 ;; Clojure
 (projectile-register-project-type 'lein-test '("project.clj")
                                   :project-file "project.clj"
--- a/projectile.el
+++ b/projectile.el
@@ -3477,9 +3477,7 @@ a manual COMMAND-TYPE command is created with
                                   :test "mill __.test"
                                   :test-suffix "Test")
 
-(projectile-register-project-type 'bloop '(".bloop/bloop.settings.json")
-                                  :project-file ".bloop/bloop.settings.json"
-                                  :compile "bloop compile root"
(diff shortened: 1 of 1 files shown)


bbatsov:
I haven't used Bloop, so I'm not sure what's the best course of action here. I see it was introduced 5 years ago in Projectile (see https://github.com/bbatsov/projectile/pull/1405). I'm guessing one way to solve this would be to look for both SBT and bloop markers together (if there's also some project-specific Bloop configuration). 

> Hey Bbatsov, love all of your work.

Thanks for the kind words! 🙇

ZachFontenot:
> I haven't used Bloop, so I'm not sure what's the best course of action here. I see it was introduced 5 years ago in Projectile (see #1405). I'm guessing one way to solve this would be to look for both SBT and bloop markers together (if there's also some project-specific Bloop configuration).
> 
> > Hey Bbatsov, love all of your work.
> 
> Thanks for the kind words! 🙇

Alright, I'll look into how to do that, do you know off-hand of any project configs that currently do this? Or will it need to be a new feature? Also, I'm not sure about why those tests are failing, is that maybe something else that's in Master unrelated to my change?

LaurenceWarne:
I've run into this before too, as someone who works on sbt/mill projects it makes sense to remove to me as I'd prefer to have those projects recognised as sbt/mill anyway (a small caveat is that projectile appears to support all project types mentioned at https://scalacenter.github.io/bloop/ except `pants`, so we could add that for full coverage?).

In terms of adding bloop project markers to sbt and friends, I'm not sure it would add much value? That a project is sbt/mill if and only if the project root contains `build.sbt`/`build.sc` seems a fairly safe assumption to me, plus the `.bloop` marker in the project directory would only be present AFAIK if bloop as been run at least once and so wouldn't be present on a newly cloned project anyway.

If the bloop project type would be removed, I think the main (breaking?) change would be that sbt/mill projects start being recognised as `sbt`/`mill` rather than `bloop` since I think the latter has a higher precedence, which would mean the compile/test commands would change (as mentioned in https://github.com/bbatsov/projectile/pull/1405).  Though this could be changed e.g. via:

```elisp
   'sbt
   :compile "bloop compile root")
```

A possible alternate fix where the `bloop` project type could be kept could be to change the `bloop` marker file to `.bloop/bloop.settings.json`, which appears to be present in bloop project directories, but not in `~/.bloop` (though I can't find any doc confirming this).

bbatsov:
@LaurenceWarne I'm fine with whatever you suggest on the topic.

LaurenceWarne:
I asked on the bloop discord channel, and it sounds like `.bloop/bloop.settings.json` is fairly safe.  Thoughts on changing the marker files for `bloop` to that instead of deleting the project type @ZachFontenot?

That way I think we don't have a breaking change, and it should hopefully fix your issue.  Though as mentioned before I think there is a good argument for removing the project type completely - but runs the risk of breaking someone somwehere's workflow :sweat_smile: (https://xkcd.com/1172/).

ZachFontenot:
Yeah happy to switch it to that, thanks for helping resolve this

bbatsov:
@ZachFontenot Can you update the PR?

LaurenceWarne:
I think setting `:project-file` is unnecessary, I don't think there is an equivalent of a main project file for bloop

ZachFontenot:
Okay, I've removed the `project-file` key altogether since it's probably not doing anything useful in this context

projectile.el
```diff
@@ -3477,6 +3477,12 @@ a manual COMMAND-TYPE command is created with
                                   :test "mill __.test"
                                   :test-suffix "Test")
 
+(projectile-register-project-type 'bloop '(".bloop/bloop.settings.json")                                          :compile "bloop compile root"
```
LaurenceWarne, line 3480: Looks like a slight formatting issue :slightly_smiling_face:

LaurenceWarne:
Left small comment on some formatting, otherwise LGTM!

PR #1904, opened 2024-09-07
Ignore `CVS` by default

catap:


bbatsov:
Seems this caused a breakage (see #1904), so we'll have to revert it.

Issue #1908, opened 2024-10-04
Projectile and org-cycle cause delays during the expansion of the org tables.


arael:
Expanding org tables using tab has become extremely sluggish. It takes two to three seconds for the table to be expanded properly. The function bound to my tab key is `org-cycle`.

If I remove projectile from my installation, then the delays stop.

## Expected behavior

No delays in expanding the org tables using tab (org-cycle).

## Actual behavior

There is a delay of two to three seconds when expanding org tables using the tab key (org-cycle).

## Steps to reproduce the problem

With projectile in your Emacs installation, visit an org file, type | then a letter, and then press tab. You should see a delay.

### Projectile version information

```
Projectile version: 20240901.1554
```

### Emacs version

*29.3*

### Operating system

*Ubuntu 24.04.1*

Here is the Emacs profiler log, I hope it can be useful:

```
	3619  86% - command-execute
	2938  70%  - funcall-interactively
	2931  70%   - org-cycle
	2929  69%    - funcall-interactively
	2929  69%     - org-table-next-field
	2921  69%      - org-table-align
	1654  39%       - cl-mapcar
	1654  39%        - cl--mapcar-many
	1652  39%         - org-table--align-field
	1629  38%          - org-string-width
	1584  37%           - set-window-buffer
	1584  37%            - apply
	1477  35%             - #<subr set-window-buffer>
	1477  35%              - record-window-buffer
	1462  34%               - projectile-track-known-projects-find-file-hook
	 570  13%                - projectile-project-p
	 568  13%                 - projectile-project-root
	 554  13%                  - cl-some
	 554  13%                   - #<compiled 0x102e39d81079d55>
	 280   6%                    - file-truename
	 217   5%                     - file-truename
	 166   3%                      - file-truename
	 115   2%                       - file-truename
	  78   1%                        - file-truename
	  45   1%                         - file-truename
	  44   1%                          - tramp-completion-file-name-handler
	  39   0%                           - tramp-run-real-handler
	  39   0%                              apply
	 190   4%                    - projectile-root-marked
	 190   4%                     - projectile-root-bottom-up
	 189   4%                      - projectile-locate-dominating-file
	 155   3%                       - #<compiled -0x17ef8ac10091690f>
	  94   2%                        - mapcar
	  93   2%                           #<compiled 0x759ed441373c99c>
	  60   1%                        - cl-some
	  60   1%                           #<compiled 0x13403800b8d4454>
	   1   0%                    cl-subst
[...]

PR #1910, opened 2024-10-09
Revert "Add #'projectile-track-known-projects-find-file-hook to 'buff…

codeasone:
Responding to https://github.com/bbatsov/projectile/issues/1908

This reverts commit 3c92d28c056c3553f83a513a35502b4547d29319.

Addresses an apparent performance regression

- [-] You've added tests (if possible) to cover your change(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#1910](https://github.com/bbatsov/projectile/pull/1910): Reverts [#1895](https://github.com/bbatsov/projectile/pull/1895) as those changes appear to cause a significant performance regression across a number of use-cases.
+
 ### New features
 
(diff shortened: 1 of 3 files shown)


jtamagnan:
Hi @bbatsov,

Sorry for not having taken a look at this and you needing to deal with it. I agree with reverting this change and I'm glad that its been reverted before `projectile`'s next full release. 

Although some packages properly declare buffers that they use as `temporary` such that the `buffer-list-update-hook` is not run others like org-mode don't seem to do that - https://github.com/bzg/org-mode/blob/b6dbf8881076191e1351d7cd15e26547a2531fea/lisp/org-macs.el#L1193. Sadly it seems like the Emacs ecosystem isn't fully ready for `buffer-list-update-hook` to be used in this way, and although it "feels right"  it simply isn't and it degrades user experience.

Again sorry that these issues came up and that you had to deal with them.

Best,
Jules

bbatsov:
Dropping here a link to #1895 and some extra context for the future https://github.com/bbatsov/projectile/pull/1895#issuecomment-2336853756

Issue #1811, opened 2022-11-08
`projectile-globally-ignored-directories` not working for `projectile-ripgrep`


TheJJ:
## Expected behavior

Directories and pattterns from `projectile-globally-ignored-directories` should correctly be used by `projectile-ripgrep`.

## Actual behavior

The patterns are regexes, but `ripgrep` expects globs.
`ripgrep` is invoked with `--hidden --glob ! ^\.git$`, and `.git` is not a valid glob pattern.

## Steps to reproduce the problem

* be in a git project
* `echo rolf > .git/lolfile`
* projectile-ripgrep for `rolf`
* see that there's a result in `.git/lolfile` even though .git should have been ignored.
* a simpler test: `rg --hidden --glob '! ^\.git$' rolf` yields the same result, since that is how the ignorepatterns are passed.

## Possible fix
The regexes have to be converted to globs so `ripgrep` will like them.
Alternatively `ripgrep` must gain a new regex-file-filter option.

### Projectile version information
```
Projectile version: 20221101.848
```

### Emacs version

*GNU Emacs 29.0.50*

Heviwulansari:
> ## Expected behavior
> Directories and pattterns from `projectile-globally-ignored-directories` should correctly be used by `projectile-ripgrep`.
> 
> ## Actual behavior
> The patterns are regexes, but `ripgrep` expects globs. `ripgrep` is invoked with `--hidden --glob ! ^\.git$`, and `.git` is not a valid glob pattern.
> 
> ## Steps to reproduce the problem
> * be in a git project
> * `echo rolf > .git/lolfile`
> * projectile-ripgrep for `rolf`
> * see that there's a result in `.git/lolfile` even though .git should have been ignored.
> * a simpler test: `rg --hidden --glob '! ^\.git$' rolf` yields the same result, since that is how the ignorepatterns are passed.
> 
> ## Possible fix
> The regexes have to be converted to globs so `ripgrep` will like them. Alternatively `ripgrep` must gain a new regex-file-filter option.
> 
> ## Environment & Version information
> ### Projectile version information
> ```
> Projectile version: 20221101.848
> ```
> 
> ### Emacs version
> _GNU Emacs 29.0.50_

kisp:
It would be nice if this could be fixed. :slightly_smiling_face: 

I guess - as a workaround - the easiest thing would be to redefine `projectile-ripgrep` as `my/projectile-ripgrep`.

But for a real fix, probably the main question is how it should be done...

[...]

PR #1914, opened 2024-10-25
Fix searching ignored directories for ripgrep

shutosheep:
Currently `projectile-ripgrep` does not properly hide directories of `projectile-globally-ignored-directories` as mentioned in issue #1811 . `projectile-ripgrep` produces following arguments for searching.

```bash
rg  --fixed-strings --hidden --glob !TAGS --glob !^\.idea$ --glob !^\.vscode$ ...
```

I found that there are two problems:
- Both ignored file and directory name is not quoted with `'`
- Directory name includes regexp which I believe `--glob` option does not support
  - Looking at [ripgrep's `-g` flag docstring](https://github.com/BurntSushi/ripgrep/blob/79cbe89deb1151e703f4d91b19af9cdcc128b765/crates/core/flags/defs.rs#L2482), it says that it uses gitignore rules

In this PR, I made both globally ignored file and directory names to be quoted with `'` and remove unnecessary regexp (`^\` and `$`) from `projectile-globally-ignored-directories`. This results in making arguments such as...

```bash
rg  --fixed-strings --hidden --glob '!TAGS' --glob '!.idea' --glob '!.vscode' ...
```

resulting with ripgrep properly ignoring contents of `projectile-globally-ignored-directories`.

By the way I am new to emacs lisp, so feel free to edit my changes.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -4388,9 +4388,12 @@ installed to work."
    (list (projectile--read-search-string-with-default
           (format "Ripgrep %ssearch for" (if current-prefix-arg "regexp " "")))
          current-prefix-arg))
-  (let ((args (mapcar (lambda (val) (concat "--glob !" val))
+  (let ((args (mapcar (lambda (val) (concat "--glob '!" val "'"))
                       (append projectile-globally-ignored-files
-                              projectile-globally-ignored-directories))))
+                              ;; remove "^\" and "$"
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks for the PR - it reminded me that #1762 was problematic and I decided to revert it instead, as I think that's the better solution.

PR #1916, opened 2024-10-31
add variable to ignore specific projects

dradetsky:
I want to add a git repo to my home directory in order to track dotfiles (I'm aware there are other approaches that don't involve creating `~/.git/`, but for reasons beyond the scope of this PR, I want to do it this way). When I do this, in doom-emacs `doom doctor` complains

```
    ! Your $HOME is recognized as a project root
      Emacs will assume $HOME is the root of any project living under $HOME. If this
      isn't desired, you will need to remove ".git" from
      `projectile-project-root-files-bottom-up' (a variable), e.g.
      
        (after! projectile (setq projectile-project-root-files-bottom-up (remove ".git"
          projectile-project-root-files-bottom-up)))
```

Maybe I'm misunderstanding something, but it seems like the resolution here isn't something I want to do. I still want projectile to recognize dirs with `.git` as project roots, I just want it to not consider `$HOME` to be the universal project root, as threatened by `doom doctor`. Therefore, I need it to say that one or more named directories are not project roots, regardless of what the root-determination machinery would say otherwise.

So I add a variable to hold a list of such directories, and consult it inside `projectile-locate-dominating-file`.

NOTE: the PR is intentionally missing some required cleanup (like comments) in case there is something fundamentally wrong with my whole approach (I suck at elisp). That's why it's marked as a draft. AFAIK, the code itself does what it should, but I could be wrong.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master (unreleased)
 
+* [#number](link): Add variable `projectile-non-project-roots`
 * [#1910](https://github.com/bbatsov/projectile/pull/1910): Reverts [#1895](https://github.com/bbatsov/projectile/pull/1895) as those changes appear to cause a significant performance regression across a number of use-cases.
 
 ### New features
--- a/doc/modules/ROOT/pages/configuration.adoc
+++ b/doc/modules/ROOT/pages/configuration.adoc
@@ -515,3 +515,14 @@ NOTE: The project name & type will not appear when editing remote files
 Projectile supports https://cmake.org/cmake/help/git-stage/manual/cmake-presets.7.html[CMake presets]. Preset support is disabled by default, but can be enabled by setting `projectile-enable-cmake-presets` to non-nil. With preset-support enabled Projectile will parse the preset files and present the command-specific presets when executing a lifecycle command. In addition a `*no preset*` option is included for entering the command manually.
 
 NOTE: Preset support requires a CMake version that supports preset and for `json-parse-buffer` to be available.
+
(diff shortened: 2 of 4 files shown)


bbatsov:
The approach seems sound, although I have some small concern about how this will affect performance, as it's an extra check in a function that's used quite often. On the other hand - by default this variable will be `nil`, so likely the overhead then will be close to none.

One alternative approach to consider would be to use some variable marker set via `.dir-locals.el`, but I assume some people might not like this, as it will require them to have `.dir-local.el` in any folder they'd like to ignore.

dradetsky:
@bbatsov if you're concerned about the performance (I really have no idea whether you're right to be concerned or not), we could check whether `projectile-non-project-roots` is non-nil before we even bother to expand file & create the closure that we'd map over `projectile-non-project-roots`. That way, instead of some extra misc elisp stuff per call, it's just an extra compare-to-nil per call (at least for those who aren't using this variable), which I'd assume wouldn't matter (although I haven't benchmarked obviously).

I didn't do this to begin with because I was concerned about the readability of the code, so I tried to avoid getting to clever. But if you think it's worth it here I can do it.

bbatsov:
I don't think an explicit nil check will be much of a difference - iterating over an empty/nil list is more or less the same. Just pursue your original approach.

dradetsky:
I subsequently decided not to store dotfiles via creating a `~/.git` so this is no longer relevant to my needs. You can close the PR if you want (or IDK, keep it around & finish yourself).

Issue #1744, opened 2022-01-25
`projectile-go-project-p` doesn't work when under sub-folder of the one has `project marker`


ccqpein:
My project structure is:

```
|-A
  |- .git
  |- B
    |- go.mod
    |- main.go
```

## Expected behavior

When I open `main.go`, the `projectile-bookmarks.eld` in my `emacs.d` should added `/A/B`. 

## Actual behavior

in my `projectile-bookmarks.eld`, it added `/A`

## Steps to reproduce the problem

`M-x projectile-clear-known-projects` , then restart emacs.

then open the `main.go` file located like I showed upper. `cat ~/.emacs.d/projectile-bookmarks.eld` to see which path added to projectile. 

### Projectile version information
I install it from elpa, but the `M-x projectile-version` gives me:

```
Projectile nil
```

### Emacs version

`GNU Emacs 29.0.50 (build 1, x86_64-apple-darwin21.2.0, NS
 appkit-2113.20 Version 12.1 (Build 21C52)) of 2022-01-23`

### Operating system

`macOS Monterey 12.1`

ccqpein:
I have read [doc](https://docs.projectile.mx/projectile/projects.html#adding-custom-project-types) and play around a bit with `projectile`. 

If, first I manually add 

```elisp
(projectile-register-project-type
   'go
   '("go.mod")
   :project-file "go.mod")
```

secondly, change `projectile-project-root-functions` to 

```elisp
'(projectile-root-local
     projectile-root-top-down
     projectile-root-bottom-up
     projectile-root-top-down-recurring)
```

It works fine (which means add `/A/B` inside `projectile-bookmarks.eld` instead of `/A`.

ccqpein:
I give several `(message)` inside source code and looks like it runs `projectile-project-root` right after I open my `main.go`. Then, put the `/A` inside `projectile-project-root-cache` because `projectile-root-bottom-up` gonna return `/A`. I think it `projectile-root-top-down` should return value `/A/B` before `projectile-root-bottom-up` but actually it returns nothing.

ccqpein:
To be honest, I am not sure if it is a problem or be designed this way. 

Because `projectile-root-bottom-up`'s pair custom value `projectile-project-root-files-bottom-up` are all the folders. And the `projectile-project-root-files` (`projectile-root-top-down` uses) are all the project configuration files (especially `projectile-register-project-type` hard codes a lot of languages' configuration files). 

[...]

PR #1928, opened 2025-02-12
Try to fix `projectile-go-project-p` under sub-folder of the one has project marker

ccqpein:
Try to fix the issue #1744 

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -369,8 +369,8 @@ containing a root file."
 (defcustom projectile-project-root-functions
   '(projectile-root-local
     projectile-root-marked
-    projectile-root-bottom-up
     projectile-root-top-down
+    projectile-root-bottom-up
     projectile-root-top-down-recurring)
   "A list of functions for finding project root folders.
 The functions will be run until one of them returns a project folder.


bbatsov:
Thanks for taking a crack at this. I don't think we can flip the default order of `projectile-project-root-functions` here, though - it isn't specific to Go. Moving `projectile-root-top-down` ahead of `projectile-root-bottom-up` changes the root-detection precedence for *every* project: instead of the nearest (innermost) marker winning, the outermost one would. That quietly changes which directory is treated as the project root in nested repos, submodules and monorepos for everyone, which is a much bigger and riskier change than the Go case that prompted it.

For the original issue I'd rather not touch the global default - better handled with a Go-specific root function/marker, or by the affected user reordering `projectile-project-root-functions` in their own config (it's a defcustom for exactly this reason).

I'll leave this open as a design discussion rather than merge as-is.

bbatsov:
As discussed, reordering the default root functions changes detection precedence for everyone, and it turns out it would not even fix #1744 under the default config (the go type registers a function marker with no `:project-file`, so `go.mod` never enters `projectile-project-root-files`). For nested projects the supported approaches are a `.projectile` marker in the subproject, or reordering `projectile-project-root-functions` in your own config.

Issue #1926, opened 2025-02-07
Repositories are polluted with .projectile.cache files


drobilla:
From the documentation:

> Prior to Projectile 2.9 the cache for all projects was serialized to the same file. In Projectile 2.9 this was changed and now each project has it’s own cache file relative to the project’s root directory. 

I can see how this is a technical improvement in some respects, but in practice, it's really problematic.  Now, any time a file is opened in any project, that repository becomes "dirty" because of untracked files.  I don't think tools should ever do this sort of thing by default, but they absolutely shouldn't ever do it unconditionally.  I'd rather not disable caching entirely, though, because I work on some pretty large projects.  Would it be possible to bring back the old behaviour under a configuration option, or is there some other workaround to avoid projectile stomping all over the filesystem?

As the simplest resolution I can think of, I could maybe live with an include list so I can configure things to only do this for my own projects, but the prospect of adding `.projectile.cache` to a ton of `.gitignore` files is really unappealing and, in some cases, impossible anyway.  Ideally, I just want the old behaviour back, or in general, some way to have a cache and also never have files written to repositories.  Like many, I use emacs as my primary code _reading_ tool, but suddenly my reading tool is _writing_ to repositories!

matheusvcorrea:
+1 for an alternative behavior.

bbatsov:
I'm guessing you're not running the most recent snapshot, as the behavior you've mentioned lasted for like 2 days, before I made the cache "transient" by default. (see https://github.com/bbatsov/projectile/commit/a4a6cacd908bc614d60c6233a7f224baebfe1178)

There are reasons why I think have the cache relative to the project dir is not a bad idea in general:

- the cache contents are relative to the project root, and in plain text, so they are actually usable by anyone if someone commits it
- make it a bit easier to discover the cache

I've toyed with the idea of making the cache location configurable to something like XDG Base and this might happen in the future, but it would require hashing the project paths and adding some simple way to go to the cache if someone wants to check it, so I've opted for the simplest possible approach as the starting point. I acknowledge that the cache should not be persistent by default, and since I've changed this already and I doubt many people will enable persistent caching I'm guessing this issue is mostly resolved.

drobilla:
> I'm guessing you're not running the most recent snapshot, as the behavior you've mentioned lasted for like 2 days

I'm pulling from MELPA.  Looks like I had 20250130, I just updated to 20250205 and the file's indeed no longer created, thanks.

[...]

PR #1929, opened 2025-02-12
[#1926] Fix more place to not create any cache file when persistent c…

leuven65:
Fix some other case that the file ".projectile-cache.eld" is created when "projectile-enable-caching" is not "persistent".


bbatsov:
Thanks!

PR #1930, opened 2025-02-13
Add build.mill as a project file name for the mill project type

LaurenceWarne:
Hi, this PR adds `build.mill` as a project file name for the mill project type, it seems to be the recommended extension now, see https://mill-build.org/mill/scalalib/intro.html.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -2902,7 +2902,8 @@ ones and overrule settings in the other lists."
 
 A project type is defined by PROJECT-TYPE, a set of MARKER-FILES,
 and optional keyword arguments:
-PROJECT-FILE the main project file in the root project directory.
+PROJECT-FILE the main project file in the root project directory.  It may be a
+             single file or a list of possible files.
 COMPILATION-DIR the directory to run the tests- and compilations in,
 CONFIGURE which specifies a command that configures the project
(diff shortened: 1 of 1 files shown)


projectile.el
```diff
@@ -3538,8 +3547,8 @@ a manual COMMAND-TYPE command is created with
                                   :test "sbt test"
                                   :test-suffix "Spec")
 
-(projectile-register-project-type 'mill '("build.sc")
-                                  :project-file "build.sc"
+(projectile-register-project-type 'mill #'projectile-mill-project-p
+                                  :project-file '("build.sc" "build.mill")
```
bbatsov, line 3551: Won't this require some code changes or it was already accounted for?

LaurenceWarne: As in allowing lists? - If so I believe it should already be accounted for: https://github.com/bbatsov/projectile/blob/db52cf48b6cc3af892ba699ec13c61334bd39d1b/projectile.el#L2934

bbatsov: Ah, ok. I just noticed you updated the docstrings, but there were no other changes, so it seemed prudent to ask about this. :-)

LaurenceWarne: Ah yeap, probably should have pointed that out, sorry!

PR #1944, opened 2025-05-26
Use template for jujutsu file listing

joeri:
In jujutsu users can configure the default behaviour of `jj file list` so that it no longer puts out file names separated by newlines (which is the default behaviour). This means the output of `jj file list` can't be trusted to be stable. On the other hand the same mechanism allows you to output the files separated with a null byte without invoking `tr`, this has the additional benefit that filenames with newlines in them are now working.

-----------------

Thanks!

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -757,7 +757,7 @@ Set to nil to disable listing submodules contents."
   :group 'projectile
   :type 'string)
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

Issue #1936, opened 2025-02-27
Invalidate projectile-project-root-cache


raxod502:
I find that a frequent operation is:
* Attempt to `projectile-find-file`
* Realize my current directory is not marked as a project
* Create `.projectile` file
* Attempt to `projectile-find-file` again
* To my surprise, Projectile still thinks the directory is not marked as a project, even though it is
* Attempt to use `projectile-invalidate-cache`
* To my further surprise, Projectile forces me to select a project to clear the cache for, even though what I need is just to clear `projectile-project-root-cache`, which is not associated with any project

Suggestions for improvement:
* Detect cases where `projectile-project-root-cache` should be cleared automatically to avoid user confusion
* Add a user command for clearing `projectile-project-root-cache` without needing to select a project
* Document the fact that this cache exists (since it affects user-visible behavior)

I could help out with this, but creating an issue first for discussion.

raxod502:
The nil return value of the entry in `projectile-project-root-functions` is cached if no project is detected:

https://github.com/bbatsov/projectile/blob/0163b335a18af0f077a474d4dc6b36e22b5e3274/projectile.el#L1335-L1340

I'm not sure if that is what you meant.

bbatsov:
Ah, yeah - I had forgotten about this.

raxod502:
Don't close this issue. This is an automatic message by [Fresh](https://github.com/raxod502/fresh-bot) - a bot against stale bots.

/remove-lifecycle stale

PR #1945, opened 2025-06-11
Handle the projects cache better

raxod502:
Handle part of https://github.com/bbatsov/projectile/issues/1936 by allowing for `M-x projectile-invalidate-cache` to still clear the global cache even if the user invokes it from outside of any project, and no longer force the user to select a project to invalidate the cache for in that case. Also, document how this command may be used for troubleshooting.

-----------------

- [x] You've updated the [docs](../blob/master/doc/modules/ROOT/pages) (when adding new project types, configuration options, commands, etc)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Set `projectile-auto-discover` to `nil` by default.
 * [#1943](https://github.com/bbatsov/projectile/pull/1943): Consider `projectile-indexing-method` to be safe as a dir-local variable if it is one of the preset values.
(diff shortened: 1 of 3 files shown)


raxod502:
I haven't added tests since this is a fairly small logic change and there aren't existing tests for `projectile-invalidate-cache`, so I'm not sure offhand what if anything would need to be mocked, but if it is desired, I can look into it and figure it out.

raxod502:
The snapshot tests appear to also be failing on master, so let me know if I should look at those as well.

bbatsov:
Thanks!

Seems the tests are failing because of some upstream deprecations - I'll handle those myself.

PR #1892, opened 2024-05-26
Add category metadata to completing-read

juergenhoetzel:
I added the metadata only to the default Emacs `completing-read`. as I am not familiar with the 3rd party systems helm and ivy.

- [x] The commits are consistent with our [contribution guidelines](../blob/master/CONTRIBUTING.md)
- [ ] You've added tests (if possible) to cover your change(s)
- [x] All tests are passing ([`eldev test`](https://github.com/doublep/eldev))
- [x] The new code is not generating bytecode or `M-x checkdoc` warnings
- [ ] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)
- [ ] You've updated the readme (if adding/changing user-visible functionality)

I'm not sure about checking the last 2 checkboxes. The changes are only ***visible*** if using third-party packages like
[marginalia](https://github.com/minad/marginalia) and [embark](https://github.com/oantolin/embark) which use the metadata.

`projectile-switch-project` without metadata:

image

with metadata: 
image

PR #1957, opened 2025-07-26
Identification of caller for ivy-read

ThibautVerron:
Hi,

With `ivy`, projectile uses `ivy-read` for completion. This function takes a named parameter `:caller` that allows `ivy` to change its behavior (e.g. `ivy-rich` annotations, or `display-buffer` behavior). As it stands, projectile always sets this parameter to `projectile-completing-read`. 

This PR changes it to set a different caller according to what is being read: `projectile-read-project` for a project name, `projectile-read-buffer` for a buffer name, `projectile-read-file` for a file name, `projectile-read-directory` for a directory.

Based on the changelog, the purpose appears to be similar to https://github.com/bbatsov/projectile/pull/1892.

It is also similar to what [counsel-projectile](https://github.com/ericdanan/counsel-projectile) does, but that package is much larger in scope. This PR won't change user-facing behavior without additional configuration on their side.

One possible friction point is if a user has already configured `ivy-rich`, etc. to match on `projectile-completing-read`. Then they will need to adjust their configuration to the new identifiers. If that is a big problem, we can add a configuration variable to make the new behavior opt-in or opt-out.

Examples of what can be done:

Show the current branch of the projects (very useful when juggling between multiple worktrees):
image

Have the same information with `projectile-switch-buffer` and `switch-buffer`:
image

 -----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 ### Changes
+* (TODO link): Add :caller information to calls to `ivy-read` (used by packages like `ivy-rich`)
 * [#1954](https://github.com/bbatsov/projectile/issues/1954): update ELisp for usage.html / "Removal of missing projects"
 * [#1947](https://github.com/bbatsov/projectile/issues/1947): `projectile-project-name` should be marked as safe
 * Set `projectile-auto-discover` to `nil` by default.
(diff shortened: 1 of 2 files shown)


ThibautVerron:
I'm not sure what action is expected or appropriate here. I ticked the last tasks (not applicable), and I've been using this code without issues for months now.

bbatsov:
My bad! I had totally missed your PR. 

It adds a bit of verbosity, but it seems like a step in the right direction. Thanks!

PR #1970, opened 2026-01-31
feat: apply .projectile ignore/ensure patterns to file listing

### LINKED ISSUE #1771: Globbing ignored in .projectile (projectile-current-project-files)
## Expected behavior

The globbing pattern in `.projectile` file is ignored in a subdir when running `projectile-current-project-files`.

In the example below, I would expect `y.text` not to be in the result, because it has been excluded in the `.projectile` file.

## Actual behavior

I get excluded patterns in the project files list.

```
("test.el" "mysubdir/x.txt" "mysubdir/y.text")
```

1. Create sample project like this:

```
mkdir test-hybrid{,/mysubdir}
touch test-hybrid/mysubdir/{x.txt,y.text}
echo "-*.text" > test-hybrid/.projectile
echo "(projectile-current-project-files)" > test-hybrid/test.el
```

2. Start emacs.

3. Set projectile indexing to `hybrid`.

```
(setq projectile-indexing-method 'hybrid)
```
4. Open `test.el`

5. Run the sexp with `C-u C-x C-e`.

```
Projectile version:  20220402.624
```

### Emacs version

```
GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.31, cairo version 1.16.0) of 2022-03-24
```

### Operating system

GNU/Linux

takeokunn:
## Summary

Fix #1771 - Apply `.projectile` ignore/ensure patterns to file listing using regex matching for recursive glob support.

## Changes

- Replace `file-expand-wildcards` with `wildcard-to-regexp` for pattern matching
- `projectile--pattern-to-regex`: Convert glob pattern to regex
- `projectile--file-matches-patterns-p`: Match patterns against file paths
  - Patterns without `/` (e.g., `*.text`) match against basename (recursive)
  - Patterns with `/` (e.g., `src/*.c`) match against full relative path
- Update `projectile-remove-ignored` to use new pattern matching
- Filter path traversal patterns (`..`) in `projectile-normalise-patterns`

## Rationale

The original implementation used `file-expand-wildcards` which only searches the current directory and requires files to exist on the filesystem.
By using `wildcard-to-regexp`, patterns are converted to regular expressions that can match against any file path string, enabling recursive matching across all subdirectories.

## Test Coverage

- Test for basic glob pattern filtering (`*.text`, `*.log`)
- Test for ensure pattern priority (unignore with `!`)
- Test for false positive prevention (basename matching)
- Test for path traversal pattern filtering (`..`)


bbatsov:
The approach here was the right one and it directly shaped what landed: master now applies dirconfig patterns in hybrid's post-filter via a compiled gitignore-style matcher shared with native indexing (23599b4), which also covers directory patterns like `-vendor/`. The code has diverged too far for a rebase to make sense, so closing. Thanks!

PR #1975, opened 2026-02-15
Add projectile-last-known-project variable

### LINKED ISSUE #1809: Add variable to remember most recent project
I think it would be useful to provide a variable such as `projectile-most-recent-project` that remembers the most recent project that you were working in. It is common that I am working on a project and then I switch to a buffer that is related to what I am doing even though it is not a member of any project. Unfortunately when I try to switch projectile buffers again I get "You're not in a project".

If such a variable existed I could easily write my own projectile wrapper functions that default to the most recent project.

I am not the only person that has wanted a way to access the most recent project: #664

I looked into opening a PR that adds this variable, but I could not figure out which function should set this variable so it works reliably across all of projectile. If you could give me some advice on how to do this I would be happy to make the change myself!

bbatsov:
- Adds `projectile-last-known-project` variable that tracks the root of the most recently accessed project
- Set automatically by `projectile-acquire-root`, which is called by most Projectile commands
- Allows users to reference the last project from non-project buffers (e.g., `*scratch*`, `*Messages*`)

Closes #1809

### DIFF
--- a/projectile.el
+++ b/projectile.el
@@ -662,6 +662,11 @@ See also `projectile-remove-known-project',
 Contains a copy of `projectile-known-projects' when it was last
 synchronized with `projectile-known-projects-file'.")
 
+(defvar projectile-last-known-project nil
+  "The root of the most recently accessed project.
+This is set automatically by `projectile-acquire-root' and can be
+used to reference the last project from non-project buffers.")
+
 (defcustom projectile-known-projects-file
   (expand-file-name "projectile-bookmarks.eld"
                     user-emacs-directory)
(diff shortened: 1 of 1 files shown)


bbatsov:
Superseded by the #1809 work that's landing in 3.0: `projectile-most-recent-project` plus `projectile-switch-to-most-recent-project` already cover jumping back to the previous project. Adding a second, similarly-named variable written on every `projectile-acquire-root` (a hot path) for the narrower non-project-buffer case would be more confusing than helpful, so I'm going to pass on this one. Thanks for the PR, and sorry it sat so long!

PR #1983, opened 2026-02-16
Add subproject compile and test commands

### LINKED ISSUE #1653: Feature request: Support build of single maven module in multi-module maven projects
In maven multi module builds, I frequently want to build only the module I'm currently working in (because that's a lot faster than building the entire reactor build).

It would be nice if projectile had a command to build just the current  maven module.

I guess it would be nice to have current-module specific versions of much of the functionality that currently takes you to the top of the projects (e.g. open dired on the module top).

But what I'm primarily missing is building just the current module.

bbatsov:
Closes #1653.

- Add `projectile-subproject-root` helper that walks up from the current directory to find the nearest build file (`:project-file` marker), stopping at the project root
- Add `projectile-compile-subproject` — runs the project's compile command in the nearest subproject directory
- Add `projectile-test-subproject` — runs the project's test command in the nearest subproject directory
- Keybindings: `c m c` (compile subproject), `c m t` (test subproject)

This works for any project type that defines `:project-file` — Maven (`pom.xml`), Gradle (`build.gradle`), Leiningen (`project.clj`), etc. The implementation let-binds `projectile-project-compilation-dir` so the existing `projectile--run-project-cmd` infrastructure handles everything else (command caching, buffer naming, prompting, etc.).

### Example: Maven multi-module project

```
myproject/
  pom.xml          <- project root
  module-foo/
    pom.xml        <- subproject root (found when editing Foo.java)
    src/main/java/
      Foo.java
  module-bar/
    pom.xml
    src/main/java/
      Bar.java
```

With cursor in `Foo.java`, `C-c p c m c` runs `mvn -B clean install` in `module-foo/` instead of the project root.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [#1964](https://github.com/bbatsov/projectile/issues/1964): Implement `project-name` and `project-buffers` methods for the `project.el` integration, so that code using `project.el` APIs returns correct results for Projectile-managed projects.
 * [#1837](https://github.com/bbatsov/projectile/issues/1837): Add `eat` project terminal commands with keybindings `x x` and `x 4 x`.
 * Add keybinding `A` (in the projectile command map) and a menu entry for `projectile-add-known-project`.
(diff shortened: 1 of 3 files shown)


steinarb:
@bbatsov I noticed today that the `C-c C-p c m c RET` command works as expected when on a source file in a sub-project, but not when done from a dired inside the subproject.

Not sure if that is intentional or not?

But it was confusing because it seemed like sometimes the command worked and sometimes it didn't (and to me a dired or a file inside the project is the same thing).

Thanks!

PR #1984, opened 2026-02-20
feat: make VCS marker detection order configurable

stribb:
## Motivation

`projectile-project-vcs` checks VCS markers in a fixed order — `.git`
before `.jj` — so a colocated jj+git repo is always detected as `git`.

This is a common situation rather than an edge case: jj's most popular
backend *is* git. When you run `jj git init` in an existing repo, or
clone with `jj git clone`, you get a directory that contains both a
`.jj/` folder and a `.git/` folder. Users of jj-on-git who want
projectile to treat such repos as jj projects have no recourse short of
advising `projectile-project-vcs`.

## Changes

Add `projectile-vcs-markers` — an ordered alist of `(marker . vcs-symbol)`
pairs — and rewrite `projectile-project-vcs` to iterate it for both the
root check and the dominating-file walk, replacing the hardcoded `cond`.

The default order is **unchanged** from the current behaviour, so this is
backwards-compatible for all existing users.

## Usage

To prefer jj over git in colocated repos:

```elisp
(setq projectile-vcs-markers
      (cons '(".jj" . jj)
            (assoc-delete-all ".jj" projectile-vcs-markers)))
```

## Tests

Four new Buttercup specs in `projectile-test.el` covering:
- single `.git` repo → `git`
- single `.jj` repo → `jj`
- colocated `.git`+`.jj` with default order → `git`
- colocated `.git`+`.jj` with custom order (jj first) → `jj`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1984](https://github.com/bbatsov/projectile/pull/1984): Add `projectile-vcs-markers` to make VCS detection order configurable.
 * [#1964](https://github.com/bbatsov/projectile/issues/1964): Implement `project-name` and `project-buffers` methods for the `project.el` integration, so that code using `project.el` APIs returns correct results for Projectile-managed projects.
 * [#1837](https://github.com/bbatsov/projectile/issues/1837): Add `eat` project terminal commands with keybindings `x x` and `x 4 x`.
 * Add keybinding `A` (in the projectile command map) and a menu entry for `projectile-add-known-project`.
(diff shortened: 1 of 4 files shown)


stribb:
Rebased onto master and added docs. Ready for merge.

bbatsov:
Landed in spirit on master (17ddffc): the 3.0 VCS detection rewrite already carried the marker alist internally, so it is now the `projectile-vcs-markers` defcustom, with your colocated jj+git case covered by tests. One semantic note: the order only breaks ties between markers in the same directory these days; the nearest marker always wins during the upward walk. Thanks!

PR #2010, opened 2026-06-12
Batch directory listings in project detection

bbatsov:
Follow-up to #2009 with the TRAMP perf work from that review. Two cache-miss hot paths in project detection were doing one `file-exists-p` per candidate; both now issue a single `directory-files` listing per directory, the same trick `projectile--vcs-from-directory-listing` already uses.

- `projectile-root-bottom-up`: `depth * markers` round-trips up a deep tree become `depth`.
- `projectile-detect-project-type`: dozens of stats across all registered types on first detection of a remote project become one.

Markers with a path separator fall back to `file-exists-p`, so nothing changes for `debian/control` and friends.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changes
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Landed on master as 568be01 with a few tweaks from review: the fallback stays on the caching `projectile-file-exists-p`, marker matching is now exact-case (which flushed out the `GNUMakefile` typo, fixed to `GNUmakefile`), lowercase `makefile` is recognized via a predicate, and single-marker walks short-circuit without a listing.

PR #2022, opened 2026-06-19
Add projectile-run-ghostel project terminal commands

Silex:
## Summary

Wire the [ghostel](https://github.com/dakra/ghostel) terminal emulator (libghostty-powered, a faster mouse-capable alternative to vterm/eat) into the `C-c p x` prefix, mirroring the existing vterm/eat integration:

- `projectile-run-ghostel` (`x G`) and `projectile-run-ghostel-other-window` (`x 4 G`), backed by a shared `projectile--ghostel` helper.
- ghostel is loaded lazily via `require ... noerror`, so it stays an optional dependency just like vterm and eat (no change to `Package-Requires`).
- Added to the `transient` dispatch menu and the `Run...` easy-menu.
- Documented in `usage.adoc`, `projectile_vs_project.adoc`, and the changelog.

## Implementation notes

`ghostel`'s entry point `(ghostel &optional arg)` creates its buffer from the dynamic value of `ghostel-buffer-name` and reuses buffers by a stored *identity* rather than the live buffer name — so reuse keeps working even after ghostel asynchronously retitles the visible buffer (OSC 2 / OSC 7). The helper therefore just `let`-binds `ghostel-buffer-name` to a project-scoped name (via `projectile-generate-process-name`) inside `projectile-with-default-dir`, the same shape as `projectile--eat`.

`ghostel-buffer-name` is declared special with a valueless `defvar` (next to the existing `eat-buffer-name`/`eshell-buffer-name` declarations) so the `let`-binding is dynamic under lexical-binding.

## Testing

- `emacs -Q --batch -L . -f batch-byte-compile projectile.el` — clean, no warnings.
- Verified the keybindings resolve (`x G` → `projectile-run-ghostel`, `x 4 G` → `projectile-run-ghostel-other-window`) and that invoking the command without ghostel installed errors with `Package 'ghostel' is not available`.
- The live terminal behaviour (opening/reusing a buffer, the other-window placement, `C-u` for an extra process) requires ghostel's native module and was not exercised in CI/batch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 * Add `projectile-forget-zombie-projects` as an alias for `projectile-cleanup-known-projects`, for discoverability and parity with `project.el`'s `project-forget-zombie-projects`.
 * `projectile-kill-buffers-filter` now also accepts a composable list of conditions (buffer-name regexps, predicates, and `major-mode`/`derived-mode`/`not`/`and`/`or` forms), modeled on `project.el`'s `project-kill-buffer-conditions`. The existing `kill-all`, `kill-only-files`, and predicate-function values keep working unchanged.
 * [#1837](https://github.com/bbatsov/projectile/issues/1837): Add `eat` project terminal commands with keybindings `x x` and `x 4 x`.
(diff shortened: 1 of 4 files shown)


bbatsov:
Thanks!

Issue #2042, opened 2026-06-29
getting projectile--dir-files-alien-await: Projectile indexing failed


pbgc:
Hi,

On MacOS 15.7.4

Upgrading to the last dev version i'm getting this error when trying to use projectile-find-file

projectile--dir-files-alien-await: Projectile indexing failed: exit code 1: fd -H -0 -E .git -tf --strip-cwd-prefix -c never.  See the *projectile-files-errors* buffer for details

helm-projectile gives:

In ‘Projectile files’ source: ‘#[nil ((when (projectile-project-p) (with-temp-buffer (hack-dir-local-variables-non-file-buffer) (helm-projectile--files-display-real (projectile-current-project-files) (projectile-project-root))))) (t)]’ 
 (user-error "Projectile indexing failed: exit code 1: fd -H -0 -E .git -tf --strip-cwd-prefix -c never.  See the *projectile-files-errors* buffer for details")

*projectile-files-errors* is empty ..

Thanks

pbgc:
@bbatsov now I'm getting:

Find file: [{: Command not found.
}: Command not found,] [Matched]

projectile-files-errors is still empty

I'm not using invalid projects! I'm using projects that I use with projectile for long (some for many years!) and none works.

pbgc:
@bbatsov 

the problem is caused by this line on my config:

(setq explicit-shell-file-name "/bin/tcsh")

I have this for ages and always worked with projectile .... removing it makes it work again

bbatsov:
Glad to hear this! I was quite puzzled by the problem until now. 😅

pbgc:
Sorry for your lost time! I was not really also understanding and tried to isolate the problem until spotting the line.
I had it for ages ... but it was only useful when using the emacs shells .. now I use vterm / ghostel everytime and can use Fish.

projectile always worked before .. but now it seems that it requires a POSIX shell-file-name

Thanks!

bbatsov:
Heads-up: this should now be properly fixed on master (#2048). The async indexer was building a POSIX-sh command (`{ ...; } 2>file`) but running it through your interactive shell, which tcsh can't parse. It now runs that command under `/bin/sh` regardless of your login shell, so you shouldn't need to drop `explicit-shell-file-name` anymore. Thanks for tracking down the root cause!

PR #2048, opened 2026-06-30
Run async indexing under a POSIX shell

bbatsov:
Fixes the regression reported in the #2042 follow-up.

The asynchronous indexer wraps its command as `{ ...; } 2>file` and ran it via the user's interactive `shell-file-name`. That wrapper (and the `shell-quote-argument` quoting of the stderr file) is POSIX-sh syntax, so a user whose `shell-file-name` is `csh`/`tcsh`/`fish` got shell parse errors (`{: Command not found`) instead of a file list - which the recently-added non-zero-exit leniency then surfaced as bogus completion candidates, with an empty `*projectile-files-errors*`. (The reporter traced it to `(setq explicit-shell-file-name "/bin/tcsh")`.)

The indexing command is a plain POSIX command, so the fix is to run it under `/bin/sh` regardless of the user's login shell. That also makes the `{ ...; } 2>file` wrapper and `shell-quote-argument` quoting consistent with the shell that actually runs them.

Added a spec that binds `shell-file-name` to a bogus value and confirms async indexing still works (proving we no longer use it).

Note: the synchronous runner uses `process-file-shell-command` with a *plain* command (no POSIX-only wrapper), which is why it kept working under tcsh - so it's left as-is.


jeremymeng:
Windows and `'alien` indexing now broken by this?

PR #2104, opened 2026-07-20
Honor Projectile's ignore rules under alien indexing

bbatsov:
`alien` is the default indexing method and it applied none of Projectile's ignore rules - not the globally-ignored files/dirs/suffixes, not the `-` entries of a `.projectile`. We even shipped a warning whose only job was to say the dirconfig was being bypassed.

Post-filtering the output is what `hybrid` already does, so instead the rules are handed to the tool: exclude pathspecs for `git ls-files`, `-E` globs for `fd`. Tools that can't express exclusions (svn, fossil, bzr, darcs, pijul, plain `find`) get their output filtered in Lisp. `projectile-alien-honors-ignores` turns it all off.

Two follow-on changes fell out of testing it against `native`: `+` keep and `!` unignore entries stay hybrid/native only (no tool equivalent), and ignore matching is now case-sensitive everywhere - `native`/`hybrid` used to fold case, which `fd` can't reproduce at all.

Alien projects will list fewer files after this. That's the point, but it's the kind of change worth knowing about.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,22 +8,22 @@
 
 ### New features
 
-- Alien indexing now honors Projectile's ignore rules, which it previously skipped entirely.
+- [#2104](https://github.com/bbatsov/projectile/pull/2104): Alien indexing now honors Projectile's ignore rules, which it previously skipped entirely.
   - The rules are pushed into the external tool (`git ls-files` exclude pathspecs, `fd --exclude`), so the filtering still happens outside Emacs.
   - Tools that can't express exclusions (svn, fossil, bzr, darcs, pijul, plain `find`) have their output filtered in Emacs Lisp instead.
   - Dirconfig `+` keep and `!` unignore entries have no equivalent in the tools and stay `hybrid`/`native` only.
   - Alien projects will list fewer files than before; set `projectile-alien-honors-ignores` to nil for the old behavior.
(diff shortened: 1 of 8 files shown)


bbatsov:
Trimmed `.ensime_cache` and `.eunit` from the default ignored dirs, and added a table to indexing.adoc for which tools take the exclusions and which get filtered in Lisp.

Comparing alien against `native` (not hybrid - both start from the same listing, so the original test proved nothing) turned up two bugs: hybrid was getting the exclusions pushed down too, and a `*` in `projectile-globally-ignored-directories` was read as a wildcard when it's really an "any depth" marker. Both fixed.

Separately: `native` treats `*tmp` as a literal dir name while `hybrid` strips the `*`, and ignoring.adoc claims they match. Left alone, probably wants its own issue.

bbatsov:
Made ignore matching case-sensitive everywhere instead of teaching the tools to fold case. `git` can do it with `icase` pathspecs but `fd --exclude` can't at all, and `projectile-git-use-fd` defaults on when fd is installed - so replicating it would make the file set depend on whether fd is on PATH.

The folding was incidental anyway: `string-suffix-p` called with IGNORE-CASE, and `string-match-p` inheriting `case-fold-search`. One spec pinned the old behaviour (`foo.tar.GZ` removed by `.tar.gz`); it comes from f866041 in 2017, which was about treating the entries as suffixes at all, so the spec got updated rather than the behaviour kept.

